PR #17, opened 2016-01-08
FUZZ-234 Changed version number to match PEP 0440. Moved to __init__.…

jtpereyda:
…py (setup.py reads it dynamically).


jtpereyda:
This PR was based on a branch with the wrong base. Closed and moved to PR #18

PR #18, opened 2016-01-08
Changed version number to PEP 440 style. Added to __init__.py and referenced in setup.py.

jtpereyda:


jtpereyda:
Works in by-hand functional test.

PR #22, opened 2016-01-15
Udp checksum fixes

jtpereyda:
Boring, complicated, refactors....


jtpereyda:
Passes manual inspection functional test comparing against earlier version. We need a better test method.

PR #44, opened 2016-03-07
Fixed link to boofuzz github in install.rst

wirkungsneutral:
https://github.com/jtpereyda/boofuzz was pointing to pydbg github


jtpereyda:
@wirkungsneutral Thank you for the fix and good find!

Issue #37, opened 2016-02-16
Too much boilerlate!


jtpereyda:
The first five lines of this example are all boilerplate: https://github.com/jtpereyda/boofuzz-ftp/blob/master/ftp.py

This is way more boilerplate than necessary. We can add a default logger to FuzzLogger, or even to Session. And the Target class is pretty meaningless in this example. If we make Session the direct manager for connections (and process monitors, etc.), we can nix this Target middle man altogether.

And if we go really stir crazy, we can even make sleep_time default to zero!

The brave new boilerplate may look like:
```
session = sessions.Session(
        connection=SocketConnection(“127.0.0.1”, 8021, proto=’tcp’))
```

jtpereyda:
Easy steps to reduce boilerplate:

1. Make `Session.__init__` `sleep_time` default to zero.
2. Make `Session.__init__` choose its own default `fuzz_data_logger`.
3. Make `Session.__init__` take a Connection and generate a default target upon construction. Note: This is risky/kludgy, since I don't have enough use cases involving Target's other features (netmon/procmon/vmcontrol). But long term I see these being added to Session each on their own.

PR #76, opened 2016-07-27
Reduce boilerplate

jtpereyda:
Issue #37


jtpereyda:
Passes functional test

PR #86, opened 2016-09-18
Fixed #85.

### LINKED ISSUE #85: socket.error: [Errno 10054] An existing connection was forcibly closed by the remote host
One user has reported and I've personally seen errors like this before:

```
  File "Z:\ftp\ftp.py", line 119, in <module>
    sess.fuzz()
  File "C:\Python27\lib\site-packages\boofuzz\sessions.py", line 414, in fuzz
  File "C:\Python27\lib\site-packages\boofuzz\sessions.py", line 868, in _fuzz_current_case
    self.transmit(target, self.fuzz_node, edge)
  File "C:\Python27\lib\site-packages\boofuzz\sessions.py", line 739, in transmit
    self.last_recv = self.targets[0].recv(10000)
  File "C:\Python27\lib\site-packages\boofuzz\sessions.py", line 110, in recv
    data = self._target_connection.recv(max_bytes=max_bytes)
  File "C:\Python27\lib\site-packages\boofuzz\socket_connection.py", line 143, in recv
socket.error: [Errno 10054] An existing connection was forcibly closed by the remote host
```

This seems to indicate a crash of the unit under test, which boofuzz should report as an error instead of crashing itself.

I don't know if there are times when this is valid behavior; perhaps in some protocols.

If anyone has any further details or ideas, please post.

marpie:
Hi there,

this fixes the issue (#85) for me. I tried it on the executable that triggered the issue yesterday and now the fuzzing continues.

In my case the fuzzee just closed the connection due to a malformed request, so no crash or any abnormal behavior happend that needs to be investigated further.

Therefore I think simply catching the exception and setting last_recv to None will do the trick.

Kind regards.

---
This change is [Reviewable](https://reviewable.io/reviews/jtpereyda/boofuzz/86)

### DIFF
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -191,7 +191,7 @@ class Session(pgraph.Graph):
         @kwarg crash_threshold     (Optional, def=3) Maximum number of crashes allowed before a node is exhaust
         @type  restart_sleep_time: int
         @kwarg restart_sleep_time: (Optional, def=5) Time in seconds to sleep when target can't be restarted
-        @type  web_port:	       int
+        @type  web_port:           int
         @kwarg web_port:           (Optional, def=26000) Port for monitoring fuzzing campaign via a web browser
         @type fuzz_data_logger:    fuzz_logger.FuzzLogger
         @kwarg fuzz_data_logger:   (Optional, def=Log to STDOUT) For saving test data and results.
@@ -736,7 +736,11 @@ class Session(pgraph.Graph):
 
         # Receive data
         # TODO: Remove magic number (10000)
-        self.last_recv = self.targets[0].recv(10000)
+        try:
+            self.last_recv = self.targets[0].recv(10000)
+        except socket.error:
+            print("[debug] here")
+            self.last_recv = None
 
         if self._check_data_received_each_request:
             self._fuzz_data_logger.log_check("Verify some data was received from the target.")


marpie:
After testing with another binary the issue does not seem to be resolved entirely. I will work on the solution and update this pull-request as soon as I have a solution.

jtpereyda:
@marpie That is tricky, it may indicate an error sometimes but a non-error other times. Thank you for the pull request!

I think you did exactly the right thing by just setting last_recv to None. The code right afterward can check to see if a failure should be raised, based on `self._check_data_received_each_request`.

Here are some suggestions, but they aren't necessarily a prerequisite for accepting the PR.
 1. One of my semi long-term refactoring goals is to move all `socket` code out of `sessions.py`. The ideal place to catch this exception would be in `socket_connection.py`.
 2. You can define a new exception in `boofuzz/sex.py`. You might call it something like `BoofuzzConnectionReset`. Then, in `socket_connection.py` `recv()`, you'd catch the socket error and raise your custom exception.
3. Finally, catch the custom exception in sessions.py.

This might seem like a goofy run-around, but the major upshot is that it keeps `sessions.py` socket-agnostic. Then, any new connection layers that people write will be able to hook into all the new functionality just by raising your custom exception when the connection is reset.

Another simpler option would be to just catch the exception in `socket_connection.py` and return no data. Actually, I think that might be all that's necessary, unless `sessions.py` ever needs to do anything special in that case.

jtpereyda:
@marpie Gentle bump

marpie:
Just pushed my new pull-request that implements error handling and reconnection if the application dropped it.

https://github.com/jtpereyda/boofuzz/pull/90

PR #98, opened 2016-10-13
Fix bug: Oversized UDP payloads would cause crash

jtpereyda:
Fix an issue in which I was getting `[Errno 90] Message too long` AKA `errno.EMSGSIZE`. There were a few problems:

1. SocketConnection had a default `MAX_PAYLOADS['udp']` value of 65536, where it should be 65507.
2. `helpers.get_max_udp_size()` was return a way oversized value.
3. There was no sanity check to make sure `helpers.get_max_udp_size()` didn't return a theoretically impossible value.

This PR fixes 1 and 3. It stop-gaps but doesn't resolve problem 2; see Issue #99 for that.

---
This change is [Reviewable](https://reviewable.io/reviews/jtpereyda/boofuzz/98)


jtpereyda:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jtpereyda/boofuzz/98)*

jtpereyda:
AppVeyor failures OK since AppVeyor isn't set up yet.

PR #104, opened 2016-11-07
Fix excess packages

jtpereyda:
Fix setup.py to not include superfluous packages. Regression occurred at 81e5d61bbcdf1cb6ade20f595a670aa84759afd5, Pull Request #43.

This is a really lame error, as it will obscure a user's installed `requests` package. Unfortunately, IMO, the Python packaging system makes this slip-up kind of easy. Oh well.

---
This change is [Reviewable](https://reviewable.io/reviews/jtpereyda/boofuzz/104)


jtpereyda:
AppVeyor fail OK since it isn't configured properly yet.

PR #108, opened 2017-01-10
adding expanded command injection strings

bsamuels453:
adding command injection strings sourced from https://github.com/fuzzdb-project/fuzzdb

maintainer may want to change command from "reboot" to "notepad", or other some other command.

---
This change is [Reviewable](https://reviewable.io/reviews/jtpereyda/boofuzz/108)


jtpereyda:
@bsamuels453 Thanks! Very cool. The fuzzer could be used against any target, so I wouldn't choose notepad over reboot necessarily. It might be nice to use a number of different commands though...

PR #119, opened 2017-02-17
Upgrade to newer version of pydot to avoid import error message

cbisnett:
This patch removes the temporary pydot2 dependency and replaces it with a newer pydot release with support for pyparsing > 1.5.7. pydot2 1.0.33 gives the following error when imported from a new boofuzz install.

```
$ pip install boofuzz
Collecting boofuzz
... snip ...
Successfully installed Flask-0.10.1 Jinja2-2.9.5 MarkupSafe-0.23 Werkzeug-0.11.15 backports.ssl-match-hostname-3.5.0.1 boofuzz-0.0.6 certifi-2017.1.23 future-0.16.0 impacket-0.9.15 itsdangerous-0.24 pydot2-1.0.33 pyserial-3.2.1 tornado-4.0.2

$ pip freeze | egrep 'pydot|pypars'
pydot2==1.0.33
pyparsing==2.1.10

$ python -c 'import pydot'
Couldn't import dot_parser, loading of dot files will not be possible.
```

---
This change is [Reviewable](https://reviewable.io/reviews/jtpereyda/boofuzz/119)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
         packages=find_packages(exclude=['unit_tests', 'requests', 'examples', 'utils', 'web', 'new_examples']),
         package_data={'boofuzz': ['web/templates/*', 'web/static/css/*']},
         install_requires=[
-            'future', 'pyserial', 'pydot2==1.0.33', 'tornado==4.0.2',
+            'future', 'pyserial', 'pydot==1.2.3', 'tornado==4.0.2',
             'Flask==0.10.1', 'impacket'],
         extras_require={
             # This list is duplicated in tox.ini. Make sure to change both!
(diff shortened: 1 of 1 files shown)


jtpereyda:
Very cool! That was pretty annoying. What do you think of switching from `==` to `>=`? This would allow higher versions of `pydot` as they become available. Reference: https://packaging.python.org/requirements/#install-requires

Right now only a few dependencies are pinned, and I don't think there's any good reason for it. `tornado` doesn't need to be pinned either, but I haven't taken the time to make sure it works with more up to date versions.

cbisnett:
I don't see any reason it needs to be pinned. I was just following along with what was already there ;)

jtpereyda:
@cbisnett Dangerous thing to do in this code base... :) Thanks again, and happy fuzzing!

PR #121, opened 2017-02-17
Add with statement support to blocks

cbisnett:
Added support for automatically scoping blocks using Python's `with` statement. For example what used to be written like this:
```
if s_block_start("myblock"):
    s_string("foobar")
    ...
s_block_end()
```
Can now be simplified to this:
```
with s_block("myblock"):
    s_string("foobar")
    ...
```

---
This change is [Reviewable](https://reviewable.io/reviews/jtpereyda/boofuzz/121)

### DIFF
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -126,13 +126,12 @@ def s_switch(name):
 # ## BLOCK MANAGEMENT
 
 
-def s_block_start(name, group=None, encoder=None, dep=None, dep_value=None, dep_values=(), dep_compare="=="):
-    # TODO: Either convert this to a with() statement, or add a new one that is compatible.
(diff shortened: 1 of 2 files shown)


cbisnett:
Actually now that I think about it, the default argument values aren't being passed when calling through `s_block_start()`. I'll update the branch.

jtpereyda:
Very cool! Did you do a manual run of unit_test/blocks.py? I don't think it is included in the automatically run tests; IIRC some of them weren't passing, but if the new test you wrote is passing, that's probably fine.

jtpereyda:
I would try now but I'm having issues on my current system.

jtpereyda:
I got my system up and tested it. Good to go. Thank you for the contributions! Feel free to add yourself to the CONTRIBUTORS.txt file.

PR #142, opened 2017-04-16
Move coredumps

ettisan:
Added an option to process_monitor_unix.py to move coredumps for later analysis

Only Linux is supported.

---
This change is [Reviewable](https://reviewable.io/reviews/jtpereyda/boofuzz/142)


jtpereyda:
Very cool, nice work and thank you!

Issue #107, opened 2016-11-29
s_checksum issues


WenhaoYu:
Hello everyone,I find if i pass a my own defined algorithm function to s_checksum,it may cause a error. so i check the code,and do some changes as blow:
in checksum.py:
add algorithm argument's type judgement
if type(self._algorithm) is not str:
            self.checksum_lengths[self._algorithm.__name__] = self._length
and in _get_dummy_value function :
def _get_dummy_value(self):
        if type(self._algorithm) is str:
            return self.checksum_lengths[self._algorithm] * '\x00'
        else:
            return self.checksum_lengths[self._algorithm.__name__] * '\x00'
hope it would be helpful  or anyone have good idea?

jtpereyda:
WenhaoYu,

You said you already tried to pass in your own function. Could you post the code you used to do this? It appears that this should have worked.

If you want to add an algorithm into the code (which would make sense if it something that a lot of others may want to use), you should modify `Checksum._checksum()` and `Checksum.checksum_lengths`.

jtpereyda:
Closed due to inactivity from opener and ottimo's related fix in PR #149.

PR #149, opened 2017-06-21
Fix return dummy value in custom checksum (issue #107)

ottimo:
When a custom checksum is used like in Issue #107, dummy value is not returned correctly because it always search in checksums_lenghts dictionary but custom function cannot be here.
I fix this error using length value to generate the dummy value.

---
This change is [Reviewable](https://reviewable.io/reviews/jtpereyda/boofuzz/149)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Features
 
 Fixes
 -----
+-  Fixed: add a dummy value custom checksum in checksum helper (issue #107)
 
 0.0.8
 =====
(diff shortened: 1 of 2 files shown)


jtpereyda:
Good change! Could you add a bullet point to CHANGELOG.rst under Upcoming, under Fixes? Thanks! (This is documented in CONTRIBUTING.rst for the record)

ottimo:
Sorry, I forgot it. :)

jtpereyda:
Thanks much @ottimo ! Is your Twitter handle the same? Sometimes I tag people in release tweets.

ottimo:
On Twitter I am [@ottimorosi](https://twitter.com/ottimorosi) :)
Thanks a lot.

PR #151, opened 2017-07-01
Add new s_fuzz_list primitive to read fuzz value from files

ottimo:
Now with s_fuzz_list it is possible to add values from files generated by radamsa
By example, I create 3 mutated value of "user" string
```
echo -ne "user" | radamsa -n 3 -o /tmp/fuzzlist%n.rad
```
Now to use them in test cases, I add this primitive in boofuzz model
```
s_fuzz_list("anonymous", filename="/tmp/fuzzlist*.rad")
```

---
This change is [Reviewable](https://reviewable.io/reviews/jtpereyda/boofuzz/151)

### DIFF
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -489,7 +489,7 @@ def s_string(value, size=-1, padding="\x00", encoding="ascii", fuzzable=True, ma
 
 def s_from_file(value, size=-1, padding="\x00", encoding="ascii", fuzzable=True, max_len=0, name=None, filename=None):
     """
-    Push a string onto the current block stack.
+    Push a value from file onto the current block stack.
 
     @type  value:    str
     @param value:    Default string value
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -23,7 +23,7 @@ from .ifuzz_logger_backend import IFuzzLoggerBackend
 from .itarget_connection import ITargetConnection
 from .primitives import (BasePrimitive, Delim, Group,
                          RandomData, Static, String, BitField,
-                         Byte, Word, DWord, QWord, FuzzList)
+                         Byte, Word, DWord, QWord, FromFile)
 from .serial_connection import SerialConnection
 from .sessions import Session, Target
 from .sex import SullyRuntimeError, SizerNotUtilizedError, MustImplementException
@@ -487,7 +487,7 @@ def s_string(value, size=-1, padding="\x00", encoding="ascii", fuzzable=True, ma
     s = primitives.String(value, size, padding, encoding, fuzzable, max_len, name)
     blocks.CURRENT.push(s)
 
-def s_fuzz_list(value, size=-1, padding="\x00", encoding="ascii", fuzzable=True, max_len=0, name=None, filename=None):
+def s_from_file(value, size=-1, padding="\x00", encoding="ascii", fuzzable=True, max_len=0, name=None, filename=None):
     """
     Push a string onto the current block stack.
 
@@ -509,7 +509,7 @@ def s_fuzz_list(value, size=-1, padding="\x00", encoding="ascii", fuzzable=True,
     @param filename: (Mandatory) Specify filename where to read fuzz list
     """
 
-    s = primitives.FuzzList(value, size, padding, encoding, fuzzable, max_len, name, filename)
+    s = primitives.FromFile(value, size, padding, encoding, fuzzable, max_len, name, filename)
     blocks.CURRENT.push(s)
 
 # noinspection PyTypeChecker
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -487,16 +487,12 @@ def s_string(value, size=-1, padding="\x00", encoding="ascii", fuzzable=True, ma
     s = primitives.String(value, size, padding, encoding, fuzzable, max_len, name)
     blocks.CURRENT.push(s)
 
-def s_from_file(value, size=-1, padding="\x00", encoding="ascii", fuzzable=True, max_len=0, name=None, filename=None):
+def s_from_file(value, encoding="ascii", fuzzable=True, max_len=0, name=None, filename=None):
     """
     Push a value from file onto the current block stack.
 
     @type  value:    str
     @param value:    Default string value
-    @type  size:     int
-    @param size:     (Optional, def=-1) Static size of this field, leave -1 for dynamic.
-    @type  padding:  Character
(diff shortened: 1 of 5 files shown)


jtpereyda:
Sounds like this is mostly for getting fuzz values from a file or files. What do you think of renaming to `s_from_file` or something similar? `s_fuzz_list` sounds to me like I would be specifying a list of values. `s_group` currently fulfills the list functionality.

jtpereyda:
Very cool! I'll have a few suggestions on the code details...

boofuzz/__init__.py
```diff
@@ -487,6 +487,30 @@ def s_string(value, size=-1, padding="\x00", encoding="ascii", fuzzable=True, ma
     s = primitives.String(value, size, padding, encoding, fuzzable, max_len, name)
     blocks.CURRENT.push(s)
 
+def s_fuzz_list(value, size=-1, padding="\x00", encoding="ascii", fuzzable=True, max_len=0, name=None, filename=None):
+    """
+    Push a string onto the current block stack.
```
jtpereyda: Modify description to indicate use of fuzz values from file.

boofuzz/primitives/fuzz_list.py
```diff
@@ -0,0 +1,125 @@
+import random
+import glob
+
+from .base_primitive import BasePrimitive
+
+
+class FuzzList(BasePrimitive):
```
jtpereyda: This could be simplified by utilizing more of what `BasePrimitive` gives us for free. `delim.py` is an excellent example of a primitive that uses `BasePrimitive` well. Basically you can just add stuff to `_fuzz_library` and `BasePrimitive` does the rest of the work.

boofuzz/primitives/fuzz_list.py
```diff
@@ -0,0 +1,125 @@
...
+    # store fuzz_library as a class variable to avoid copying the ~70MB structure across each instantiated primitive.
+    _fuzz_library = []
```
jtpereyda: These are unused: remove?

boofuzz/primitives/fuzz_list.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __init__(self, value, size=-1, padding="\x00", encoding="ascii", fuzzable=True, max_len=0, name=None, filename=None):
+        """
+        Primitive that cycles through a list of "bad" values from a file. The primitive take filename and open the file to read
```
jtpereyda: Suggested rewording:
```
        Cycles through a list of "bad" values from a file(s). Takes filename and open the file(s) to read
        the values to use in fuzzing process. filename may contain glob characters.
```

boofuzz/primitives/fuzz_list.py
```diff
@@ -0,0 +1,125 @@
...
+            _file_handle = open(fname,"r")
+            self.this_library.append(_file_handle.readlines())
+            _file_handle.close()
+
+        # TODO: Make this more clear
+        if max_len > 0:
+            # If any of our strings are over max_len
+            if any(len(s) > max_len for s in self.this_library):
+                # Pull out only the ones that aren't
+                self.this_library = list(set([s[:max_len] for s in self.this_library]))
```
jtpereyda: Lines 50-54 can be deleted. Line 55 by itself will be enough, since it is basically a filter. If there is nothing larger than `max_len`, the statement will simply have no impact.

boofuzz/primitives/fuzz_list.py
```diff
@@ -0,0 +1,125 @@
...
+        self.padding = padding
+        self.encoding = encoding
+        self._fuzzable = fuzzable
+        self._name = name
+        self._filename = filename
+        self.this_library = []
+        list_of_files = glob.glob(self._filename)
+        for fname in list_of_files:
+            _file_handle = open(fname,"r")
+            self.this_library.append(_file_handle.readlines())
```
jtpereyda: I think you want `.extend` instead of `.append`. `readlines()` returns a list, so using `append` will make `self.this_library` a list of lists. Example:
```python
>>> a = [1,2,3]
>>> a.extend([4,5])
>>> a
[1, 2, 3, 4, 5]
>>> a.append([6,7])
>>> a
[1, 2, 3, 4, 5, [6, 7]]
```

jtpereyda: For safety, use the Python with construct to open files instead of manually opening and closing. There is a small risk that an exception will happen between open and close, leaving the file handles unresolved. https://stackoverflow.com/a/3277516/461834

boofuzz/primitives/fuzz_list.py
jtpereyda: It looks like the major added functionality in mutate is:
 1. If size parameter is enabled, skip oversized values.
 2. If size parameter is enabled, pad undersized values.

These could be moved to pre-filters, kind of like you did with `max_len`. If you do that, you can drop the custom mutate function and have a really simple class like `Delim` . What do you think?

jtpereyda:
Thank you for the pull request! I made a few comments; feel free to push back on anything you disagree with as I may have misunderstood some things.

boofuzz/primitives/fuzz_list.py
ottimo: I prefer use max_len to use only anomalies with maximum length minor or equals to max_len.
For now I don't like truncate or add padding to anomalies generated by radamsa (I made a mistake in the code before your review), I think that modifies its output make it less useful. I hope I made clear my idea.

boofuzz/primitives/from_file.py
```diff
@@ -0,0 +1,50 @@
...
+        for fname in list_of_files:
+            with open(fname,"r") as _file_handle:
+                self._fuzz_library.extend(_file_handle.readlines())
+            
+        # TODO: Make this more clear
+        if max_len > 0:
+            # If any of our strings are over max_len
+            if any(len(s) > max_len for s in self._fuzz_library):
+                # Pull out only the ones that aren't
+                self._fuzz_library = list(set([s for s in self._fuzz_library if len(s) <= max_len]))
```
jtpereyda: Do you think it's better to truncate oversized values or just take them out? Seems like truncate would help in some cases, depending on what variety of strings are in the file.

boofuzz/primitives/from_file.py
```diff
@@ -0,0 +1,50 @@
+import random
```
jtpereyda: Unused import random

boofuzz/primitives/from_file.py
jtpereyda: Sorry, I didn't see your earlier comment about truncating!

PR #170, opened 2017-11-10
In the transmit() function read response only when the check_data_re…

hooki:
In the transmit() function read response data only when the check_data_received_each_request is True. This is to allow the user to process the response data through overriding the post_send()  function.

### DIFF
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -727,11 +727,11 @@ class Session(pgraph.Graph):
             self.targets[0].send(data)
             self.last_send = data
 
-            # Receive data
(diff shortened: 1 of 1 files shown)


jtpereyda:
This is great. An underlying assumption inherited from Sulley is that one always wants to receive after sending -- not necessarily true. The check_data_received_each_request was added to remove the failure but it still insisted on trying to receive each time.

jtpereyda:
@hooki Thank you for the PR! Would you be willing to add an entry in the CHANGELOG?

hooki:
@jtpereyda Thanks for the quick merge :)

PR #180, opened 2018-03-20
Fixed "No such process" error process_monitor_unix

hrtejada:
OSError: [Error 3] No such process thrown on line 91 when restarting a process. Specifically when the process has already crashed due to a fuzzing test case. Added a try-catch statement to see if pid exists.

### DIFF
--- a/process_monitor_unix.py
+++ b/process_monitor_unix.py
@@ -89,8 +89,12 @@ class DebuggerThread:
         return self.exit_status
 
     def stop_target(self):
-        os.kill(self.pid, signal.SIGKILL)
-        self.alive = False
+        try:
+            os.kill(self.pid, signal.SIGKILL)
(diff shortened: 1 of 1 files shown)


process_monitor_unix.py
```diff
@@ -89,8 +89,12 @@ def get_exit_status(self):
         return self.exit_status
 
     def stop_target(self):
-        os.kill(self.pid, signal.SIGKILL)
-        self.alive = False
+        try:
+            os.kill(self.pid, signal.SIGKILL)
+        except OSError:
+            return False
```
jtpereyda: Fantastic! This method didn't formerly return anything, so what do you think of dropping the `return False`? The return value is nowhere used in boofuzz; are you using it?

jtpereyda:
Thank you again!

Issue #197, opened 2018-07-05
boofuzz.s_size() is not calculating the size correctly for boofuzz.s_dword


hduarte:
When boofuzz.s_size() is used with the default math function, the size of a boofuzz.s_dword() is not being correctly calculated. For example if one calculates the size of a block with only an s_dword() inside the block and with a value of 1 digit, the reported size is 4. Changing the length of sizer doesn't change this.

What am I missing @jtpereyda ?

Thks for your help.

jtpereyda:
@hduarte at least for binary format, 4 bytes is the correct size. byte = 1 byte. word = 2 bytes. dword = 4 bytes. qword = 8 bytes.

jtpereyda:
@hduarte I confirmed that the length is calculated incorrectly in format 'ascii'

hduarte:
@jtpereyda yes that's the issue. I have submitted a pull request with the patch that calculates the correct length when the format is ascii https://github.com/jtpereyda/boofuzz/pull/198

jtpereyda:
@hduarte very good find.

PR #198, opened 2018-07-05
fixed how s_size works when the output_format is ascii

hduarte:
Related to this issue https://github.com/jtpereyda/boofuzz/issues/197


jtpereyda:
@hduarte Thank you and great find!

hduarte:
@jtpereyda Thks for your effort on boofuzz.

PR #227, opened 2018-11-20
add default_value param for s_group()

cq674350529:
The default behavior of `s_group()` when fuzzing is complete is to choose `values[0]` as its `original_value`. Sometimes, the same `values` may be used in many places, the `original_value` of which may be different.

The param `default_value` added allows users to specify the value they want. If not specified, `values[0]` is used instead, same as the default behaviour.


jtpereyda:
@cq674350529 very cool, thank you for the push!

Issue #226, opened 2018-11-19
Is there an easy way to keep multi fields in sync?


cq674350529:
Hi, sometimes when fuzzing xml, I need to keep two fields in sync.  Take the following xml as an example:

    <key>aaaa</key>

The possible value of xml tag may be [`key1`, `key2`]. Thus, I choose the `s_group()` like this:

    s_group('group_start', values=TEST_TAGS)
    with s_block('tmp'):
        s_static('>')
        s_static('aaaa')
        s_static('</')
    s_group('group_end', values=TEST_TAGS)

It will procedure 3 combinations (`<key1>aaaa</key1>`, `<key2>aaaa</key1>`, `<key1>aaaa</key2>`). But what I want are `<key1>aaaa</key1>` and `<key2>aaaa</key2>`.
> above example taken is kept as simple as possible

So is there an easy way to accomplish this?  Although I can do this with `edge.callback()` or `session.pre_send()`.

Any advice would be appreciated! Thanks in advance.

jtpereyda:
@cq674350529 I don't believe there's any element that does exactly this. This would be a cool feature to have. It would be doable and could be based off of `s_checksum`.

jtpereyda:
Thanks for s_mirror @cq674350529 !

PR #228, opened 2018-11-30
add s_mirror(), used to keep updated with other elements

cq674350529:
As discussed in [Issue #226](https://github.com/jtpereyda/boofuzz/issues/226#issue-382093545), a new primitive `s_mirror()` is added, which is used to keep updated with other elements. Also, a simple testcase has been given in `unit_tests/primitives.py`. 

Any suggestions? Thanks.

### DIFF
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -23,7 +23,7 @@ from .ifuzz_logger_backend import IFuzzLoggerBackend
 from .itarget_connection import ITargetConnection
 from .primitives import (BasePrimitive, Delim, Group,
                          RandomData, Static, String, BitField,
-                         Byte, Word, DWord, QWord, FromFile)
+                         Byte, Word, DWord, QWord, FromFile, Mirror)
 from .serial_connection import SerialConnection
 from .sessions import Session, Target, open_test_run
 from .exception import SullyRuntimeError, SizerNotUtilizedError, MustImplementException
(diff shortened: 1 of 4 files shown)


jtpereyda:
@cq674350529 Very cool!

A couple potential improvements, but neither of which need to be done now IMO. I'll go ahead and merge. Thank you for the PR!
1. In theory it would be nice to add the ability to mutate. The only strategy I can think of would be to mirror the target primitive's mutations. With the current way mutations work though, that could be tricky. The mutations are all stateful, so it's not trivial to get a mutation without changing the target primitive as well. A lot of things would work nicer if the primitives gave you a generator from which you could get mutations.
2. For test coverage you could add a test that checks the recursive checks. Something where the target of s_mirror is a length primitive (s_size), which is itself a length of a block containing s_mirror. But honestly that's getting a little crazy given the stability

cq674350529:
@jtpereyda  Thanks.

> 1. In theory it would be nice to add the ability to mutate. The only strategy I can think of would be to mirror the target primitive's mutations. With the current way mutations work though, that could be tricky. The mutations are all stateful, so it's not trivial to get a mutation without changing the target primitive as well. A lot of things would work nicer if the primitives gave you a generator from which you could get mutations.

As for `s_mirror()`, I think  it should just do what it's supposed to do - a mirror of target primitive. If the target primitive can mutate, then `s_mirror()` will keep updated with it all the time. That is, it's the target primitive's duty to mutate.

    s_string('aaa', name='target')
    s_static(':')
    s_mirror('target', name='mirror')
 
So at the moment, I just set `self._fuzzable = False`.  Maybe another better choice is to implement `mutate()` for `s_mirror()` like `s_static()`.

    def mutate(self):
        """
        Always return false, don't fuzz
        """
        return False

> 2. For test coverage you could add a test that checks the recursive checks. Something where the target of s_mirror is a length primitive (s_size), which is itself a length of a block containing s_mirror. But honestly that's getting a little crazy given the stability of boofuzz overall.

Ok, I'll do some tests according to the scene you describe.

jtpereyda:
@cq674350529 With the mutation, my thinking is that there might be a situation where the second instance of an item that is supposed to match being modified causes a bug. Not always a helpful strategy, but it could be in some scenarios. Definitely not something we need now.

> Ok, I'll do some tests according to the scene you describe.
Feel free to do so or not totally at your option. Since boofuzz is still pretty experimental I think the balance leans toward new features a little more than tests, so I don't want to make you feel like you _have_ to write any more tests.

Cheers!

PR #231, opened 2019-02-08
Customisable size for target recv function

hduarte:
Added the possibility to change the size of the target recv function. I've tried to add the feature while maintaining the existing implementation, so the changes are retro compatible with existing code that relies on boofuzz.


jtpereyda:
@hduarte Cool, thank you! Please add a line to CHANGELOG.rst if you would.

PR #232, opened 2019-02-12
Log only sent data when MAX_PAYLOADS is in effect.

SR4ven:
I noticed that boofuzz sometimes creates really large test case data which it does truncate according to the specified max_payloads. But the logger gets handed the generated data instead of the actually sent data.
This fix should help to reduce db and log size when fuzzing on layer 2/3, but wouldn't have any effect on tcp.

### DIFF
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -140,13 +140,13 @@ class Target(object):
         Returns:
             None
         """
-        num_sent, data_sent = self._target_connection.send(data=data)
+        if self._fuzz_data_logger is not None:
+            self._fuzz_data_logger.log_info("Sending {0} bytes...".format(len(data)))
+
+        num_sent = self._target_connection.send(data=data)
 
         if self._fuzz_data_logger is not None:
-            self._fuzz_data_logger.log_send(data_sent)
-
-        if self._fuzz_data_logger is not None:
-            self._fuzz_data_logger.log_info("{0} bytes sent".format(num_sent))
+            self._fuzz_data_logger.log_send(data[:num_sent])
 
     def set_fuzz_data_logger(self, fuzz_data_logger):
         """
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -140,10 +140,10 @@ class Target(object):
         Returns:
             None
         """
-        if self._fuzz_data_logger is not None:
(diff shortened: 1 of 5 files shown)


SR4ven:
The check fails because the send() function is expected to return only one value.
Adjusting
`send_result = uut.send(data=raw_packet)` to
`send_result = uut.send(data=raw_packet)[0]` in test_socket_connection.py line 287 and all following fixes this.

jtpereyda:
@SR4ven  Very cool! I'll review the code.

boofuzz/sessions.py
```diff
@@ -140,10 +140,10 @@ def send(self, data):
         Returns:
             None
         """
-        if self._fuzz_data_logger is not None:
-            self._fuzz_data_logger.log_send(data)
+        num_sent, data_sent = self._target_connection.send(data=data)
 
-        num_sent = self._target_connection.send(data=data)
+        if self._fuzz_data_logger is not None:
+            self._fuzz_data_logger.log_send(data_sent)
```
jtpereyda: We could get a similar effect with something like:
```python
            self._fuzz_data_logger.log_send(data[:num_sent])
```
If that works, we could also skip the changes to the way the send function works.

boofuzz/sessions.py
```diff
@@ -140,10 +140,10 @@ def send(self, data):
         Returns:
             None
         """
-        if self._fuzz_data_logger is not None:
-            self._fuzz_data_logger.log_send(data)
```
jtpereyda: I'm leaning toward keeping some kind of log message prior to the send action. It helps in the case where something fails while the send is happening.

Likewise, if we're only logging what is actually sent, we can get rid of the log action `self._fuzz_data_logger.log_info("{0} bytes sent".format(num_sent))` just a few lines below, as that information is now redundant. In fact, we could just move those two lines right here and change the string to "Sending {0} bytes".

boofuzz/sessions.py
SR4ven: Great idea, this works perfectly fine!

boofuzz/sessions.py
SR4ven: I made some changes according to your ideas, please check.

jtpereyda:
Thanks @SR4ven !

PR #233, opened 2019-02-14
Fix db_logger write

SR4ven:
- Db is now only written to disk on test case completion
- Added option to only write the log of the last n test cases to disk when an error or a failure are detected.
  First and last test case are always logged
- Commented out some unused imports according to PyCharm
- Minor formating fix in fuzz_logger_text.py

TODO: Changelog, documentation/comments

### DIFF
--- a/boofuzz/fuzz_logger.py
+++ b/boofuzz/fuzz_logger.py
@@ -70,6 +70,14 @@ class FuzzLogger(IFuzzLogger):
         for fuzz_logger in self._fuzz_loggers:
             fuzz_logger.log_send(data=data)
 
+    def end_test_case(self):
+        for fuzz_logger in self._fuzz_loggers:
+            fuzz_logger.end_test_case()
+
+    def end_test(self):
+        for fuzz_logger in self._fuzz_loggers:
+            fuzz_logger.end_test()
+
     def failure_summary(self):
         """Return test summary string based on fuzz logger results.
 
@@ -87,6 +95,3 @@ class FuzzLogger(IFuzzLogger):
             summary += "{0}".format('\n'.join(map(str, self.error_test_cases.iterkeys())))
 
         return summary
-
-    def write_log(self, force=False):
-        self._fuzz_loggers[0].write_log(force=force)
--- a/boofuzz/fuzz_logger.py
+++ b/boofuzz/fuzz_logger.py
@@ -70,13 +70,13 @@ class FuzzLogger(IFuzzLogger):
         for fuzz_logger in self._fuzz_loggers:
             fuzz_logger.log_send(data=data)
 
-    def end_test_case(self):
+    def close_test_case(self):
         for fuzz_logger in self._fuzz_loggers:
-            fuzz_logger.end_test_case()
+            fuzz_logger.close_test_case()
 
-    def end_test(self):
+    def close_test(self):
         for fuzz_logger in self._fuzz_loggers:
-            fuzz_logger.end_test()
+            fuzz_logger.close_test()
 
     def failure_summary(self):
         """Return test summary string based on fuzz logger results.
--- a/boofuzz/fuzz_logger.py
+++ b/boofuzz/fuzz_logger.py
@@ -87,3 +87,6 @@ class FuzzLogger(IFuzzLogger):
             summary += "{0}".format('\n'.join(map(str, self.error_test_cases.iterkeys())))
 
         return summary
+
+    def write_log(self, force=False):
+        self._fuzz_loggers[0].write_log(force=force)
--- a/boofuzz/fuzz_logger_csv.py
+++ b/boofuzz/fuzz_logger_csv.py
@@ -36,7 +36,7 @@ class FuzzLoggerCsv(ifuzz_logger_backend.IFuzzLoggerBackend):
     def __init__(self, file_handle=sys.stdout, bytes_to_str=DEFAULT_HEX_TO_STR):
         """
         Args:
-            file_hanlde (io.TextIOBase): Open file handle for logging. Defaults to sys.stdout.
+            file_handle (io.TextIOBase): Open file handle for logging. Defaults to sys.stdout.
             bytes_to_str (function): Function that converts sent/received bytes data to string for logging.
         """
         self._file_handle = file_handle
@@ -69,6 +69,12 @@ class FuzzLoggerCsv(ifuzz_logger_backend.IFuzzLoggerBackend):
 
     def log_pass(self, description=""):
         self._print_log_msg(["pass", "", "", description])
+
+    def end_test_case(self):
+        pass
+
+    def end_test(self):
+        pass
 
     def _print_log_msg(self, msg):
         time_stamp = get_time_stamp()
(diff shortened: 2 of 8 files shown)


boofuzz/fuzz_logger.py
```diff
@@ -87,3 +87,6 @@ def failure_summary(self):
             summary += "{0}".format('\n'.join(map(str, self.error_test_cases.iterkeys())))
 
         return summary
+
+    def write_log(self, force=False):
+        self._fuzz_loggers[0].write_log(force=force)
```
jtpereyda: From reading the code, it looks like this method is used to signal the end of a test case (or the end of an entire test). One concern I had initially is that the implementation details of the fuzz logger (e.g., when data should be written) were being leaked outside the class. However, once I realized that it essentially signals completion of a test case, it made more sense. Here are some readability/usability suggestions:

1. Rename the method to `end_test_case`.
2. Add an `end_test` method.
3. Add these methods to ifuzz_logger.py, and to each of the fuzz logger implementation (text, csv) for consistency. They can do nothing or log a simple message in those cases.
4. Within `fuzz_logger_db.py`, rename `write_log` to `_write_log`. The `end_test_case` method will call `_write_log` without force. The `end_test` method will call it with force.

This allows us to keep all content regarding database optimization strictly within the `FuzzLoggerDb` class. It will also make the code within `FuzzLoggerDb` more readable to a newcomer, as the public methods will make clear how and when `_write_log` is called.

boofuzz/sessions.py
```diff
@@ -280,6 +280,9 @@ class Session(pgraph.Graph):
         web_port (int):         Port for monitoring fuzzing campaign via a web browser. Default 26000.
         fuzz_data_logger (fuzz_logger.FuzzLogger): DEPRECATED. Use fuzz_loggers instead.
         fuzz_loggers (list of ifuzz_logger.IFuzzLogger): For saving test data and results.. Default Log to STDOUT.
+        fuzz_loggers_keep_num_log_cases (int): Number of most recent test cases to save to db in case of an error or
+                                               failure. Set to 0 to save after every test step (high disk I/O!)
+                                               Default 0
```
jtpereyda: Here's a rewording suggestion, emphasizing the point of this setting (minimizing disk usage), and changing the second sentence a bit to emphasize which cases are saved rather than how frequently.
> Minimize disk usage by only saving passing test cases if they are in the n test cases preceding a failure or error. Set to 0 to save every single test case (high disk I/O!). Default 0.

jtpereyda: I'm thinking about names... what about `fuzz_db_keep_only_n_pass_cases`?

boofuzz/fuzz_logger_db.py
```diff
@@ -67,49 +71,60 @@ def get_test_case_data(self, index):
...
-        self._database_connection.commit()
+        self._queue.append(["INSERT INTO steps VALUES(?, ?, ?, ?, ?);\n",
+            self._current_test_case_index, 'pass', description, b'', helpers.get_time_stamp()])
+
+    def write_log(self, force=False):
+
+        if self._queue_max_len > 0:
+            while (self._current_test_case_index - next(
+                    x for x in self._queue[0] if isinstance(x, (int, long)))) >= self._queue_max_len:
+                self._queue.popleft()
```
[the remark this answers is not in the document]
SR4ven: Sorry I should have commented that in advance.
The general idea was to always keep the log of e.g. 3 full test cases in RAM and only write them to disk if an error/fail occurs or fuzzing is complete. By setting deque's maxlen we could only limit the max number of single log lines. As a testcase doesn't always output the same amount of lines, this would end up in logging half test cases.
Another way that this could possibly be done would be with a 2D list inside the deque. That way every test case could get it's own row and as many columns as needed for the actual log messages. But I think this would make appending to the queue more complicated.

Now to your initial question about what that statement does:

1. You are right about the downsizing part. It keeps removing log lines until the max number of test cases is reached.
2. It searches for the first `int` in the oldest queue list, which should be the test case index of that log line. The problem I was facing here was, that the index field is not always on the same positon of the list. For `open_test_case()` it's on the second position, for all others it's on the first.
I didn't want to change the log format or the `get_test_case_data` functions so I went with this solution instead, which is clearly not best practice.

boofuzz/fuzz_logger_db.py
```diff
@@ -67,49 +71,60 @@ def get_test_case_data(self, index):
...
+        else:
+            force=True
+
+        if force or self._fail_detected or self._log_first_case:
+            for i in range(len(self._queue)):
+                self._db_cursor.execute(self._queue[0][0], self._queue[0][1:])
+                self._queue.popleft()
```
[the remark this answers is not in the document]
SR4ven: Thanks for the hint, will adopt.

boofuzz/sessions.py
SR4ven: > 
> 
> I'm thinking about names... what about `fuzz_db_keep_only_n_pass_cases`?

That sounds more intuitive, will rename.

boofuzz/fuzz_logger.py
SR4ven: That should make the code more readable, will rework.

boofuzz/fuzz_logger_csv.py
```diff
@@ -36,7 +36,7 @@ class FuzzLoggerCsv(ifuzz_logger_backend.IFuzzLoggerBackend):
     def __init__(self, file_handle=sys.stdout, bytes_to_str=DEFAULT_HEX_TO_STR):
         """
         Args:
-            file_hanlde (io.TextIOBase): Open file handle for logging. Defaults to sys.stdout.
+            file_handle (io.TextIOBase): Open file handle for logging. Defaults to sys.stdout.
```
SR4ven: Fixed that tiny typo.

SR4ven:
When trying to access a non existent test case in the Web UI there is a constant flow of errors like here:
```
File "/home/papgft/.local/lib/python2.7/site-packages/boofuzz/fuzz_logger_db.py", line 146, in get_test_case_data
    test_case_row = next(c.execute('''SELECT * FROM cases WHERE number=?''', [index]))
StopIteration
ERROR:tornado.access:500 GET /api/test-case/17 (127.0.0.1) 1.01ms
```
That only happens when using `boo open` to view the log. We might want to take a look at that here or in a new issue/request.

boofuzz/fuzz_logger_db.py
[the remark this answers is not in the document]
jtpereyda: @SR4ven That clears up a lot, thank you.

boofuzz/ifuzz_logger.py
```diff
@@ -173,3 +173,31 @@ def log_error(self, description):
...
+
+        :return: None
+        :rtype: None
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def end_test(self):
+        """
+        Called after a test case has been completed. Can be used to inform the operator
```
jtpereyda: "after a test ~case~ has been completed"

boofuzz/fuzz_logger.py
```diff
@@ -70,6 +70,14 @@ def log_send(self, data):
...
             fuzz_logger.log_send(data=data)
 
+    def end_test_case(self):
+        for fuzz_logger in self._fuzz_loggers:
+            fuzz_logger.end_test_case()
+
+    def end_test(self):
+        for fuzz_logger in self._fuzz_loggers:
+            fuzz_logger.end_test()
+
```
jtpereyda: Oops, that probably should have been "close_test_case" and "close_test" to match better with open. 🤷‍♂️ Feel free to change if you agree but not required.

jtpereyda:
> e

Issue #217, opened 2018-10-04
Unable to click on test cases in Web GUI


RythmStick:
Getting following error on console:
WARNING:tornado.access:404 GET /static/js/index.js

jtpereyda:
@RythmStick A few questions to help debug:
1. What OS are you using?
2. What install method did you use? (pip or install from source)
3. What version or git revision did you install?

jtpereyda:
I download the .tar.gz file from pypi and it does contain index.js, so it might be an installation issue.

RythmStick:
1. Ubuntu 4.15.0-34
2. pip 
3. Master Branch (Boofuzz v0.1.1)
.
index.js file does exist but tornado just gives a 404. Looks like a tornado config issue rather than missing file.

jtpereyda:
@RythmStick A couple things you might try:
1. Try doing a hard refresh in your browser. This is browser dependent, but in Chrome you can hold Ctrl while clicking refresh.

I'm a bit at a loss as I did this with a class of 20 students and it worked pretty consistently (they were on Ubuntu 18.04). They all had a fresh install in a VM though, so I suspect it has something to do with upgrading from a previous version. Did you install from source or from Pypi?

SR4ven:
Getting the exact same problem when installing from source as well as pip. The js folder is missing in static, but when I manuall add it everything works as expected.
I verified this:
> 
> 
> I download the .tar.gz file from pypi and it does contain index.js, so it might be an installation issue.

so it indeed seems to be an installation issue.

jtpereyda:
Fix merged and now included in the latest PyPI release. Closing. Thank you @RythmStick and @SR4ven !

PR #234, opened 2019-02-15
Fixed setup.py for issue #217

SR4ven:
The setup.py for pip installation is missing the package_data path to /web/static/js/.
This should fix issue #217.


jtpereyda:
Thank you @SR4ven !

Issue #243, opened 2019-02-26
Add Docs Build to CI


jtpereyda:
Avoid issues like #240

SR4ven:
PR #245 should have fixed that but the docs status is still on failing. Are they only rebuilt on a release or a manual trigger maybe?

jtpereyda:
@SR4ven You can see the builds here: https://readthedocs.org/projects/boofuzz/builds/ Looks like it has been trying to build

SR4ven:
That helps finding the problem!
The build fails because of a ldap3 version mismatch. No idea where that comes from though.

SR4ven:
@jtpereyda Completed with #245 and #249

PR #245, opened 2019-02-26
Added error checking for docs using travis-ci

SR4ven:
This should add some light error checking for making docs to CI. #243 
At some point we should use `SPHINXOPTS    = -nTW` instead to make this test fail on any warning, right now this wouldn't really be possible as there are just too many warnings. `-W` treats warnings as errors thus making the test fail. `-n` runs it in nit-picky mode which generates warnings on missing references.
See https://www.sphinx-doc.org/en/master/man/sphinx-build.html
I also updated the version number to the current one.


SR4ven:
Right now Sphinx fails to import boofuzz. We have to install it from source first.
Also this should definetly fail the docs building process which it doesn't. Maybe the `-W` flag is mandatory here

SR4ven:
Sorry for all those commits but I didn't find an easy way to test this locally.
There are two stages in travis now, one for tests and one for deployment, which is being skipped if `TRAVIS_TAG` is blank. The `on:` part under `deploy:` should now be redundant but I wasn't brave enough to remove it just yet.
Also all Sphinx warnings are now treated as errors. This results in the latest check failing as the docs in the current state are containing 9 warnings. The `-W` flag seems to be a must here as Sphinx will hardly fail at all as seen above where it failed to import boofuzz but still succeded building.
So before merging this the docs need to be fully fixed.

jtpereyda:
@SR4ven Nice work! If it's any help, I think the release process will fail if the version number hasn't been change (PyPI doesn't permit changing of deployed releases IIRC), so it should be sort of safe. :P

I totally forgot about the version number in docs/conf.py. I'll have to add that to the release process in CONTRIBUTING.py.

jtpereyda:
Looking at the current exception in Travis:
```
SphinxWarning: /home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/boofuzz/serial_connection.py:docstring of boofuzz.SerialConnection:32:Unknown target name: "data".
```
Looks like this might be from line 38 in the file:
```
        message_separator_time (float): After message_separator_time seconds _without receiving any more data_,
```
The two underscores are Markdown-ish for emphasis/italics. It looks like RST uses asterisks for this purpose: http://docutils.sourceforge.net/docs/user/rst/quickref.html

SR4ven:
Thanks for the hint on that formatting error!
As all modules but Session and Target in the source folder weren't included in any toctree, I added a new page called Other Modules to the API-Docs toctree containing all those modules. Not sure if this is the prefered structure here but they should definetly be included in some toctree.

Also, what exactly is the point of this deploy condition?
```
on:
  tags: true
  repo: jtpereyda/boofuzz
  branch: master
```
Because right now PRs to the master branch of this repo will match both branch and repo condition as their base branch is exactly this. It's only the tags flag that currently prevents the deployment from firing.
If the goal was to prevent all PRs from triggering a deploy something like this `(NOT type IN (pull_request)) AND (branch = master)` and optinally include the tag flag too so you still have manual control over it.

jtpereyda:
"Other Modules" sounds like a reasonable solution for now.

The key aspects of the deploy condition are:
1. Must be on the master branch
2. Must be a commit with a tag

Github uses tags to track releases, so this way a deploy will only occur when a new tag/release is added from Github. This lets us do several PRs before deciding to tag a release and thus trigger the deploy to PyPI.

SR4ven:
Ok, so the current deploy stage condition should cover that. I couldn't test if a release tag on Github really does trigger the stage but I don't see why it shouldn't. Also the deploy condition redundancy is no longer present.
I think this is ready to merge if that's fine by you.

jtpereyda:
Thank you again @SR4ven !

Issue #194, opened 2018-06-30
Switch ports when multiple instances running


jtpereyda:
boofuzz will choke when running two instances because it tries to reuse the web port. Instead:
1. Add logic to search and grab an unused port
2. Allow the URL to be printed for the user's convenience

To do so:
1. Separate the server startup from the fuzz start.

starek4:
I believe this is already implemented and we can close it.

SR4ven:
> 1. Add logic to search and grab an unused port
> 2. Allow the URL to be printed for the user's convenience

Are done.

I'm not sure about
> Separate the server startup from the fuzz start.

but we already have [Session.server_init()](https://boofuzz.readthedocs.io/en/stable/source/Session.html#boofuzz.Session.server_init)

@jtpereyda are all tasks completed?

jtpereyda:
Good catch, thanks @starek4 @SR4ven

PR #246, opened 2019-02-28
Enabled multiple instances

driechers:
Added logic to look for another port for the web interface.
Added option to pass None to web_port to disable the web interface.
Added log to print url for the web interface.

Resolves issue #194  

Signed-off-by: Daniel Riechers <driechers101@gmail.com>


boofuzz/sessions.py
```diff
@@ -365,7 +365,8 @@ def __init__(self, session_filename=None, index_start=1, index_end=None, sleep_t
         self._skip_current_element_after_current_test_case = False
         self._post_test_case_methods = []
 
-        self.web_interface_thread = self.build_webapp_thread(port=self.web_port)
+        if self.web_port is not None:
+            self.web_interface_thread = self.build_webapp_thread(port=self.web_port)
```
jtpereyda: Did you run into a case where self.web_port is None? Seems like it should generally be set to something. Was the intention to allow one to set web_port=None to disable the web server?

boofuzz/sessions.py
```diff
@@ -1063,7 +1065,14 @@ def transmit_fuzz(self, sock, node, edge, callback_data):
     def build_webapp_thread(self, port=constants.DEFAULT_WEB_UI_PORT):
         app.session = self
         http_server = HTTPServer(WSGIContainer(app))
-        http_server.listen(port)
+        while True:
+            try:
+                http_server.listen(port)
+            except:
```
jtpereyda: Catch the specific exception by name to avoid silently masking other errors. E.g. `except SocktException:` (I don't know what the actual exception name is off hand)

jtpereyda:
@driechers Thank you for the PR!

boofuzz/sessions.py
driechers: Correct. The intention is to allow a user to disable the web interface.

boofuzz/sessions.py
SR4ven: Something like this should work:
``` python
from socket import error as socket_error
...
 def build_webapp_thread()
...
    except socket_error as s_err:
        if s_err.errno != 98: # 98 = Address already in use.
            raise s_err # reraise if other error
        port += 1
```
This would currently fail as the db_logger can't handle any logging happenign prior to a first test case open.
To fix this change `self._current_test_case_index = None` to `self._current_test_case_index = 0` in `fuzz_logger_db.py:45`

boofuzz/sessions.py
```diff
@@ -10,6 +10,7 @@
 import time
 import traceback
 import zlib
+import socket
```
SR4ven: Importing full `socket` seems a little overkill here.
I'd suggest using something like `from socket import error as socket_error` instead.

jtpereyda: It's a bit of a coding style question -- AFAIK there's no performance difference in this case. Either way it has to import socket, it's just a question of namespaces. I personally prefer the `import socket` style as `socket.error` in the code makes it very easy/consistent to tell where it came from.

jtpereyda: Either way is fine IMO, unless I'm wrong about the performance aspect.

driechers: No problem I already have the change made and just need to rebase.

boofuzz/sessions.py
```diff
@@ -1063,7 +1066,17 @@ def transmit_fuzz(self, sock, node, edge, callback_data):
     def build_webapp_thread(self, port=constants.DEFAULT_WEB_UI_PORT):
         app.session = self
         http_server = HTTPServer(WSGIContainer(app))
-        http_server.listen(port)
+        while True:
+            try:
+                http_server.listen(port)
+            except socket.error as exc:
+                # Only handle "Address already in use"
+                if exc.args[0] != 98:
```
jtpereyda: For code readability I'd recommend using the `errno` module: https://docs.python.org/2/library/errno.html#errno.EADDRINUSE

Something like:
```python
            if exc.errno != errno.EADDRINUSE:
```
and of course `import errno` at the top. And then the comment can be removed as it is implied in the code.

boofuzz/fuzz_logger_db.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, db_filename, num_log_cases=0):
         self._db_cursor.execute(
             '''CREATE TABLE steps (test_case_index integer, type text, description text, data blob, timestamp TEXT)''')
 
-        self._current_test_case_index = None
+        self._current_test_case_index = 0
```
jtpereyda: @driechers for the record, did `None` result in an error that you saw?

driechers: No exceptions but the log for the url does not show up in the dB unless it is set to 0.

jtpereyda: oops +1 not -1 😆

jtpereyda:
Thank you for the PR @driechers !

Issue #239, opened 2019-02-19
Custom target restart method


SR4ven:
I was looking for a way to add a custom restart method for a target but could only find the EventHook class, which doesn't seem to be fully implemented yet. Is there a way to make use of it somehow?
More precisely I ways looking for a way to simply ping the target with an ICMP-Request to see when it is back up. If you are interessted in having this feature in the project, I could try to get that to work too.
Thinking about it, two different aproaches come to my mind:

1. Check if importing scapy is possible and simply use one of it's ping functions (ARP/ICMP/TCP/UDP).
2. Implenting it from scrap, building an ICMP-Request packet, sending it and waiting for a response.

Either way the targets IP would be required and would have to be entered and passed to the Session class somehow. Maybe reuse l2_dst from the SocketConnection class?

jtpereyda:
@SR4ven It would be great to have a generic way to add a custom restart method. Some brainstorming on the details:

The most generic facility would be one or both of:
1. Execute an external command to do restart.
2. Call a custom restart method written in Python.

The integration points could be:
1. Add a mechanism to register restart methods with the `Session` class. (See the `register_post_test_case_callback` and `_post_send` methods for an example*)
2. Add a mechanism to the procmon system. This would essentially be a `restart_commands` option akin to `start_commands`

The procmon integration is a natural fit with the external command, and the Session integration is a natural fit with the custom Python method.

* The `_post_send` method is a bit cluttered. And honestly it might make more sense to pass a restart method (or list of restart methods) into the constructor rather than making a special `register..` method that simply appends an item to a list.

jtpereyda:
As for how to implement a ping method in Python, you can Google around for `Python ping.py` to see some options out there. There is at least one ping package that uses raw sockets (I haven't tested it personally), and my Python install also comes with a ping.py script that uses impacket.

SR4ven:
Executing an external command would be really nice but I've done some tests with `subprocess` and I've noticed a very anoying problem described [here](https://stackoverflow.com/questions/1367373/python-subprocess-popen-oserror-errno-12-cannot-allocate-memory). Calling `subprocess.popen()` is requesting once more as much memory as Python is already eating up.
This is very bad when running a big fuzzing session that can use multiple Gb of RAM as calling popen() will crash with `OSError: [Errno 12] Cannot allocate memory`.
I don't know if Python 3 handles that in another way but with Python 2 I don't think `subprocess` is an option and I don't know of any other save methods to execute commands.

[...]

PR #248, opened 2019-03-02
Add options for callback functions to Sessions.py

SR4ven:
- Added pre_send, restart_target callbacks plus exception handling
- Modified post_send callback (backward compatible)
- Fixed error in fuzz_logger_db when trying to force write without log lines to be written
- Minor changes to some expressions
- Implementation of #239

### DIFF
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -278,11 +278,11 @@ class Session(pgraph.Graph):
         crash_threshold_request (int):  Maximum number of crashes allowed before a request is exhausted. Default 12.
         crash_threshold_element (int):  Maximum number of crashes allowed before an element is exhausted. Default 3.
         restart_sleep_time (int): Time in seconds to sleep when target can't be restarted. Default 5.
-        restart_callback (list of method): The registered method will be called after a failed post_test_case_callback
+        restart_callbacks (list of method): The registered method will be called after a failed post_test_case_callback
                                            Default None.
-        pre_send_callback (list of method): The registered method will be called prior to each fuzz request.
+        pre_send_callbacks (list of method): The registered method will be called prior to each fuzz request.
                                             Default None.
-        post_test_case_callback (list of method): The registered method will be called after each fuzz test case.
+        post_test_case_callbacks (list of method): The registered method will be called after each fuzz test case.
                                                   Default None.
         web_port (int):         Port for monitoring fuzzing campaign via a web browser. Default 26000.
         fuzz_data_logger (fuzz_logger.FuzzLogger): DEPRECATED. Use fuzz_loggers instead.
@@ -321,9 +321,9 @@ class Session(pgraph.Graph):
                  crash_threshold_request=12,
                  crash_threshold_element=3,
                  restart_sleep_time=5,
-                 restart_callback=None,
-                 pre_send_callback=None,
-                 post_test_case_callback=None,
+                 restart_callbacks=None,
+                 pre_send_callbacks=None,
+                 post_test_case_callbacks=None,
                  fuzz_data_logger=None,
                  fuzz_loggers=None,
                  fuzz_db_keep_only_n_pass_cases=0,
@@ -371,11 +371,22 @@ class Session(pgraph.Graph):
         self._receive_data_after_fuzz = receive_data_after_fuzz
         self._skip_current_node_after_current_test_case = False
         self._skip_current_element_after_current_test_case = False
-        self._pre_send_methods = pre_send_callback
-        self._post_test_case_methods = post_test_case_callback
-        self._restart_methods = restart_callback
-
         self.web_interface_thread = self.build_webapp_thread(port=self.web_port)
+
+        if pre_send_callbacks is None:
+            self._pre_send_methods = []
+        else:
+            self._pre_send_methods = pre_send_callbacks
+
+        if post_test_case_callbacks is None:
+            self._post_test_case_methods = []
+        else:
+            self._post_test_case_methods = post_test_case_callbacks
+
+        if restart_callbacks is None:
+            self._restart_methods = []
+        else:
+            self._restart_methods = restart_callbacks
 
         self.total_num_mutations = 0
         self.total_mutant_index = 0
@@ -671,7 +682,7 @@ class Session(pgraph.Graph):
             self.export_file()
             raise
         except exception.BoofuzzCallbackError:
-            self._fuzz_data_logger.log_error("Custom callback function raised uncaught exception, exiting.")
+            self._fuzz_data_logger.log_error("A custom callback function raised an uncaught exception, exiting.")
             self.export_file()
         except exception.BoofuzzTargetConnectionFailedError:
             # exception should have already been handled but rethrown in order to escape test run
@@ -867,9 +878,6 @@ class Session(pgraph.Graph):
         Args:
             method (function): A method with the same parameters as :func:`~Session.post_send`
             """
-        if self._post_test_case_methods is None:
-            self._post_test_case_methods = []
-
         self._post_test_case_methods.append(method)
 
     # noinspection PyUnusedLocal
@@ -906,7 +914,7 @@ class Session(pgraph.Graph):
             target (session.target): Target we are sending data to
         """
 
-        if self._pre_send_methods is not None:
+        if len(self._pre_send_methods) > 0:
             try:
                 for f in self._pre_send_methods:
                     self._fuzz_data_logger.open_test_step('Pre_Send callback: "{0}"'.format(f.__name__))
@@ -914,12 +922,8 @@ class Session(pgraph.Graph):
             except Exception:
                 self._fuzz_data_logger.log_error(constants.ERR_CALLBACK_FUNC.format(func_name="pre_send")
                                                  + traceback.format_exc())
-                raise exception.BoofuzzCallbackError()
-            finally:
-                self._fuzz_data_logger.open_test_step('Cleaning up connections from callbacks')
                 target.close()
-        else:
-            pass
+                raise exception.BoofuzzCallbackError()
 
     def restart_target(self, target):
         """
@@ -953,7 +957,7 @@ class Session(pgraph.Graph):
             time.sleep(3)
 
         # if we have custom restart methods, execute them
-        elif self._restart_methods is not None:
+        elif len(self._restart_methods) > 0:
             try:
                 for f in self._restart_methods:
                     self._fuzz_data_logger.open_test_step('Target restart callback: "{0}"'.format(f.__name__))
@@ -1405,7 +1409,7 @@ class Session(pgraph.Graph):
         return "{0}.{1}.{2}".format(message_path, primitive_under_test, self.fuzz_node.mutant_index)
 
     def _post_send(self, target):
-        if self._post_test_case_methods is not None:
+        if len(self._post_test_case_methods) > 0:
             try:
                 for f in self._post_test_case_methods:
                     self._fuzz_data_logger.open_test_step('Post- test case callback: "{0}"'.format(f.__name__))
@@ -1421,7 +1425,6 @@ class Session(pgraph.Graph):
             except Exception:
                 self._fuzz_data_logger.log_error(constants.ERR_CALLBACK_FUNC.format(func_name="post_send")
                                                  + traceback.format_exc())
-                raise exception.BoofuzzCallbackError()
             finally:
                 self._fuzz_data_logger.open_test_step('Cleaning up connections from callbacks')
                 target.close()
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import
 
 import cPickle
 import datetime
-import itertools
 import logging
 import os
 import re
@@ -119,9 +118,9 @@ class Target(object):
         Returns:
             Received data.
         """
-        if max_bytes == None:
+        if max_bytes is None:
             max_bytes = self.max_recv_bytes
-        
+
         if self._fuzz_data_logger is not None:
             self._fuzz_data_logger.log_info("Receiving...")
 
@@ -279,6 +278,12 @@ class Session(pgraph.Graph):
         crash_threshold_request (int):  Maximum number of crashes allowed before a request is exhausted. Default 12.
         crash_threshold_element (int):  Maximum number of crashes allowed before an element is exhausted. Default 3.
         restart_sleep_time (int): Time in seconds to sleep when target can't be restarted. Default 5.
+        restart_callback (list of method): The registered method will be called after a failed post_test_case_callback
+                                           Default None.
+        pre_send_callback (list of method): The registered method will be called prior to each fuzz request.
+                                            Default None.
+        post_test_case_callback (list of method): The registered method will be called after each fuzz test case.
+                                                  Default None.
         web_port (int):         Port for monitoring fuzzing campaign via a web browser. Default 26000.
         fuzz_data_logger (fuzz_logger.FuzzLogger): DEPRECATED. Use fuzz_loggers instead.
         fuzz_loggers (list of ifuzz_logger.IFuzzLogger): For saving test data and results.. Default Log to STDOUT.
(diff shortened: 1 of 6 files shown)


boofuzz/sessions.py
```diff
@@ -1381,8 +1419,9 @@ def _post_send(self, target):
             except exception.BoofuzzTargetConnectionFailedError:
                 self._fuzz_data_logger.log_fail(constants.ERR_CONN_FAILED)
             except Exception:
-                self._fuzz_data_logger.log_fail(
-                    "Custom post_send method raised uncaught Exception." + traceback.format_exc())
+                self._fuzz_data_logger.log_error(constants.ERR_CALLBACK_FUNC.format(func_name="post_send")
+                                                 + traceback.format_exc())
+                raise exception.BoofuzzCallbackError()
```
jtpereyda: Raising an exception again here will cause the entire test to halt due to an error in the callback function. I originally chose to just log the test case as a failure and keep testing. The ideal choice may depend on the situation -- thoughts?

boofuzz/sessions.py
```diff
@@ -363,7 +371,9 @@ def __init__(self, session_filename=None, index_start=1, index_end=None, sleep_t
         self._receive_data_after_fuzz = receive_data_after_fuzz
         self._skip_current_node_after_current_test_case = False
         self._skip_current_element_after_current_test_case = False
-        self._post_test_case_methods = []
+        self._pre_send_methods = pre_send_callback
+        self._post_test_case_methods = post_test_case_callback
+        self._restart_methods = restart_callback
```
jtpereyda: Should these be lists? They seem to be treated as iterable later on.

boofuzz/sessions.py
jtpereyda: The three callbacks all use similar code. Could we factor it out into one function? E.g.:
```python
def _execute_callback_list(callbacks, name):
    ...

...

def _post_send(self, target):
    ...
    self._execute_callback_list(self._post_test_case_methods, 'post- test case')
```

jtpereyda:
By the way, have you tested the functionality?

boofuzz/sessions.py
SR4ven: They should be yes, but I thought checking for None is a little more performant than calling len().
The main reason I went for this approach is, that setting a default value to [] triggers a warning in PyCharm. I can't remember which one exactly, will look it up.

boofuzz/sessions.py
```diff
@@ -660,6 +670,9 @@ def _main_fuzz_loop(self, fuzz_case_iterator):
             self._fuzz_data_logger.log_error("Restarting the target failed, exiting.")
             self.export_file()
             raise
+        except exception.BoofuzzCallbackError:
+            self._fuzz_data_logger.log_error("Custom callback function raised uncaught exception, exiting.")
```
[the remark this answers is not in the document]
SR4ven: Will change

boofuzz/sessions.py
```diff
@@ -889,47 +903,75 @@ def pre_send(self, sock):
...
+                    f(target=target, fuzz_data_logger=self._fuzz_data_logger, session=self, sock=target)
+            except Exception:
+                self._fuzz_data_logger.log_error(constants.ERR_CALLBACK_FUNC.format(func_name="pre_send")
+                                                 + traceback.format_exc())
+                raise exception.BoofuzzCallbackError()
+            finally:
+                self._fuzz_data_logger.open_test_step('Cleaning up connections from callbacks')
+                target.close()
+        else:
+            pass
```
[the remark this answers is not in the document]
SR4ven: Will remove, it was only temporary and I forgot about it.

boofuzz/sessions.py
SR4ven: My thoughts about this were that when an unhandled exception is raised, there is something going wrong in the callback function that shouldn't happen. Just continuing in this case might result in the user not noticing that there is a fault in the callback function.
You can still raise some of the cought exceptions above to let boofuzz know of connection problems.
Just a thought though, it's up to you how this is handled.

SR4ven: I thought about it but they do differ in some points like exception handling.
I will try to figure a way out.

SR4ven:
Yes it all works just as I expected.

boofuzz/sessions.py
SR4ven: So this is the soft warning you get in PyCharm when setting default values to list type.
`Inspection info: This inspection detects when a mutable value as list or dictionary is detected in a default value for an argument. Default argument values are evaluated only once at function definition time, which means that modifying the default value of the argument will affect all subsequent calls of the function.`
The argument stated there isn't really valid for an empty list as that value won't be modified.
So using [] should be possible in this case but currently I don't get why setting them to None would be an issue or let's say not best practice.

boofuzz/sessions.py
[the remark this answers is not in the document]
SR4ven: I just noticed that this would require `exception.BoofuzzCallbackError` to take the name of the failing callback as an argument. Not that this would be a problem but if we do that we will get the same line twice in the log.
This is how it currently looks like:
```
[2019-03-04 17:49:56,334] Test Case: 1: Test.verbs.1
[2019-03-04 17:49:56,334]     Info: Type: Group. Default value: 'GET'. Case 1 of 384 overall.
[2019-03-04 17:49:56,334]     Info: Opening target connection (127.0.0.1:80)...
[2019-03-04 17:49:56,334]     Info: Connection opened.
[2019-03-04 17:49:56,334]   Test Step: Fuzzing Node 'Test'
[2019-03-04 17:49:56,335]     Info: Sending 8 bytes...
[2019-03-04 17:49:56,336]     Transmitted 8 bytes: 47 45 54 20 0d 0a 0d 0a 'GET \r\n\r\n'
[2019-03-04 17:49:56,336]     Info: Closing target connection...
[2019-03-04 17:49:56,336]     Info: Connection closed.
[2019-03-04 17:49:56,336]   Test Step: Post- test case callback: "ping"
[2019-03-04 17:49:56,336]     Info: Ping Pong
[2019-03-04 17:49:56,337]     Error!!!! A custom post_send callback function raised an uncought error. Stopping test run.
                                File "/home/dev/PycharmProjects/boofuzz/boofuzz/sessions.py", line 1412, in _post_send
                                  f(target=target, fuzz_data_logger=self._fuzz_data_logger, session=self, sock=target)
                                File "/home/dev/PycharmProjects/boofuzz/requests/http.py", line 17, in ping
                                  a = 5 / 0
                              ZeroDivisionError: integer division or modulo by zero
                              
[2019-03-04 17:49:56,338]   Test Step: Cleaning up connections from callbacks
[2019-03-04 17:49:56,338]     Info: Closing target connection...
[2019-03-04 17:49:56,338]     Info: Connection closed.
[2019-03-04 17:49:56,339]     Error!!!! Custom callback function raised uncaught exception, exiting.
```
Using the `ERR_CALLBACK_FUNC`, the last line would change to
```
[2019-03-04 17:49:56,339]     Error!!!! A custom post_send callback function raised an uncought error. Stopping test run.
```
If we keep the current one, I guess `Error!!!! A custom callback function raised an uncaught exception, exiting.` would be a little better, but that's all up to you.

boofuzz/sessions.py
SR4ven: Looking at the code, the only part I see that we could factor out would be:
```python
for f in self._post_test_case_methods:
    self._fuzz_data_logger.open_test_step('{1} callback: "{0}"'.format(f.__name__), test_step_name)
    f(target=target, fuzz_data_logger=self._fuzz_data_logger, session=self, sock=target)
```
The exception handling takes up the largest part of those functions but we can't put that in a generic callback method. Not that I see a way to do so, at least.

If you like I can still put those 3 lines in an extra function.

boofuzz/sessions.py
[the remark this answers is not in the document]
jtpereyda: OK, I didn't think about that complexity of passing the information up the exception chain. Looks good as is.

boofuzz/sessions.py
jtpereyda: @SR4ven Thanks for your thoughts. It's a close call, but in this case let's toward supporting long-running tests. Logging the error will result in an error/fail test case which should be noticeable (if it's not noticeable then the UI should be changed).

So let's take out the exception raise for now, and keep the log_error as is. Thanks!

boofuzz/sessions.py
jtpereyda: > I thought checking for None is a little more performant than calling len().
The performance difference should be fairly negligible compared to what else happens in a fuzz test. See https://softwareengineering.stackexchange.com/questions/80084/is-premature-optimization-really-the-root-of-all-evil

> setting a default value to [] triggers a warning in PyCharm
This is a good reason, and the warning is valid. It's definitely a language corner case. For an extended discussion, see: https://stackoverflow.com/questions/9158294/good-uses-for-mutable-function-argument-default-values

Now that I look at it again, I see you documented that the arguments to Session should be lists themselves:
```
        pre_send_callback (list of method): The registered method will be called prior to each fuzz request.
                                            Default None.
```

This would be clearer if the argument names were plural. `pre_send_callbacks` instead of `pre_send_callback`, etc.

As for the default value situation, the way one can resolve this is to keep the default value at None, and add some code inside `__init__`:
```
        if pre_send_callbacks is None:
            pre_send_callbacks = []
        self._pre_send_methods = pre_send_callbacks
```

This makes the constructor more bulky, but has the significant advantage of giving maintainers one less thing to worry about -- `self._pres_send_methods` is always a list. No checks for None.

Requested changes:
1. Pluralize Session argument names.
2. (If you have time to change and test) Initialize callback lists to empty lists by default, and remove the None checks later in the code.

boofuzz/sessions.py
jtpereyda: @SR4ven I think it would be good if more of the code were the same anyway -- e.g., it makes sense for `BoofuzzTargetConnectionReset` to be handled the same way no matter which kind of callback is in play... at least I think it makes sense.

That said, this code cleanliness recommendation can be tabled for another time. No need to work through it right now.

jtpereyda:
@SR4ven comments addressed with a couple requests still active. Thanks for slogging through the initial PR and the changes!

SR4ven:
Two more quetions:

1. Is it still correct that `pre_send()` and `restart_target()` are marked as "public" and `_post_send()` as "private"? Talking about the preceding `_`
2. The `pre_send()` method is called after opening a connection but before creating the fuzzed data. Wouldn't it be better to first generate the fuzzed data, then call the `pre_send()` method and after that send it. That way it would be possible to manipulate the data just before sending it.
Possible scenarios I can think of are generating custom checksums or generally something to make the data valid again.

boofuzz/sessions.py
```diff
@@ -373,13 +373,27 @@ def __init__(self, session_filename=None, index_start=1, index_end=None, sleep_t
         self._receive_data_after_fuzz = receive_data_after_fuzz
         self._skip_current_node_after_current_test_case = False
         self._skip_current_element_after_current_test_case = False
-        self._pre_send_methods = pre_send_callback
-        self._post_test_case_methods = post_test_case_callback
-        self._restart_methods = restart_callback
+
+        self._post_test_case_methods = []
```
jtpereyda: looks like this bit is redundant with lines 387-390.

boofuzz/sessions.py
```diff
@@ -910,20 +921,16 @@ def pre_send(self, target):
...
             except Exception:
                 self._fuzz_data_logger.log_error(constants.ERR_CALLBACK_FUNC.format(func_name="pre_send")
                                                  + traceback.format_exc())
-                raise exception.BoofuzzCallbackError()
-            finally:
-                self._fuzz_data_logger.open_test_step('Cleaning up connections from callbacks')
                 target.close()
-        else:
-            pass
+                raise exception.BoofuzzCallbackError()
```
jtpereyda: Remove exception raise, just like in `_post_send`. Do the same for the restart methods section.

jtpereyda:
> Is it still correct that pre_send() and restart_target() are marked as "public" and _post_send() as "private"? Talking about the preceding _

You are right, I believe it makes the most sense to make them private by adding the underscore.

> The pre_send() method is called after opening a connection but before creating the fuzzed data. Wouldn't it be better to first generate the fuzzed data, then call the pre_send() method and after that send it. That way it would be possible to manipulate the data just before sending it.
Possible scenarios I can think of are generating custom checksums or generally something to make the data valid again.

This is an interesting idea. Unless it will be a super simple change (like one line), I'd recommend making it a separate PR.

I think there are some complexities, especially in a multi-message session -- `pre_send` happens prior to any messages being sent IIRC, so you could render the first message but not necessarily the others. There is also a per-message callback.

boofuzz/sessions.py
SR4ven: Thanks for the hint, that one must have sneaked in when I merged the upstream changes.

boofuzz/sessions.py
SR4ven: Will change.

SR4ven:
Honestly, I feel really uncomfortable with just catching any exception coming from custom callback functions.

Let's say your defined restart target function fails and the target won't get restarted. Boofuzz would only log that and continue testing, even though the target won't be able to handle the sent data.
If you have a working post_send check, it would detect that the target is down, hence logging the following test cases as possible failures.
That would lead to missleading logs and test cases possibly going into the emty.

Also, an unexpected exception in a custom post_send function doesn't trigger a restart. This might result in the target being down, your callback function somehow detecting that thus raising an error, but boofuzz just logs it and keeps testing like nothing happened.

This might be fine if your callbacks work 100% of the time and you allways have someone observing the logs, but for e.g. CI integration that's not really the desired behavior.

Again, that's just my point of view and the decision is up to you.

jtpereyda:
@SR4ven I hear what you're saying. Are these concerns that you're seeing your current use cases, or just what you're thinking about how others may use it?

SR4ven:
@jtpereyda Both actually. The thing is that I'm not the person running boofuzz most of the time. I'm just writing the definitions and I'm trying to make it as easy as possible to use.
The people actually using the fuzzer, can't and shouldn't have to observe the log output the whole time to check whether something went wrong. And I just can't guarantee that the callbacks always work flawlessly 100% of the time.
The worst case would be someone running the test, instrumentation callbacks failing, hence not detecting failures of the SUT, but displaying the test as successful. If the error was still persistent on the last test case I could have a chance of the person reporting that error to me, if it only occurred a few times somewhere in the middle, none would ever see it.
If I was running the fuzzer, I would agree with you. But I see possible issues with users running this tool that don't exactly know how it works.

The current state of this PR should suit your requests. If you don't agree with me, I'm totally fine with that and would just change the error handling locally.

jtpereyda:
@SR4ven 
> The worst case would be someone running the test, instrumentation callbacks failing, hence not detecting failures of the SUT, but displaying the test as successful.

Don't errors result in an apparent failure? I'm looking at this excerpt from fuzz_logger_db.py:
```
    def log_error(self, description):
        self._queue.append(["INSERT INTO steps VALUES(?, ?, ?, ?, ?);\n", self._current_test_case_index, 'error',
                            description, b'', helpers.get_time_stamp()])
        self._fail_detected = True
        self._write_log()
```

If errors look like failures, would that work for your use case?

SR4ven:
@jtpereyda If an error during a callback function is logged, it will appear in that specific test case log but never again. With tens of test cases per second you will lever notice that happening if the test just continues.

Logging unhandled callback errors as test case failures would at least prevent the worst case I described. The user would then be looking at what went wrong during that test case and notice that it was a callback error.
Would you like me to change that?

jtpereyda:
@SR4ven I like how that looks! A few notes:

1. Ideally, we could still log the cases as "errors" but then display them to the user in a way that looks much more like a failure. See `FuzzLoggerDbReader.failure_map` -- we could change the query line to something like:

    ```python
            failure_steps = c.execute('''SELECT * FROM steps WHERE type="fail" or type="error"''')
    ```

    This allows us to keep flexibility on how we handle the errors, how we display them in the future, etc. I believe a change like this (some accompanying changes may be necessary) would get the error to show up in the failure but still say "error" in the log.
1. If we go that route, let's use a separate PR if that works. If only for my sake so I can keep track of 
what's going on. :)

Another thought: If killing the test is still useful for you, you could also add an option to `Session` to kill the test on unhandled callback exceptions. This would allow one to enable a sort of "strict" mode. If the option becomes extremely popular, we could even change the default in the future.

jtpereyda:
@SR4ven Thank you very much for the quality work and for working through the feedback, use cases, etc. Feel free to open a PR or issue to implement the changes you've mentioned and/or continue the discussion. :)

PR #249, opened 2019-03-03
Fix setup.py

SR4ven:
Impacket requires both ldap3 and ldapdomaindump, while ldapdomaindump requires ldap3==2.5.1.
Apparently ldap3 2.5.2 is installed before knowing of this requirement so the installation fails.

This should only be a temporary fix, we might want to consider removing impacket completely.
The only occurence of it is in network_monitor.py which is marked as depreciated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,6 @@ sudo: required
 
 before_install: sudo sed -i -e 's/^Defaults\tsecure_path.*$//' /etc/sudoers
 
-install:
-  - pip install tox sphinx
-  - pip install .
-
 stages:
   - Test
   - name: Deploy
@@ -16,8 +12,12 @@ stages:
 jobs:
   include:
     - stage: Test
+      install: pip install tox .
+      script: tox --recreate
+    - # Same stage name
+      install: pip install sphinx
       script:
-        - tox --recreate
+        - python setup.py install --force
         - make -C ./docs dummy
     - stage: Deploy
       script: skip
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,10 @@ stages:
 jobs:
   include:
     - stage: Test
-      install: pip install tox .
-      script: tox --recreate
-    - # Same stage name
-      install: pip install sphinx
+      install: pip install tox sphinx .
       script:
-        - python setup.py install --force
+        - pip check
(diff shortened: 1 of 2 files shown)


SR4ven:
CI now spawns two tests in parallel to check both install methods `pip install .` and `python setup.py install --force`.
There is also no longer an installation in place for the deploy stage. I take it deploying to PyPi doesn't require boofuzz to be installed.

SR4ven:
The latest pytest version works just fine. It even seems to perform more in depth testing as there are several test case warnings now too.

jtpereyda:
@SR4ven A few questions...

I take it this is from the Read The Docs failure... Why did it fail in RTD but not on Travis? Ideally we could identify the reason for this discrepancy and thereby clarify the root cause.

Do we know that this fixes things in RTD?

Why test `python setup.py install --force`?

SR4ven:
Travis did not fails as it was installing boofuzz with pip only. RTD is using the setup.py install method which seems to handle dependencies less intelligently somehow. (See my first message here)
RTD failed at exactly that `python setup.py install --force`. Testing with force shouldn't be necessary as the environment should be cleared at every test. I thought testing this would be nice to see if both ways of installing boofuzz work and be notified of any dependencie issues coming up with external updates. This PR will also fix the RTD issue, yes. You can test this by simply trying to install boofuzz with the setup.py before and after the changes from this PR. (You should have a clean python env)

SR4ven:
I just installed boofuzz from soure using `pip install -e . --user` and when calling `boo open` I get the following error:
```
pkg_resources.ContextualVersionConflict: (ldap3 2.5.2 (/home/papgft/.local/lib/python2.7/site-packages), Requirement.parse('ldap3==2.5.1'), set(['ldapdomaindump']))
```
So aparently even the pip install method does have issues with the current setup. It just doesn't fail directly during the installation. Normal fuzzing works without any errors.

SR4ven:
Found the same issue on the `impacket` repo: https://github.com/SecureAuthCorp/impacket/issues/571
They didn't really fix it though, just provided a workaround for manual installation.

jtpereyda:
@SR4ven Thank you for the PRs! And the references -- that all makes the issue pretty clear.

PR #251, opened 2019-03-08
Fixed double close and post send behavior

driechersRC:
Moved target close to after post send callbacks and removed extra close.

Signed-off-by: Riechers, Daniel J <daniel.riechers@rockwellcollins.com>


jtpereyda:
Thank you @driechers !

Issue #254, opened 2019-03-10
Travis not deploying to pip


SR4ven:
With the recent changes to the travis config, the deployment to pip seems to be broken.
I will take a look at this as soon as I can.

SR4ven:
So I took a brief look at it and it seems to be an issue with the `branch = master` condition.
Looking at the travis log, the branch it is cloning is `v0.1.4`
```
git clone --depth=50 --branch=v0.1.4 https://github.com/jtpereyda/boofuzz.git jtpereyda/boofuzz
```
I have no idea why this worked with the previous deploy contion like this:
```YAML
  on:
    tags: true
    repo: jtpereyda/boofuzz
    branch: master
```
If the idea behind this condition was to only deploy if a GitHub tag is present and the commit is not coming from a fork, maybe this would do it:
```YAML
stages:
  - Test
  - name: Deploy
    if: (repo = jtpereyda/boofuzz) AND (fork = false) AND (tag IS present)
```
Sadly this is hard to test and I don't really know how to fix this for the current release. Maybe just manually deploy it.

jtpereyda:
Would there be a downside if we changed the deploy steps back to the previous way? It was a while ago that I wrote it, but I figure I was following some guidelines from the Travis documentation.

SR4ven:
Apart from cosmetics, the only little downside would be that you'd have to exclude all the files created during the tests, which you already did anyways. On the other hand, removing `skip_cleanup: true` would have the same effect, I don't see the need of it as we're not building or compiling anything, just testing. See [docs](https://docs.travis-ci.com/user/deployment#uploading-files-and-skip_cleanup)
I still don't understand why the previous way with `branch: master` works...

jtpereyda:
@sparticvs 

> I'm not 100% sure but I think the old deploy conditions would allow anyone to trigger a deploy by tagging their commits to a PR.
PRs cannot contain tags; see: https://stackoverflow.com/a/12279290/461834

> Also I don't know how one would protect himself against that, as you can always edit the travis.yml in a PR, or am I missing something here?
Well, a PR can contain literally any code changes, so yes that is a risk. Hence the need for code review & approval. Another thing that helps is keeping PRs simple enough to be easily understood so that no malicious activity can hide in the details. The PR system relies on trusted committers. Even when a trusted committer is the one making changes, code review by another trusted committer helps avoid accidental damage.

[...]

PR #255, opened 2019-03-10
Fix pypi autodeploy

SR4ven:
Addressing #254.
I'm not exactly sure if `fork = false` is what we need here. The [docs](https://docs.travis-ci.com/user/conditions-v1) aren't too detailed sadly. This condition might be redundant with `repo = jtpereyda/boofuzz`

The `branch = master` condition doesn't match your release workflow. You create a release prep branch, change things there and merge it back with as PR. So when travis runs, it sees that the branch is not master and skips the deploy stage.
So either we get rid of that condition or you have to push a last, possibly empty but tagged, commit directly to master to trigger future releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ before_install: sudo sed -i -e 's/^Defaults\tsecure_path.*$//' /etc/sudoers
 
 stages:
   - Test
-  - name: Deploy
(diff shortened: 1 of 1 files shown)


jtpereyda:
Thanks @SR4ven ! We'll see if this one works...

PR #258, opened 2019-03-12
Moved checks to tox, added syntax check

SR4ven:
- All checks previously performed on travis can now also be run locally with tox
- Added flake8 syntax checking for both critical errors and code-style warnings
- Updated [dev] dependencies

The idea behind this is to make it possible to easily perform the same checks as travis does locally with tox.
It would be nice to know how the equivalent to `make -C ./docs dummy` for Windows too.

I've already fixed one error in the ini.py but I could use some help fixing utils/ida_fuzz_library_extender.py which seems to be missing some dependencies. Checks will fail because of this.

@jtpereyda I also looked into ways to add codecov to the CI if you're interested. Personally, I don't like the fact that a bot comments the report after every commit to a PR. Would look something like [this](https://github.com/codecov/example-python/pull/26).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,8 @@ stages:
 jobs:
   include:
     - stage: Test
-      install: pip install tox sphinx .
-      script:
-        - pip check
-        - tox --recreate
-        - make -C ./docs dummy
+      install: pip install tox
+      script: tox --recreate
     - stage: Deploy
       script: skip
       deploy:
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -477,7 +477,7 @@ def s_mirror(primitive_name, name=None):
     :param name:            (Optional, def=None) Name of current primitive
     """
     if primitive_name not in blocks.CURRENT.names:
-        raise sex.SullyRuntimeError("CAN NOT ADD A MIRROR FOR A NON-EXIST PRIMITIVE CURRENTLY")
+        raise exception.SullyRuntimeError("CAN NOT ADD A MIRROR FOR A NON-EXIST PRIMITIVE CURRENTLY")
 
     mirror = primitives.Mirror(primitive_name, blocks.CURRENT, name)
     blocks.CURRENT.push(mirror)
--- a/docs/make.bat
+++ b/docs/make.bat
@@ -10,6 +10,7 @@ if "%SPHINXBUILD%" == "" (
 set SOURCEDIR=.
 set BUILDDIR=_build
 set SPHINXPROJ=boofuzz
+set SPHINXOPTS=-TW
 
 if "%1" == "" goto help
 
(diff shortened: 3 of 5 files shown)


jtpereyda:
@SR4ven Very cool!

The pre_commands don't seem to be running on my system:
```
> tox
GLOB sdist-make: /home/jtpereyda/code/boofuzz/setup.py
________________________________________________________ summary ________________________________________________________
SKIPPED:  py27-unix: platform mismatch
SKIPPED:  py27-windows: platform mismatch
  congratulations :)
```

The flake8 warnings look like they might be too aggressive at first glance -- running the commands results in failures currently. We can remove warning options until we're closer to success -- and mop up a few warnings if there are some stragglers.

SR4ven:
@jtpereyda you could try a tox --recreate, but apparently it doesn't detect python2.7 on your system. Are you on Windows or Linux?
I don't think F821 is too aggressive, I mean trying to execute that script would result in an error, as it is calling undefined classes/functions. I also couldn't find anything about what ida_fuzz_library_extender.py actually does, but I doubt anyone could put it into use in the current state. It is not used by boofuzz in any way as far as my search in PyCharm could tell.
If you say fixing that has lower priority, we could simply exclude that one file from the check.

jtpereyda:
oh oops should have checked my output better o_o

jtpereyda:
OK, my build is working now-- and failing with a few warnings.

> If you say fixing that has lower priority, we could simply exclude that one file from the check.

Sounds good.

jtpereyda:
@SR4ven any idea why there's no Travis build on this PR?

SR4ven:
@jtpereyda there is but GitHub has been acting up for me for the last few hours. You can see them if you look at the travis site directly. I'm currently trying to get it to exclude that one file.

BTW, you don't happen to know what the Windows equivalent to `make -C ./docs dummy` is?
Also are you interested in a codecov report? Would open another PR for that of course.

SR4ven:
I added full tox support for Windows, couldn't really test it all together though. Would be nice if someone else could verify that it's working.

@jtpereyda no idea why GitHub doesn't show Travis builds on this PR, but you can see the current one [here](https://travis-ci.org/jtpereyda/boofuzz/builds/505460975).

jtpereyda:
@SR4ven Merging! Thank you for the contribution -- it makes sense for all the tests to run from tox.

I noticed in the most recent Travis build that there are a lot of warnings, but they don't cause the build to fail. I take it this is the intended behavior.

Cheers!

SR4ven:
@jtpereyda The warnings are indeed intended. Once the current ones are eliminated, you can check for new warnings on upcoming PRs to check the code style. Not mandatory of course.
If you think it's worth it, open an issue about resolving the current warnings.

Issue #166, opened 2017-10-31
GUI - Keep UI persistent after test complete


jtpereyda:
Right now boofuzz quits as soon as it is done, which frees up the command line, but closes the HTTP interface, which is frustrating if you wanted to actually see the results.

Some options:
1. Spin off a process that lets you browse results after the test is complete. As a security consideration, would need to verify server only listens on localhost.
2. Have the command line app stay alive with a "Press any key to quit" or similar.

starek4:
I think that this is already implemented.

SR4ven:
Second option was implemented by #260

Issue #259, opened 2019-03-12
Memory optimisations


SR4ven:
When loading large protocol definitions like the http_header request, boofuzz uses incredible ammounts of memory (Somewhat around 10Gb). With current hardware that's not too much of an issue but, if possible, still worth optimising I think.
I initially found this to be an issue when implementing an option to keep the webinterface up after the test has finished. At that point all the generated fuzz-data is no longer required and should be picked up by the garbage collector. But that never happens, because there are still lots of references active. Even when deactivating the webinterface with `web_port=None` the memory is not freed, there are a few less references though.
More optimisation could also be done handling the way fuzz-data is being generated. In the current state all nodes are being generated during initialisation, when it would be much better to only generate the fuzz-data "on the go" for the node or even field that is currently being fuzzed.

Not having looked closer at the data generation, I'm sure this optimisation will be rather complex and a lot of work. Either way I think that it would be worth it as boofuzz would become much more lightweight and able to run extended protocol definitions in one go.

PR #260, opened 2019-03-13
Added keep_web_open session option

SR4ven:
Easiest way I found to implement #166.
Default value is True currently as I think that's the most user friendly. If anyone is using boofuzz in an automated environment, this will break the flow, but is easy to fix. Maybe add a warning to the changelog?

This can still be optimised, e.g. spin off another process and release the command line, but I'll leave that for another time. Thinking about it, how would you close the now seperated web-process? Also the next session would have to reconnect to that process so that there aren't more and more spawning. Seems to be quite complicated...

### DIFF
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -693,8 +693,6 @@ class Session(pgraph.Graph):
                 print("\nFuzzing session completed. Keeping webinterface up on localhost:{}".format(self.web_port),
                       "\nPress ENTER to close webinterface")
                 raw_input()
-            else:
-                print("\nFuzzing session completed. Webinterface will be available as long as this script is running.")
         except KeyboardInterrupt:
             # TODO: should wait for the end of the ongoing test case, and stop gracefully netmon and procmon
             self.export_file()
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
 
 import cPickle
 import datetime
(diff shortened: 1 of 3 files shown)


SR4ven:
@jtpereyda something seems to be wrong with Travis, still no status on the commits, but it is [building](https://travis-ci.org/jtpereyda/boofuzz/builds/506044910)
I also noticed that other repos have their checks logged in the [Checks](https://github.com/jtpereyda/boofuzz/pull/260/checks) tab, which we never had for some reason.

boofuzz/sessions.py
```diff
@@ -685,6 +688,13 @@ def _main_fuzz_loop(self, fuzz_case_iterator):
             self._fuzz_data_logger.close_test()
             if self._reuse_target_connection:
                 self.targets[0].close()
+
+            if self._keep_web_open and self.web_port is not None:
+                print("\nFuzzing session completed. Keeping webinterface up on localhost:{}".format(self.web_port),
+                      "\nPress ENTER to close webinterface")
+                raw_input()
+            else:
+                print("\nFuzzing session completed. Webinterface will be available as long as this script is running.")
```
jtpereyda: I'd lean toward omitting the else statement -- in the normal case the script will exit right after it's printed, right?

SR4ven: Yes, as long as the script ends after `sess.fuzz()`. The session itself won't close the web-thread as it is a daemon and will run until the main thread exits. My idea was to make the user aware of the web-thread still running as long as the script is too. I see that there are practically no cases where that could happen though, so I'm just going to remove it.

For #259 it might help to manually close the web-thread when fuzz() is completed.

PR #263, opened 2019-04-01
add len() for primitives.mirror

cq674350529:
To refine the test case for `s_mirror()` recommended in https://github.com/jtpereyda/boofuzz/pull/228#issuecomment-443479596
> 2. For test coverage you could add a test that checks the recursive checks. Something where the target of s_mirror is a length primitive (s_size), which is itself a length of a block containing s_mirror. But honestly that's getting a little crazy given the stability of boofuzz overall.

When adding this test case, an issue was found. 

1.  `self._value` is `None` in original implemention of `s_mirror()`. When calling `size.render()`, it will calculate each item's length in block. An error occurs since the `self._value` of `mirror` is `None`.

https://github.com/jtpereyda/boofuzz/blob/bad385117c4176c4dde770dab7cacccaed1c5b02/boofuzz/primitives/base_primitive.py#L89-L90

Since `mirror` doesn't have default value at the moment, When directly set `self._value` to its target's value, there would be some errors if the target of `s_mirror` is a length primitive (`s_size`).

So, just add an len() for `mirror`,  which gets its length based on its target's length. It works well in my environment.


jtpereyda:
@cq674350529 Cool, thank you! Please pardon my tardiness in reviewing.

Issue #191, opened 2018-06-23
Static Console Screen


jtpereyda:
It would be cool to display stats/general info on the console while running a test. Would be more helpful than spitting out the entire log on the console. Items to display:
1. URL to fuzz control
2. Summary stats (total test cases, portion done, etc.)
3. Info on current node?

driechers:
I've started playing around with some code for this and I'm looking for an opinion. I'm thinking the static part of the console could be the bottom row since and the rest of the screen can be the current log to console behavior. Unfortunately I can't  seem to get this to work with curses but I can get it working using raw vt100 escape codes. Before I go too far down the rabbit hole, does this seem like an acceptable behavior/approach?

```python
#!/usr/bin/python

import time
import sys
import readchar

def getWidth():
    sys.stdout.write("\033[999C\033[6n\033[999D")
    yx = ''
    c = readchar.readchar()
    while c != 'R':
        yx += c
        c = readchar.readchar()
    yx = yx[2:]
    y,x = yx.split(';')
    return int(x)

def test():
    w = getWidth()
    for i in range(120):
        # Print something and delete remaining chars on current line
        # This would be the exisitng logs printed to the console
        print "test\033[K"
        draw_progress(i, 119, w)
        time.sleep(0.03)

def draw_progress(complete, total, w):
    progress = 100*complete / total

    status = " %d/%d %d%%" % (complete, total, progress)

    numChars = (w-len(status)-2) * progress / 100
    blankChars = w - numChars - len(status) - 2
    if numChars != 0:
        sys.stdout.write('[' + '='*numChars + ' '*blankChars + ']')
        sys.stdout.write(status + '\r')
        sys.stdout.flush()

test()
print ''
```

SR4ven:
I took a look at this a little while ago and thought about adding an extra class using `curses`.
Currently I only have a non functional example of how it could look like. (I was too lazy to color the rest)
![boofuzz_static_clonsole](https://user-images.githubusercontent.com/46794237/53837264-7c9dc400-3f92-11e9-9d71-23c96d9bfc17.JPG)
```python
import curses


def draw_menu(stdscr):
    k = 0

    # Clear and refresh the screen for a blank canvas
    curses.curs_set(0)
    stdscr.clear()
    stdscr.refresh()

    # Start colors in curses
    curses.start_color()
    curses.init_pair(1, curses.COLOR_CYAN, curses.COLOR_BLACK)
[...]

PR #265, opened 2019-04-15
Added curses logger class to provide a console GUI

SR4ven:
First attempt to add a console GUI.
Use session option `console_gui` to enable. If constructed with default arguments `keep_web_open` is redundant, so it's being disabled.
There is still a bug where the screen dimensions aren't correctly read on the first startup, just resize the window to fix this. Resizing the window smaller than 130x40 (WxH) will output a black screen to avoid curses errors.

I'd like to show the paused status too, but I couldn't find a way to pass the value to the logger class without adding another function to all logger classes as I already did with the window resize function. This is something for another time though.

The SIGWINCH handler is necessary as the curses default handler interrupts sleeps, which might result in errors when the minimum sleep time between cases is suddenly shortened. With the custom signal handler the sleep time is restarted when a resize event is caught. That again might result in longer sleep times than expected, but in my opinion this is the evil. To prevent this we'd have to write a custom sleep function.

I haven't tested this under Windows at all, it would be nice if someone else could. [windows-curses](https://pypi.org/project/windows-curses/) is definitely needed.

Suggestions, especially about the design, are welcome.

### DIFF
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -1,10 +1,11 @@
 from __future__ import absolute_import, print_function
 
-import cPickle
+import pickle
 import datetime
 import logging
 import os
 import re
+import six
 import threading
 import time
 import traceback
@@ -30,6 +31,9 @@ from . import primitives
 from . import exception
 from .web.app import app
 
+from future.utils import listitems
+from io import open
+from builtins import input
 
 class Target(object):
     """Target descriptor container.
@@ -97,7 +101,7 @@ class Target(object):
                 time.sleep(1)
 
             # connection established.
-            for key, value in self.procmon_options.items():
+            for key, value in listitems(self.procmon_options):
                 getattr(self.procmon, 'set_{0}'.format(key))(value)
 
         # If the network monitor is alive, set it's options
@@ -109,7 +113,7 @@ class Target(object):
                 time.sleep(1)
 
             # connection established.
-            for key in self.netmon_options.keys():
+            for key in list(self.netmon_options):
                 eval('self.netmon.set_%s(self.netmon_options["%s"])' % (key, key))
 
     def recv(self, max_bytes=None):
@@ -518,10 +522,10 @@ class Session(pgraph.Graph):
             src = self.root
 
         # if source or destination is a name, resolve the actual node.
-        if type(src) is str:
+        if isinstance(src, six.string_types):
             src = self.find_node("name", src)
 
-        if type(dst) is str:
+        if isinstance(dst, six.string_types):
             dst = self.find_node("name", dst)
 
         # if source or destination is not in the graph, add it.
@@ -563,7 +567,7 @@ class Session(pgraph.Graph):
         }
 
         fh = open(self.session_filename, "wb+")
-        fh.write(zlib.compress(cPickle.dumps(data, protocol=2)))
+        fh.write(zlib.compress(pickle.dumps(data, protocol=2)))
         fh.close()
 
     def feature_check(self):
@@ -707,7 +711,7 @@ class Session(pgraph.Graph):
             if self._keep_web_open and self.web_port is not None:
                 print("\nFuzzing session completed. Keeping webinterface up on localhost:{}".format(self.web_port),
                       "\nPress ENTER to close webinterface")
-                raw_input()
+                input()
         except KeyboardInterrupt:
             # TODO: should wait for the end of the ongoing test case, and stop gracefully netmon and procmon
             self.export_file()
@@ -739,8 +743,8 @@ class Session(pgraph.Graph):
 
         try:
             with open(self.session_filename, "rb") as f:
-                data = cPickle.loads(zlib.decompress(f.read()))
-        except (IOError, zlib.error, cPickle.UnpicklingError):
+                data = pickle.loads(zlib.decompress(f.read()))
+        except (IOError, zlib.error, pickle.UnpicklingError):
             return
 
         # update the skip variable to pick up fuzzing from last test case.
@@ -928,8 +932,8 @@ class Session(pgraph.Graph):
             session (Session): Session object calling post_send.
                 Useful properties include last_send and last_recv.
 
-            args: Implementations should include \*args and \**kwargs for forward-compatibility.
-            kwargs: Implementations should include \*args and \**kwargs for forward-compatibility.
+            args: Implementations should include \\*args and \\**kwargs for forward-compatibility.
+            kwargs: Implementations should include \\*args and \\**kwargs for forward-compatibility.
         """
         # default to doing nothing.
         self._fuzz_data_logger.log_info("No post_send callback registered.")
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -12,7 +12,6 @@ import traceback
 import zlib
 import socket
 import errno
-import signal
 
 from tornado.httpserver import HTTPServer
 from tornado.ioloop import IOLoop
@@ -377,13 +376,12 @@ class Session(pgraph.Graph):
         if fuzz_data_logger is not None:
             raise exception.BoofuzzError('Session fuzz_data_logger is deprecated. Use fuzz_loggers instead!')
         if fuzz_loggers is None:
-            if self.console_gui and os.name != 'nt':
-                default_signal_handler = signal.getsignal(signal.SIGWINCH)
-                fuzz_loggers = [fuzz_logger_curses.FuzzLoggerCurses(web_port=self.web_port)]
-                self._keep_web_open = False
-                signal.signal(signal.SIGWINCH, default_signal_handler)
-            else:
-                fuzz_loggers = [fuzz_logger_text.FuzzLoggerText()]
+            fuzz_loggers = []
+        if self.console_gui and os.name != 'nt':
+            fuzz_loggers.append(fuzz_logger_curses.FuzzLoggerCurses(web_port=self.web_port))
+            self._keep_web_open = False
+        if len(fuzz_loggers) == 0:
+            fuzz_loggers = [fuzz_logger_text.FuzzLoggerText()]
 
         helpers.mkdir_safe(os.path.join(constants.RESULTS_DIR))
         self._run_id = datetime.datetime.utcnow().replace(microsecond=0).isoformat().replace(':', '-')
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -208,11 +208,12 @@ class SessionInfo(object):
 
     @property
     def total_num_mutations(self):
-        return 100  # TODO upgrade database format to store this info
+        return None
 
     @property
     def total_mutant_index(self):
-        return 100  # TODO SELECT COUNT(*) FROM cases -- But watch out for partially finished case
+        x = self._db_reader.query('''SELECT COUNT(*) FROM cases''').next()[0]
+        return x
 
     def test_case_data(self, index):
         """Return test case data object (for use by web server)
@@ -386,6 +387,8 @@ class Session(pgraph.Graph):
         self._db_logger = fuzz_logger_db.FuzzLoggerDb(db_filename=self._db_filename,
                                                       num_log_cases=fuzz_db_keep_only_n_pass_cases)
 
+        self._crash_filename = 'boofuzz-crash-bin-{0}'.format(self._run_id)
+
         self._fuzz_data_logger = fuzz_logger.FuzzLogger(fuzz_loggers=[self._db_logger] + fuzz_loggers)
         self._check_data_received_each_request = check_data_received_each_request
         self._receive_data_after_each_request = receive_data_after_each_request
@@ -437,6 +440,7 @@ class Session(pgraph.Graph):
         self.add_node(self.root)
 
         if target is not None:
+            target.procmon_options['crash_filename'] = self._crash_filename
             try:
                 self.add_target(target=target)
             except exception.BoofuzzRpcError as e:
@@ -1349,12 +1353,7 @@ class Session(pgraph.Graph):
                 self._fuzz_data_logger.open_test_step('Calling netmon pre_send()')
                 target.netmon.pre_send(self.total_mutant_index)
 
-            try:
-                if not self._reuse_target_connection:
-                    target.open()
-            except exception.BoofuzzTargetConnectionFailedError:
-                self._fuzz_data_logger.log_error(constants.ERR_CONN_FAILED_TERMINAL)
-                raise
+            self._open_connection_keep_trying(target)
 
             self._pre_send(target)
 
@@ -1424,12 +1423,7 @@ class Session(pgraph.Graph):
             target.netmon.pre_send(self.total_mutant_index)
 
         try:
-            try:
-                if not self._reuse_target_connection:
-                    target.open()
-            except exception.BoofuzzTargetConnectionFailedError:
-                self._fuzz_data_logger.log_error(constants.ERR_CONN_FAILED_TERMINAL)
-                raise
+            self._open_connection_keep_trying(target)
 
             self._pre_send(target)
 
@@ -1451,13 +1445,31 @@ class Session(pgraph.Graph):
 
             if self.sleep_time > 0:
                 self._fuzz_data_logger.open_test_step("Sleep between tests.")
-                self._fuzz_data_logger.log_info("sleeping for %f seconds" % self.sleep_time)
-                time.sleep(self.sleep_time)
+                self._sleep(self.sleep_time)
         finally:
             self._process_failures(target=target)
             self._stop_netmon(target=target)
             self._fuzz_data_logger.close_test_case()
             self.export_file()
+
+    def _open_connection_keep_trying(self, target):
+        """ Open connection and if it fails, keep retrying.
+
+        Args:
+            target (Target): Target to open.
+        """
+        if not self._reuse_target_connection:
+            while True:
+                try:
+                    target.open()
+                    break  # break if no exception
+                except exception.BoofuzzTargetConnectionFailedError:
+                    self._fuzz_data_logger.log_info(constants.WARN_CONN_FAILED_TERMINAL)
+                    self._restart_target(target)
+
+    def _sleep(self, seconds):
+        self._fuzz_data_logger.log_info("sleeping for %f seconds" % seconds)
+        time.sleep(seconds)
 
     def _test_case_name_feature_check(self, path):
         message_path = "->".join([self.nodes[e.dst].name for e in path])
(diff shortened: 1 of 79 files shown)


jtpereyda:
@SR4ven sweet! This is cool! My initial thought on the design: I think it'd be cool if we could use whatever the default terminal background color is. That's a bit of a personal preference issue admittedly, but I like being able to choose the dark terminal background color I like, and it makes an application feel less obtrusive. It gives it a lightweight feel IMO instead of feeling like a whole new environment. Counterpoints welcome!

Some of the bugs you mentioned could use working out, but I'd be OK merging this as experimental. This is a great start.

I should have some comments on the code itself when I get a chance.

jtpereyda:
While testing (running on Linux, fuzzing a Windows target using the procmon), it seems to be that resizing the terminal sometimes causes connection issues with the procmon. I guess this is big enough to make me apprehensive about merging.

To reproduce, I just resize the terminal window a bunch of times, and eventually it causes a problem. Probably has something to do with timing. It may be that it's harder to reproduce with a local procmon since the latency is smaller; haven't tried.

This seems like it will be a bear to debug. Given this and some of the issues you mentioned, I would recommend looking at a higher level framework like asciimatics: https://github.com/peterbrittain/asciimatics

Reasoning: Curses is pretty low level and requires a lot of legwork. A major legwork item will be using windows-curses to get cross-platform compatibility. A framework like asciimatics should be able to cover a lot of these chores for us and simplify the implementation. There's a decent chance the bugs we've discussed so far will go away with the use of a good framework.

I mentioned asciimatics in the corresponding issue: https://github.com/jtpereyda/boofuzz/issues/191

Given this high level issue, I won't be reviewing the code in detail at this point. But one further thought: The active console module is a little more like the web server than the logger. If you poke around in the web server code, you'll notice that it uses the FuzzLoggerDb instance to query results and present them on the web page. A similar tack could be taken for the interactive curses module. If it's easier to do FuzzLoggerCurses for now though, that's fine.

SR4ven:
As far as I could see, asciimatics uses the same method to handle terminal resizing, which is the SIGWINCH signal. The problem with this is, that any signal will interrupt the sleep function and won't continue sleeping for the remaining time after finishing the interrupt routine. If that's the case for asciimatics too, it won't fix our problem. Apparently, Curses comes with it's own [sleep function](https://stackoverflow.com/questions/39852148/curses-breaks-a-time-sleep-when-terminals-size-has-been-changed) to handle this issue, however I couldn't get it to work when I first came across this problem.

I'll see if I can disable the SIGWINCH handling and just periodically check for a change in the terminal size.
Alternatively using the sleep function curses brings or even writing a custom sleep function might work.
Thinking of something like this:
```python
def custom_sleep(delay):
  for ms in range(delay * 1000):
    time.sleep(0.001)
```
An interrupt wouldn't have a big impact on the actual sleep time, as it would at max skip 1ms per resize. However, this is definitely not a clean or efficient way of solving this.

SR4ven:
Resetting the SIGWINCH handler after initializing curses turned out to work quite well as fas as I can see.
Now, resizing the terminal doesn't cause an interrupt anymore and sleep functions are no longer affected. I added a `min_refresh_rate` option to the FuzzLoggerCurses class that guarantees a screen refresh rate. Additionally a refresh is triggered after a test_case has been closed.

To make this compatible to python3, I added a version check prior to reading the current terminal size. I now figured that I should only read this once during initialisation and store it in a variable. Currently the whole screen is redrawn every second no matter if the terminal size has changed or not. If this fix works as intended I will add a check so it only refreshed when necessary.

@jtpereyda As I'm not sure how you found the issue with resizing the terminal, I couldn't verify that it's fixed now. Could you please test for this issue again and let me know if it's fixed?

BTW, tox doesn't run anymore because of a platform mismatch (linux/linux2), PR incoming.

jtpereyda:
@SR4ven I just tried again on my Ubuntu box. While the Curses display worked smoothly, upon refresh I got this error:
```
> python ftp.py fuzz --target-host $TARGETIP --target-port 21112 --username james --password jamjam --procmon-host $TARGETIP --procmon-start 'C:\\ftpd\\smallftpd.exe' 

Exception in thread curses_logger:
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/jpereyda/code/boofuzz/boofuzz/fuzz_logger_curses.py", line 307, in _draw_screen
    self._draw_case()
  File "/home/jpereyda/code/boofuzz/boofuzz/fuzz_logger_curses.py", line 248, in _draw_case
    self._casescr.refresh(pos, 0, 2, 1, self._height - 18, self._width - 2)
error: prefresh() returned ERR
```
The curses display had crashed, but the web view showed that the test was still running.

Here's another one (this one fatal):
```
> python ftp.py fuzz --target-host $TARGETIP --target-port 21112 --username james --password jamjam --procmon-host $TARGETIP --procmon-start 'C:\\ftpd\\smallftpd.exe' 

  File "ftp.py", line 94, in <module>
    cli()
  File "/home/jpereyda/.local/lib/python2.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/jpereyda/.local/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/jpereyda/.local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jpereyda/.local/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/jpereyda/.local/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "ftp.py", line 57, in fuzz
    session.fuzz()
  File "/home/jpereyda/code/boofuzz/boofuzz/sessions.py", line 605, in fuzz
  File "/home/jpereyda/code/boofuzz/boofuzz/sessions.py", line 710, in _main_fuzz_loop
  File "/home/jpereyda/code/boofuzz/boofuzz/sessions.py", line 1426, in _fuzz_current_case
    self._open_connection_keep_trying(target)
  File "/home/jpereyda/code/boofuzz/boofuzz/sessions.py", line 1464, in _open_connection_keep_trying
    target.open()
  File "/home/jpereyda/code/boofuzz/boofuzz/sessions.py", line 84, in open
    self._target_connection.open()
  File "/home/jpereyda/code/boofuzz/boofuzz/socket_connection.py", line 158, in open
  File "/usr/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
socket.error: [Errno 4] Interrupted system call
```

These failures don't happen on every resize, but often enough that it's not hard to reproduce.

I'm running in gnome-terminal on Ubuntu 18.04 and fuzzing against a remote host on my LAN, with procmon running on said host.

SR4ven:
Thanks for the exceptions, they help alot!
Without having tested this myself, I'm quite sure the first non fatal issue is related to some offset/overflow error. This error occures, when the program attempts to write outside the defined window. So either I messed something up with the cursor offsets or the current test case generates more than 200 lines (not sure how \n inside the log payload is handled here, but I think it's ignored if not set elsewise in curses; needs to be tested). This can be verified, by changing `max_log_lines=200` L25 in FuzzLoggerCurses to a high number.
The program doesn't quit because the curses class runs in it's own thread and the main thread just continues.

As for the second error, this happens because both Curses and Socket run on a C layer. It seems that while setting up curses, not only a signal handler in python, but also one in C is created. In the recent commit I reset this in python but not in C.
So my guess is that if you resize the window while any C layer function is run, it will be interrupted. As the python signal handler is no longer present, this will result in an error. As far as I know the only way to disable the C layer signal handler would be to compile curses with custom arguments; that's not an option. One possibility could be to catch socket errno 4 and retry if it fails. Not the cleanest solution for sure, as SIGINT might be caught by that as well and prevent the program from exiting; I'm not exactly sure about this though. Alternatively writing a C library that resets the signal handler might be a better option if that works.

tl;dr First error seems easy to fix, second one might be a little more tricky.

If a try/except around every socket call is not acceptable, we cannot use curses anymore. The problem with that is, that even asciimatics seems to use curses for unix platforms [Link](https://github.com/peterbrittain/asciimatics/blob/f471427d7786ce2d5f1eeb2dae0e67d19e46e085/tests/test_screen.py#L29). If this is true, I'm quite sure switching over to asciimatics wouldn't solve issues related to terminal resizing.
Sadly I somehow overread that you mentioned asciimatics and I've done it all with curses now. Porting the existing code over to asciimatics might be more work right now, but will make future changes/additions easier. Windows compatibility with windows-curses still needs to be checked and if it doesn't work, curses alone is probably not the way to go.
However, it's up to you @jtpereyda to decide which one we use.

jtpereyda:
@SR4ven I like the idea of asciimatics the best, but 1) I don't have strong evidence about how well it resolves these various issues and 2) I'm not the one writing the code at this point 😛 

I'm not too familiar with the particular signal handling issues you've mentioned. Do you have any links or technical references by chance? 

My recommendation is to check out asciimatics. My hunch is that if some tricky solution is needed to make curses really work, a good framework will have already implemented that tricky solution, as that's what any good framework tends to do. But there is a risk that it won't really solve the problems. I might be able to give it a shot in the next few weeks if you don't have the time.

As far as a decision, I think it should be working pretty reliably on at least one platform before we merge into master. That doesn't strictly mean we need to use asciimatics, but it does mean the module shouldn't be crashing. asciimatics is my best hunch for an effective way to get there.

By the way, what OS and terminal emulator have you tested on?

SR4ven:
@jtpereyda I'm testing on Ubuntu 18.10 with Unity.

As for the references:
 - https://stackoverflow.com/a/3016868
 - https://stackoverflow.com/questions/39852148
- https://github.com/ThomasDickey/ncurses-snapshots/blob/8f4b15d59993e0f7755146e6efb11007b1eb6f22/INSTALL#L647
- https://github.com/ThomasDickey/ncurses-snapshots/blob/8f4b15d59993e0f7755146e6efb11007b1eb6f22/INSTALL#L688
- https://linux.die.net/man/3/napms

I might be completely wrong with this, but to my understanding no framework could fix the error we're facing here. Asciimatics uses curses for unix platforms. Curses sets up a signal handler for SIGWINCH and as far as I know the only way to disable is to compile curses with different options. So Asciimatics wouldn't be able to fully disable the signal handlers of curses.
Now, as long as the default time.sleep() function is used, asciimatics wouldn't have any influence on neither sleep time nor signals/interrupts. The only way for asciimatics or curses to take control over the actual sleep time would be to use the sleep function that comes with those libraries. Curses has [napms](https://linux.die.net/man/3/napms) for this. It keeps track of the actual sleep time and in case of an interrupt, continue sleeping for the remaining time afterwards. I didn't find anything like this in asciimatics yet, but they might handle it differently.

The general issue with a SIGWINCH handler is, that it halts the main program for the time the screen gets adjusted to the new terminal size. As we are working with low level sockets, interrupting those socket functions like send() and recv(), written in C, is not really a good idea. Also, resizing the terminal should have much less priority than operations on the socket in our case.
That's why I'd prefer to just periodically check for a change in the terminal size manually rather than setting up a signal handler for it. As said before, this seems to be tricky with curses on the C layer, but simple in Python itself.

Anyway, I'll try to get asciimatics working and see if they somehow found a way to get around these problems. For this I'll open a seperate PR so we can keep track of each ones bugs more easily, once it's ready.

jtpereyda:
Thank you for the links, that helps. `asciimatics` has a note on screen resizing, but it does not seem to directly address the interrupt issue: https://asciimatics.readthedocs.io/en/stable/io.html#screen-resizing

Since you're doing some rework, I'll share some high level thoughts on the implementation. Neither of these are strictly requirements, but I'm sharing them just in case they end up helping:

1. It may be better to interface with the existing FuzzLoggerDb module, or a FuzzLoggerDbReader. I think you can create a reader that reads the file even as it's being written. The module/class that handles terminal drawing could just read from the database... However this could easily cause its own set of headaches so 🤷‍♂ 
2. I recommend avoiding adding window_resize to IFuzzLogger as it is implementation specific to this particular logger. It may be cleaner just to call the single FuzzLoggerCurses object (assuming you still have FuzzLoggerCurses or something like it).

SR4ven:
After having played around with asciimatics a little, I can say that the signal handling is the same as with curses. Resizing the terminal while e.g. socket.recv() is waiting for input, crashes the program with socket errno 4 (just as the fatal error you posted above). Restoring the default signal handler fixes this, but the resizing has to be done manually (timer based). This is the same with curses.
If you have the time, please test for the two errors you posted before again, with a higher `max_log_lines` setting. As I mentioned before I couldn't reproduce them anymore and it would be important to know whether the signal handling is still an issue or not. If it still is, we first need to find a fix/workaround before working on things like asciimatics or bugs/looks.

About your suggestions:
 1. Good point, I thought about this too, but for now I'd just like to get it working as it is. Changing the way the log data is handed to the console logger class shouldn't be an issue at a later point.
But keep in mind that when not all cases are saved in the db, you wouldn't see them on the console screen (`fuzz_db_keep_only_n_pass_cases`). This might be the expected behaviour, just saying. On top of that, the db only holds finished test cases (it only updates when `close_test_case()` is called). So if a case is stuck in the middle of a case, the console and web gui would only display the last test case.
For later, when you want to scroll thru previous test cases like you can on the webinterface, a db-reader is definitely needed.
2. `window_resize()` is no longer needed when we don't use a signal handler anymore. If you can confirm that the current PR version works fine, we can remove it.

jtpereyda:
> That's why I'd prefer to just periodically check for a change in the terminal size manually rather than setting up a signal handler for it. As said before, this seems to be tricky with curses on the C layer, but simple in Python itself.

@SR4ven That seems to make sense. Any major downsides?

SR4ven:
Not that I know. Depending on the checking interval it might feel a little less responsive, but I don't think that's going to be a problem.
Reading the terminal size also takes up more resources than simply waiting for a signal, but this might be only a very small impact with checking intervals of around 1 second.

SR4ven:
So I tried my luck with asciimatics, sadly without any success. I couldn't even get it to display more than one window. Either I completely missunderstood something or it isn't as easy to setup as it appears to be.

Therefore I continued working with curses and applied some changes according to your requests (see commit messages).
As for the Windows compatibility, `windows-curses` doesn't work "out of the box" with the current code. Some functions seem to work sightly different, but there is a good [documentation](https://github.com/wmcbrine/PDCurses/blob/master/man/README.md), so it might be possible to get this working later.

On Linux, the GUI works quite nicely in my eyes but it needs some more testing. @jtpereyda it would be nice if you could check for bugs and share your thoughts.

boofuzz/sessions.py
```diff
@@ -365,13 +370,20 @@ def __init__(self, session_filename=None, index_start=1, index_end=None, sleep_t
...
         self.restart_sleep_time = restart_sleep_time
         if fuzz_data_logger is not None:
             raise exception.BoofuzzError('Session fuzz_data_logger is deprecated. Use fuzz_loggers instead!')
         if fuzz_loggers is None:
-            fuzz_loggers = [fuzz_logger_text.FuzzLoggerText()]
+            if self.console_gui and os.name != 'nt':
+                default_signal_handler = signal.getsignal(signal.SIGWINCH)
+                fuzz_loggers = [fuzz_logger_curses.FuzzLoggerCurses(web_port=self.web_port)]
+                self._keep_web_open = False
+                signal.signal(signal.SIGWINCH, default_signal_handler)
```
jtpereyda: I realize why it was bugging out for me -- I was manually creating a `FuzzLoggerCurses()` and passing it into the Session class without configuring all of these.

What if we move the two signal handler lines into the FuzzLoggerCurses constructor? A quick test worked for me with these lines surrounding the thread start.

boofuzz/sessions.py
```diff
@@ -365,13 +370,20 @@ def __init__(self, session_filename=None, index_start=1, index_end=None, sleep_t
...
         self._keep_web_open = keep_web_open
+        self.console_gui = console_gui
         self._crash_threshold_node = crash_threshold_request
         self._crash_threshold_element = crash_threshold_element
         self.restart_sleep_time = restart_sleep_time
         if fuzz_data_logger is not None:
             raise exception.BoofuzzError('Session fuzz_data_logger is deprecated. Use fuzz_loggers instead!')
         if fuzz_loggers is None:
-            fuzz_loggers = [fuzz_logger_text.FuzzLoggerText()]
+            if self.console_gui and os.name != 'nt':
```
jtpereyda: Let's move the if statement up and out of `if fuzz_loggers is None` -- they seem like two separate possibilities to me. If we put the `console_gui` check first, then if it triggers, it will cause `if fuzz_loggers is None` to always be false.

I think the console gui should eventually be default.

boofuzz/sessions.py
SR4ven: > What if we move the two signal handler lines into the FuzzLoggerCurses constructor? A quick test worked for me with these lines surrounding the thread start.

That totally makes sense, I will do some testing and move it over.

> 
> 
> A couple review comments below. This latest commit didn't run for me without a modification -- I suspect it has to do with some of the Py2/Py3 changes. Here's the patch:
> 
> ```
> > git diff
> diff --git a/boofuzz/pedrpc.py b/boofuzz/pedrpc.py
> index d4d9fa1..92c9ca3 100644
> --- a/boofuzz/pedrpc.py
> +++ b/boofuzz/pedrpc.py
> @@ -9,7 +9,7 @@ import select
>  from . import exception
>  
>  
> -class Client:
> +class Client(object):
> ```

Is this patch still required when using the session option to construct the curses-class?

boofuzz/sessions.py
SR4ven: Good point, but we have to consider the case where a logger class is passed into `fuzz_loggers`.
This requires appending the FuzzLoggerCurses class to the passed loggers.
Should be doable, I'll test this.

boofuzz/sessions.py
jtpereyda: I didn't try specifically with the Session option but without this patch -- I can make it a separate PR too.

jtpereyda: @SR4ven by the way can you explain a bit about what those signal handling lines do? I checked the the [docs for signal](https://docs.python.org/2/library/signal.html#signal.getsignal) but I don't understand why they make things work here.

SR4ven: @jtpereyda So, `default_signal_handler = signal.getsignal(signal.SIGWINCH)` saves the current SIGWINCH handler, `signal.signal(signal.SIGWINCH, default_signal_handler)` restores it. [Reference](https://stackoverflow.com/questions/22916783/reset-python-sigint-to-default-signal-handler)
The position of the default_signal_handler saving  after ac4afff might be a little too far down in the code, moving it above `self._stdscr = curses.initscr()` would be safer, but it passed my tests as it is. 

This is necessary because curses sets up a custom handler. That handler allows interrupting the program at any point, messing with the timings and causing errors.
The default signal handler does literally nothing, so we just use that and check for a terminal size change periodically. Not the most responsive way, but it works.

I hope that makes it more clear.

jtpereyda:
Here goes! Thank you again @SR4ven .

PR #266, opened 2019-04-22
Added python3 compatibility

aedrax:
This pull request is for adding compatibility for python 3 without losing compatibility with python 2. In many cases I am utilizing the "[six](https://pythonhosted.org/six/)" library. 

For future work, but as a separate issue, I think we should work on making all data stored as a "six.binary_type" to guarantee compatibility between python versions and to avoid needing to worry about the encoding of the string object.

Some additional issue I believe we should open would be to update the contribution guide, and to add more testing to the framework


SR4ven:
Nice job, I was planning on doing this for a while now but never had the time.
However, there appear to be some problems with the CI and the compatibility.

Currently, the Linux tox platform is set to `linux2`, which means a Linux platform with Python2 is required. This is no longer needed so we should change it to just `linux`.
Second, we now also want to have tox testing on python3 so we need to add a python3 environment to the envlist like `envlist = py27-{unix,windows}, py36-{unix,windows}`.
Finally we want to make sure tox uses the python version of the current environment that is being tested on. To do so all `python` commands need to replaced with `{envpython}`.
Summing this up, the tox.ini file should look like this now
```ini
[tox]
envlist =
    py27-{unix,windows}
    py36-{unix,windows}

[testenv]
whitelist_externals =
    sudo
    make
platform =
    windows: win32
    unix: linux|darwin
extras = dev
install_command =
    {envpython} -m pip install {opts} {packages}
commands_pre =
    {envpython} -m pip check
    # Stop the test if there are Python syntax errors or undefined names.
    flake8 . --count --select=E901,E999,F821,F822,F823 --per-file-ignores='utils/ida_fuzz_library_extender.py:F821' --show-source --statistics
    # Code-Style check. exit-zero treats all errors as warnings. Ignoring F405 because we use many star imports.
    # Use '# noqa: F401' for in-line ignores.
    flake8 . --count --exit-zero --ignore F405 --max-complexity=10 --max-line-length=120 --statistics
commands =
    windows: {envpython} -m pytest
    unix: sudo {envpython} -m pytest
    {envpython} -m check_manifest
    windows: .\docs\make.bat dummy
    unix: make -C ./docs dummy
commands_post =
    windows: rmdir /S /Q .\boofuzz-results .\docs\_build
    unix: sudo rm -rf ./boofuzz-results/ ./docs/_build/
```
That done, you can now run `tox` and check for any errors (there are still quite a few present).

Also, you might want to add a python3 classifier to the setup.py like `'Programming Language :: Python :: 3.6',` so it appears on pypi.

aedrax:
ah, yes. Thank you for the tox file update, I see the additional problems now. I will get those fixed up this evening.

jtpereyda:
@aedrax this is great! Thank you for the PR.

The major blocker is to get the unit tests passing again. I looked at one of the first failures here: https://travis-ci.org/jtpereyda/boofuzz/builds/524192826#L989

```
        return (src_addr +
                dst_addr +
                six.binary_type(b"\x00") +
>               chr(ip_constants.IPV4_PROTOCOL_UDP) +
                struct.pack(">H", msg_len))
E       UnicodeDecodeError: 'ascii' codec can't decode byte 0x89 in position 4: ordinal not in range(128)
```

So `chr` returns a unicode string in Python 3 but a byte string in Python 2. My favorite Python 2/3 cheat sheet describes a solution: https://python-future.org/compatible_idioms.html#chr

I recommend using the `from builtins import bytes` approach. The line above would become:

```
>               bytes([ip_constants.IPV4_PROTOCOL_UDP]) +
```

This strategy should resolve at least some of the errors. The cheat sheet I linked above contains a pretty thorough index of tricks like this. But please post here if you'd like help figuring out some of the test failures.

Good luck!

aedrax:
Thanks for the suggestion! I've made the changes but am currently going down a rabbit hole of new bugs popping up with the change, so I changed the PR to WIP for now until I can get it fixed up. If you would prefer, I can close the PR and reopen a new one when the bugs are gone.

jtpereyda:
@aedrax Gotcha, leaving it as WIP is great. This is going to be a pretty tricky change. Some of the bytestring vs. unicode issues will likely be the toughest.

If you can get unit tests passing and at least be pretty sure that the major functionality works, we can merge and work out the rest as bugs later. Getting it merged ASAP is important as it will get more complicated as more and more PRs get merged into master.

An alternative tack is to make a few compatibility changes at a time and merge them in separate PRs. This will avoid the merging problem and allow progress to move forward a little at a time.

boofuzz/utils/crash_binning.py
```diff
@@ -197,7 +197,7 @@ def export_file(self, file_name):
         self.last_crash = self.pydbg = None
 
         fh = open(file_name, "wb+")
```
cq674350529: maybe try another elegant way:

    json.dump(self.bins, open(file_name, "wb+"), default=lambda o: o.__dict__))

boofuzz/utils/crash_binning.py
```diff
@@ -218,11 +218,15 @@ def import_file(self, file_name):
         @return:            self
         """
 
-        fh  = open(file_name, "rb")
-        tmp = pickle.loads(zlib.decompress(fh.read()))
-        fh.close()
-
-        self.bins = tmp.bins
+        self.bins = {}
+        with open(file_name, "rb") as fh:
```
cq674350529: another elegant way:

    bin_dict = json.load(open(file_name, "rb"))

boofuzz/utils/crash_binning.py
```diff
@@ -197,7 +197,7 @@ def export_file(self, file_name):
         self.last_crash = self.pydbg = None
 
         fh = open(file_name, "wb+")
-        fh.write(zlib.compress(pickle.dumps(self, protocol=2)))
+        fh.write(json.dumps(self.bins, default=lambda o: o.__dict__))
```
cq674350529: Maybe try another elegant way:

    json.dump(self.bins, open(file_name, "wb+"), default=lambda o: o.__dict__)

boofuzz/utils/crash_binning.py
```diff
@@ -218,11 +218,15 @@ def import_file(self, file_name):
...
         """
 
-        fh  = open(file_name, "rb")
-        tmp = pickle.loads(zlib.decompress(fh.read()))
-        fh.close()
-
-        self.bins = tmp.bins
+        self.bins = {}
+        with open(file_name, "rb") as fh:
+            bin_dict = json.loads(fh.read())
```
cq674350529: another elegant way:

    bin_dict = json.load(open(file_name, "rb"))

boofuzz/utils/crash_binning.py
jtpereyda: @cq674350529 That's great. This change is just a merge-in from master, so I'll leave it up to @aedrax if he wants to address it here.

boofuzz/utils/crash_binning.py
jtpereyda: Neat! As with the last one I'll leave it up to @aedrax whether to make this change since he's focusing on Py2/Py3. I was the one who wrote this code so my bad.

boofuzz/utils/crash_binning.py
aedrax: I probably won't touch that since it is out of scope of the issue

cq674350529: If needed, I can issue another pull request.

unit_tests/test_fuzz_logger_text.py
```diff
@@ -456,7 +456,7 @@ def test_several(self):
         self.virtual_file.seek(0)
         self.assertTrue(self.some_test_case_id in self.virtual_file.readline())
         self.assertTrue(self.some_test_step_msg in self.virtual_file.readline())
-        self.assertTrue(self.some_recv_data in self.virtual_file.readline())
+        self.assertTrue(str(len(self.some_recv_data)) in self.virtual_file.readline())
```
cq674350529: @jtpereyda Why do the assertions for `log_recv()` and `log_send()` differ each other? 

Besides,  the assertions for log_recv() elsewhere seem inconsistent. 

1) using `fuzz_logger_text.DEFAULT_HEX_TO_STR(self.some_recv_data)`
https://github.com/jtpereyda/boofuzz/blob/0a78c15a781b9feea5321a0f350e3ad28ee0da68/unit_tests/test_fuzz_logger_text.py#L197-L212

2) using `self.some_recv_data` directly
https://github.com/jtpereyda/boofuzz/blob/0a78c15a781b9feea5321a0f350e3ad28ee0da68/unit_tests/test_fuzz_logger_text.py#L453-L455

aedrax: that's a great question, I will look into why that's like that

cq674350529: According to the format of log, it seems both `self.assertTrue(self.some_recv_data in self.virtual_file.readline())` and `self.assertTrue(str(len(self.some_recv_data)) in self.virtual_file.readline())` would work.

SR4ven: This would only work for Python2 though. Python3 expects a string type for the left operand of "a in b"
```python
TypeError: 'in <string>' requires string as left operand, not bytes
unit_tests/test_fuzz_logger_text.py:459: TypeError
```
In the current state the test passes, but I don't think it works as intended. Right now it only check if the `len(self.some_recv_data)` exists, when it actually should check if `self.some_recv_data` itself does.
So this is what I belive the line should look like (correct me if I'm wrong):
```python
self.assertTrue(str(self.some_recv_data) in self.virtual_file.readline())
```
@cq674350529 Good catch! I wouldn't have noticed as the test passes either way.
@aedrax I'll include this in the next PR on your branch or you can just change it yourself.

jtpereyda: Be aware that send/recv data can be truncated in some loggers... I don't _think_ this occurs in FuzzLoggerText though.

I may have loosened up a check after changing the format of the output. I'm fine with somebody making the check more strict as long as it makes sense and tests still pass.

SR4ven:
@jtpereyda  @aedrax would it make sense to test against multiple minor versions of python like some other repos are doing? [pip](https://travis-ci.org/pypa/pip/builds/532263898?utm_source=github_status&utm_medium=notification) [impacket](https://travis-ci.com/SecureAuthCorp/impacket/builds/111381392)

jtpereyda:
@SR4ven makes enough sense to me. For this PR, at least one version of Python 3 is enough, but I don't have a problem with more.

aedrax:
@jtpereyda this should be passing, but the version of python on the testing server is 2.7.6, is this the version of python we want to test against? I didn't realize there were features that weren't available in 2.7.6

jtpereyda:
@aedrax Testing with older versions allows us to maintain compatibility and flexibility with different Python installs. Python 2.7.6 is fine unless it's limiting us. It doesn't hurt much to leave UTs on an older version.

We should add at least one Python 3 version. If we want to add a more up to date Python 2.7.x version that's fine.

Travis should make it super easy to add a couple different Python versions -- hopefully a 1 or 2 line change.

Looks like there are still some UT failures in Travis btw.

jtpereyda:
@aedrax Oh OK they're not passing because of the Python version. Is it that the code doesn't run in Python 2.7.6? Or is it just the unit tests that don't run correctly?

jtpereyda:
@aedrax I think I misunderstood your comment. Are you saying the unit tests would pass if the version of Python 2.7 was newer?

SR4ven:
@jtpereyda that's the case, yes.
For some reason travis/tox aren't using the latest py2 version. The tests are just fine on my local machine.
I reverted the tornado push to v5 too, as it requires python 2.7.9+, which was not given on travis. I'll see how other repos are dealing with this.

That still doesn't explain why it worked the whole time and now fails. Maybe it's the "unused" import of object from builtins that I removed in the second last commit. Might have not been as unused as it appeared to me and pycharm.

aedrax:
@jtpereyda that change to the travis file @SR4ven made allows all checks to pass, this is still because it's now using a more up-to-date version of python 2.7 however. I personally believe that we should be fine with using the newer version of 2.7 because it should be backwards compatible with any python 2.7 code. If this becomes an issue where it's breaking for specific cases, then I think that should be a new issue that can be updated in the future.

SR4ven:
I agree with @aedrax, but boofuzz will now require a minimum python version of 2.7.9 (not sure about the exact version). So Ubuntu 14.04 is no longer supported with it's default python version of 2.7.6.
However, this is a quite old version and I don't think that many people still use it.
Additionally, dropping support for those old versions would allow us to push tornado to version 5 for support of future python3 versions. Nothing really urgent right now though.

Some references from my PR:
>
>
>So apparently travis runs on Ubuntu 14.04 by default. https://docs.travis-ci.com/user/reference/overview/
>The latest py2 version for this distro is 2.7.6. https://askubuntu.com/questions/759687/how-can-i-install-a-newer-python-on-trusty
>

There are some errors when testing on Windows with python3, but that's something that can be addressed later.

jtpereyda:
Looking good. Although before we can say it's done we should probably make sure the UTs pass in Python 3. Is there any reason not to test both Python 2 and Python 3 at this point? See here for how: https://docs.travis-ci.com/user/languages/python/#specifying-python-versions

For example, we might add the following two lines:

```
  - "2.7"
  - "3.6"
```

If the tests don't pass with Python 3 yet, we can still merge without adding it to the Travis file -- we'd just have to be clear that Python 3 is not quite fully supported yet.

SR4ven:
@jtpereyda I think you might be misunderstanding something here.
We are testing on python 3 and all unit tests are passing on both python 2 and 3. They are all running in one travis test in multiple virtual environments by tox.

I didn't want to mess with the travis file in this PR, so I was going to open a separate one after this is merged.

jtpereyda:
@SR4ven Ah, totally forgot about tox -- thank you!

Congratulations all!  @SR4ven @aedrax @cq674350529

PR #267, opened 2019-04-28
Fix tox platform mismatch with travis

SR4ven:
It seems that the travis platform for linux has changed from linux2 to linux.
https://travis-ci.org/jtpereyda/boofuzz/builds/525628755


jtpereyda:
@SR4ven Thank you, great catch!

PR #269, opened 2019-05-02
Procmon fixes

jtpereyda:
- Process monitor now stores crash bins in JSON format instead of pickled format.
- Process monitor in Windows will use `taskkill -F` if `taskkill` fails.

### DIFF
--- a/boofuzz/utils/debugger_thread_pydbg.py
+++ b/boofuzz/utils/debugger_thread_pydbg.py
@@ -162,6 +162,9 @@ class DebuggerThreadPydbg(threading.Thread):
     def stop_target(self):
         try:
             os.system("taskkill /pid %d" % self.pid)
+            exit_code = os.system("taskkill /pid %d" % self.pid)
+            if exit_code != 0:
+                os.system("taskkill -F /pid %d" % self.pid)
         except OSError as e:
             print(e.errno)  # TODO interpret some basic errors
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,8 @@ Upcoming
 Features
 --------
 - New Session option `keep_web_open` to allow analyzing the test results after test completion.
+- Process monitor now stores crash bins in JSON format instead of pickled format.
+- Process monitor in Windows will use `taskkill -F`
 
 v0.1.4
 ======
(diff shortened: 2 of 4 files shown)


boofuzz/utils/debugger_thread_pydbg.py
```diff
@@ -162,6 +162,9 @@ def _scan_proc_names_once(self):
     def stop_target(self):
         try:
             os.system("taskkill /pid %d" % self.pid)
```
cq674350529: Why need to use `taskkill` twice?

jtpereyda: the key is the `-f` in the second `taskkill` call -- the idea is to forcefully kill the process if necessary. https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/taskkill

cq674350529: Yes, I know the `-F` option. Actually, I mean the two `taskkill` before `taskkill -F`. By the way, maybe using `taskkill /F` instead of `taskkill -F` to keep consistent?
https://github.com/jtpereyda/boofuzz/blob/45eef0c3d3585fd027e4a693d18e53d1a1916006/boofuzz/utils/debugger_thread_pydbg.py#L164-L167

jtpereyda: Ah, so the thinking is the initial taskkill will send the more typical SIGKILL and give the application a chance to more gracefully shut down, kind of like trying to close an application before breaking out Task Manager.

jtpereyda: Yeah, `/F` would be more consistent with what's already there.

cq674350529: Yes.  What I mean is to do it like this, without the first statement `os.system("taskkill /pid %d" % self.pid)`. Otherwise, the three `os.system()` will be executed in most cases.

    exit_code = os.system("taskkill /pid %d" % self.pid) 
    if exit_code != 0: 
        os.system("taskkill /F /pid %d" % self.pid)

jtpereyda: 🤦‍♂ oh no, I was totally not getting it. XD Yep that is right I will need to open another PR

cq674350529: ok.

jtpereyda: Feel free to push the fix if you're in a hurry, but it's my mess so I'll try to clean it up. Thanks for sticking with me until I got it. :)

cq674350529: I just want to make sure I'm not understood. Also, I just do some common code review. So you can do it when you're free.

PR #271, opened 2019-05-09
Add OpenBSD support to get_max_udp_size()

jasperla:
This patch allows boofuzz to be used on OpenBSD.


jtpereyda:
Cool! Thank you @jasperla

PR #272, opened 2019-05-10
use json.dump()/load() instead

cq674350529:
to make it more elegant.


jtpereyda:
Looks good, thank you @cq674350529 !

PR #274, opened 2019-05-14
Fixed Windows unit-tests

SR4ven:
Running the unit-tests on Windows fails at `TestSocketConnection.test_tcp_client`.
This is because Windows doesn't allow non-blocking sockets, hence raising [Errno 10035](https://docs.microsoft.com/en-us/windows/desktop/winsock/windows-sockets-error-codes-2)
To fix this, a [timeout](https://docs.python.org/3/library/socket.html#socket.socket.settimeout) for the `client_socket` has to be set too and not only for the server_socket.
There were some deprecation warnings in the socket_connection.py, that were caused by math.floor not returning an int type in python2.

Additionally, I made some changes to the tox commands, they now work as intended.


jtpereyda:
Thank you @SR4ven !

PR #275, opened 2019-05-19
Run tests against multiple python versions

SR4ven:
Travis will now run multiple jobs in parallel to test against multiple python versions.
Current versions are 2.7, 3.6, 3.7, 3.8 (optimal).
EDIT: testing against py3.8-dev is probably unnecessary and delays the test result.
The `skip_missing_interpreters = True` option in the tox.ini allows contributors to run `tox` on their machine, as only existing python interpreters are used, while travis enforces tox to run on all.

There is a conditional deploy stage again, that is supposed to trigger when a commit with tag is detected, which doesn't come from a fork (no PR).
We've had problems with that before #254 so I'm not 100% sure that this will work as intended. We can always remove the repo and fork condition though, which should then definitely work. But until you trigger the next release, we won't know for sure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,6 @@ jobs:
       env: TOXENV=py36-unix
     - python: 3.7
       env: TOXENV=py37-unix
-    - python: 3.8-dev
-      env: TOXENV=py38-unix
 
     - stage: Deploy
       install: true
@@ -33,10 +31,6 @@ jobs:
           tags: true
           repo: jtpereyda/boofuzz
           branch: master
-
-  fast_finish: true
-  allow_failures:
-    - python: 3.8-dev
 
 before_install: sudo sed -i -e 's/^Defaults\tsecure_path.*$//' /etc/sudoers
 install: pip install tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ jobs:
           repo: jtpereyda/boofuzz
           branch: master
 
+  fast_finish: true
   allow_failures:
     - python: 3.8-dev
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 dist: xenial
 sudo: required
 language: python
-cache: pip
 
 stages:
   - Test
@@ -17,7 +16,7 @@ jobs:
       env: TOXENV=py36-unix
     - python: 3.7
       env: TOXENV=py37-unix
-    - python: 3.8
+    - python: 3.8-dev
       env: TOXENV=py38-unix
 
     - stage: Deploy
@@ -36,7 +35,7 @@ jobs:
           branch: master
 
   allow_failures:
-    - python: 3.8
+    - python: 3.8-dev
 
 before_install: sudo sed -i -e 's/^Defaults\tsecure_path.*$//' /etc/sudoers
 install: pip install tox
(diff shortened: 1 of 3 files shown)


SR4ven:
I recently saw a travis integration into the checks tab of a PR (see [impacket](https://github.com/SecureAuthCorp/impacket/pull/621/checks))
This [blog entry](https://blog.travis-ci.com/2018-05-07-announcing-support-for-github-checks-api-on-travis-ci-com) seems to describe how to switch to the Travis GitHub Apps integration which enables this feature. [Docs reference](https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/)

@jtpereyda would be nice if you could get this working, however completely optimal until Travis doesn't enforce the new GitHub Apps integration (if ever).

jtpereyda:
@SR4ven Should we also do Py 3.8 in the Travis config? It's listed in tox.ini.

Is the added deploy check `if: (repo = jtpereyda/boofuzz) AND (fork = false) AND (tag IS present)` necessary? That is, what problem does it solve?

jtpereyda:
Looks like they don't support migrating travis-ci.org projects to travis-ci.com yet. Hopefully soon.

SR4ven:
> 
> 
> @SR4ven Should we also do Py 3.8 in the Travis config? It's listed in tox.ini.

Right now there is only a dev version of python3.8, that isn't worth testing against in my opinion.
That's why I removed it from travis, but forgot to remove it from tox too.
We can always add it back once a stable py38 has been released.

> 
> Is the added deploy check `if: (repo = jtpereyda/boofuzz) AND (fork = false) AND (tag IS present)` necessary? That is, what problem does it solve?

Right now travis boots a single vm, runs the tests and if the conditions are met, deploys to pip.
Now that we test against multiple python versions, we have a vm for each of them. Keeping the deployment as it was, it would try to deploy to pip on each vm. Not really nice.
So I defined a seperate `Deploy` stage, which performs no test, but only handles the deployment. Triggering that stage/vm every time wouldn't be reasonable or efficient, as the vm would boot only to notice that no deployment is needed. That's where `if: (repo = jtpereyda/boofuzz) AND (fork = false) AND (tag IS present)` comes in handy. If this condition isn't met, the whole deploy stage is skipped, speeding up the whole process.
I'm about 90% sure that the condition will be doing what it's supposed to be doing, but the only way to find out it to trigger a deploy by pushing a tagged commit. Should it fail, we can always change the condition to `if: tag IS present`. That would allow tagged commits coming from PRs to trigger the deploy branch, but it still wouldn't actually deploy as Travis generally prevents that on PRs.

SR4ven:
Update @jtpereyda
> Looks like they don't support migrating travis-ci.org projects to travis-ci.com yet. Hopefully soon.

It should now be possible to migrate to travis-ci.com now, but it is still a beta feature.
https://docs.travis-ci.com/user/migrate/open-source-repository-migration

If you don't trust the beta status, there is always the option to simply deactivate the old service and setup a new travis integration with GitHub Apps [Link](https://github.com/marketplace/travis-ci/).
The following settings would be lost in that case:
- Environment variables (public and encrypted)
- Repository SSH keys (default or custom)
- Repository settings
- Cron jobs configured for the repository
- Last build status - the build badge in travis-ci.com will reflect this status until new builds are triggered
- Next build number

Waiting is of course the safest option, but it's been more than a year since they support the checks-API and public repo migration seems to be on the very bottom of their list.

PR #278, opened 2019-05-24
shlex.split(command) function bug fix

hczdmr:
Hi,

shlex.split(command) function removed backslashes in my start command and process_monitory.py script throw exception. I fixed this bug with this commit.

Operating system: Windows7
Python: 2.7.10

Have a good day.


jtpereyda:
@hczdmr My apologies for the delay. I had to sit down and look closer at this one. After reading the [shlex.split docs](https://docs.python.org/2/library/shlex.html#shlex.split), it looks like we should be using the posix argument, e.g.:

```
        return shlex.split(command, posix=(os.name == 'posix'))
```

This should resolve the issue you observed on Windows and keep things working on Linux still.

jtpereyda:
Addressed in PR #311

PR #279, opened 2019-05-26
v0.1.5

jtpereyda:
Release PR.

FYI @SR4ven I put an import guard around `curses` because it was crashing boofuzz on Windows just to `import boofuzz *`.


SR4ven:
@jtpereyda Nice catch! I completely forgot about that because I had windows-curses installed.

PR #280, opened 2019-05-30
Provide readme for pypi and changelog for docs

SR4ven:
- Changed the setup.py so that PyPi should now show the readme ([Docs](https://packaging.python.org/guides/making-a-pypi-friendly-readme/))
Appending the changelog below the readme would work too - up to you.
- Added the changelog to the documentation, which required changing the headline levels
- Added FuzzLoggerCurses to the documentation

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,10 +1,13 @@
+Changelog
+=========
+
 Upcoming
-========
+--------
 
 v0.1.5
-======
+------
 Features
---------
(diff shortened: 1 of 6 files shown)


jtpereyda:
@SR4ven Thank you! I was wondering about how to get the README to show up in PyPI again. Adding the CHANGELOG bellow the README makes sense to me.

SR4ven:
@jtpereyda Alright, I'll add that.

Issue #281, opened 2019-05-30
Fuzzing HTTP server on Windows - TimeoutError: [WinError 10060]


starek4:
Hi,

I am trying to fuzz HTTP server. I use SocketConnection with `proto` set to "tcp". I need to get a response from tested target so I am using callback function (passed via `post_test_case_callbacks` argument to Session). In the callback function, I just call `target.rcv()` for getting the response.

There is a problem when the server reset a connection. In the Wireshark it looks like this:
![reset](https://user-images.githubusercontent.com/2153835/58629881-f87f8200-82dd-11e9-91b1-eefc1bf35730.png)

`recv()` function throws this:
```
Error!!!! A custom post_send callback function raised an uncought error.
Traceback (most recent call last):
File "[...]\AppData\Local\Programs\Python\Python37-32\lib\site-packages\boofuzz\socket_connection.py", line 186, in recv
    data = self._sock.recv(max_bytes)
TimeoutError: [WinError 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File "[...]\AppData\Local\Programs\Python\Python37-32\lib\site-packages\boofuzz\sessions.py", line 1484, in _post_send
    f(target=target, fuzz_data_logger=self._fuzz_data_logger, session=self, sock=target)
File "[...]\Desktop\repos\ysoft_fuzzer\fuzzer\src\post_test_case_callback.py", line 15, in post_test_callback
    response_string = target.recv()
File "[...]\AppData\Local\Programs\Python\Python37-32\lib\site-packages\boofuzz\sessions.py", line 134, in recv
    data = self._target_connection.recv(max_bytes=max_bytes)
File "[...]\AppData\Local\Programs\Python\Python37-32\lib\site-packages\boofuzz\socket_connection.py", line 208, in recv
    raise_(exception.BoofuzzTargetConnectionReset, None, sys.exc_info()[2])
File "[...]\AppData\Local\Programs\Python\Python37-32\lib\site-packages\future\utils\__init__.py", line 415, in raise_
    raise exc.with_traceback(tb)
TypeError: descriptor 'with_traceback' requires a 'BaseException' object but received a 'traceback'
```

[...]

PR #282, opened 2019-05-30
Fixed socket exception handling

SR4ven:
Addressing #281

- Python 3 doesn't have `e.message` so we use `str(e)` which does just the same.

- When using `raise_` from future.utils and not passing any arguments, we still have to call it with parentheses.


starek4:
Seems to be ok.

I also tested it on 3.7.1 and 2.7.16, working as expected.

jtpereyda:
Thanks for the fix @SR4ven, and for the review @starek4

PR #288, opened 2019-06-06
Check also timeout when opening TCP connection

starek4:
When Boofuzz is opening TCP connection and the target is (temporarily) unreachable, Boofuzz crash with the following error:

image

I suggest that `SocketConnection.open()` should throw `BoofuzzTargetConnectionFailedError` in this case.


jtpereyda:
@starek4 Thank you for the pull request! Looks great. Would you mind adding a line to the list of upcoming fixes in `CHANGELOG.rst`?

PR #291, opened 2019-06-19
Do not ignore empty default value in s_group

starek4:
Hi,

I think that an empty string as the default value for `s_group` should be valid.

Or is there any reason, why to replace it for the first value from possible values?

### DIFF
--- a/boofuzz/primitives/group.py
+++ b/boofuzz/primitives/group.py
@@ -24,6 +24,8 @@ class Group(BasePrimitive):
 
         assert len(self.values) > 0, "You can't have an empty value list for your group!"
 
+        if default_value is None:
+            default_value = self.values[0]
         self._value = self._original_value = default_value
 
         for val in self.values:
--- a/boofuzz/primitives/group.py
+++ b/boofuzz/primitives/group.py
@@ -24,8 +24,6 @@ class Group(BasePrimitive):
 
         assert len(self.values) > 0, "You can't have an empty value list for your group!"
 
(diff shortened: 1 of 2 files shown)


boofuzz/primitives/group.py
```diff
@@ -24,8 +24,6 @@ def __init__(self, name, values, default_value=None):
 
         assert len(self.values) > 0, "You can't have an empty value list for your group!"
 
-        if not default_value:
```
cq674350529: @starek4 Thanks. It is supposed to just filter out the `None` value of  `default_value`, nor the `''`.

Instead of just removing it, maybe filter out the `None` value in this way.

    if default_value is None:
        default_value  = self.values[0]

boofuzz/primitives/group.py
```diff
@@ -24,8 +24,6 @@ def __init__(self, name, values, default_value=None):
...
-            default_value = self.values[0]
         self._value = self._original_value = default_value
```
cq674350529: @starek4 Thanks. It was supposed to only filter out the `None` value of default_value, not the `''`.

Instead of just removing it, maybe filter out the `None` value in this way.

    if default_value is None:
        default_value = self.values[0]

starek4:
Yeah, it makes sense that if default value is `None`, that first value should be used. Good catch, thanks.

jtpereyda:
Thanks for the PR @starek4 ! Makes sense. Thank you for reviewing @cq674350529 .

Issue #290, opened 2019-06-18
Using s_random: TypeError: can only concatenate str (not "bytes") to str


kleest:
Version: 0.1.5, master (0e48caa4e022e6e35908a19bdb988fec31b98e64)
Python: 3.7

```python
session = Session(target=Target(connection=SocketConnection("127.0.0.1", 8021, proto='tcp')))

s_initialize("t")
s_random(0, 1, 1, name="random")
session.connect(s_get("t"))

session.fuzz()
```

fails with the following backtrace:
```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    session.fuzz()
  File "/boofuzz/boofuzz/sessions.py", line 598, in fuzz
    self._main_fuzz_loop(self._iterate_protocol())
  File "/boofuzz/boofuzz/sessions.py", line 690, in _main_fuzz_loop
    for fuzz_args in fuzz_case_iterator:
  File "/boofuzz/boofuzz/sessions.py", line 1227, in _iterate_protocol
    for x in self._iterate_protocol_recursive(this_node=self.root, path=[]):
  File "/boofuzz/boofuzz/sessions.py", line 1250, in _iterate_protocol_recursive
    for x in self._iterate_single_node(path):
  File "/boofuzz/boofuzz/sessions.py", line 1273, in _iterate_single_node
    while self.fuzz_node.mutate():
  File "/boofuzz/boofuzz/blocks/request.py", line 78, in mutate
    if item.fuzzable and item.mutate():
  File "/boofuzz/boofuzz/primitives/random_data.py", line 77, in mutate
    self._value += six.int2byte(random.randint(0, 255))
TypeError: can only concatenate str (not "bytes") to str
```
Changing `value` of `s_random` to `str` or `bytes` type fails as well.

cq674350529:
@kleest @jtpereyda This seems to be an issue the implementation of `s_random()`.

https://github.com/jtpereyda/boofuzz/blob/0e48caa4e022e6e35908a19bdb988fec31b98e64/boofuzz/primitives/random_data.py#L10-L37

First, it wraps the value with `str()`. 

https://github.com/jtpereyda/boofuzz/blob/0e48caa4e022e6e35908a19bdb988fec31b98e64/boofuzz/primitives/random_data.py#L74-L78

 Then, in the `mutate()`, `self._value` is initialized with `""`. However, the `type(six.int2byte())` is `bytes` in python3 env.

jtpereyda:
[...]

PR #292, opened 2019-06-20
Fixed s_random value type

SR4ven:
Possibly a fix for #290 
@kleest @cq674350529 please test and verify.

### DIFF
--- a/boofuzz/primitives/random_data.py
+++ b/boofuzz/primitives/random_data.py
@@ -72,7 +72,7 @@ class RandomData(BasePrimitive):
             length = self.min_length + self._mutant_index * self.step
 
         # reset the value and generate a random string of the determined length.
-        self._value = ""
+        self._value = b""
         for i in xrange(length):
             self._value += six.int2byte(random.randint(0, 255))
 
--- a/boofuzz/primitives/random_data.py
+++ b/boofuzz/primitives/random_data.py
@@ -1,9 +1,9 @@
 import random
-import six
 
-from builtins import chr
+import six
 from past.builtins import xrange
 
+from boofuzz import helpers
 from .base_primitive import BasePrimitive
 
 
@@ -33,7 +33,7 @@ class RandomData(BasePrimitive):
 
         super(RandomData, self).__init__()
 
-        self._value = self._original_value = str(value)
+        self._value = self._original_value = helpers.str_to_bytes(value)
         self.min_length = min_length
         self.max_length = max_length
         self.max_mutations = max_mutations
--- a/unit_tests/test_blocks.py
+++ b/unit_tests/test_blocks.py
@@ -114,7 +114,7 @@ def repeaters():
         s_word(0xdead, name="word", fuzzable=False)
         s_dword(0xdeadbeef, name="dword", fuzzable=False)
         s_qword(0xdeadbeefdeadbeef, name="qword", fuzzable=False)
-        s_random(0, 5, 10, 100, name="random", fuzzable=False)
+        s_random("0", 5, 10, 100, name="random", fuzzable=False)
         s_block_end()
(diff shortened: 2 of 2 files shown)


cq674350529:
@SR4ven It seems to work. Thanks.

PS: First I wonder why it works because the type of `self._original_value` differs from that of `self._value`.  
In my view, I would like to change `self._value = self._original_value = str(value)` to `self._value = self._original_value = helpers.str_to_bytes(value)`. Actually, it's already done in the `BasePrimitive`.

https://github.com/jtpereyda/boofuzz/blob/0e48caa4e022e6e35908a19bdb988fec31b98e64/boofuzz/primitives/base_primitive.py#L62-L86

**In addition, there are many cases like this in other places.**

https://github.com/jtpereyda/boofuzz/blob/0e48caa4e022e6e35908a19bdb988fec31b98e64/boofuzz/utils/scada.py#L19-L20

cq674350529:
Forgot to say, `six.binary_type(b"")` is used in other places, as well as `b""`.  I'm not familiar with the py2 and py3 compatibility.

SR4ven:
We could use `self._value = self._original_value = helpers.str_to_bytes(value)` and `six.binary_type(b"")` but it appears to make no difference. For type consistency reasons, we should probably do this.
@jtpereyda what do you think?

jtpereyda:
`self._value = self._original_value = helpers.str_to_bytes(value)` seems reasonable enough. One of the effects of using `str(value)` is that it would normalize some types to `str`, but I don't know that that's particularly important or relevant. By using `helpers.str_to_bytes(value)`, the type will accept either `bytes` or `str` which seems good enough to me.

From the [six docs](https://buildmedia.readthedocs.org/media/pdf/six/latest/six.pdf): 

> six.binary_type
> Type for representing binary data. This is str in Python 2 and bytes in Python 3. Python 2.6 and 2.7 include
> bytes as a builtin alias of str, so six’s version is only necessary for Python 2.5 compatibility

So `b""` should get us enough compatibility -- v2.5 isn't a compatibility target.

jtpereyda:
Thank you @SR4ven and @cq674350529 ! Once discussed changes are implemented, we could merge. Has anyone tested the use case from the original issue?

SR4ven:
I have to correct myself, `self._value = self._original_value = helpers.str_to_bytes(value)` works when running a fuzz session, but fails the unit tests.
```python
_____________________________ ERROR collecting unit_test.py ______________________________
unit_test.py:5: in <module>
    unit_tests.test_blocks.run()
unit_tests/test_blocks.py:11: in run
    repeaters()
unit_tests/test_blocks.py:121: in repeaters
    data = s_render()
boofuzz/__init__.py:113: in s_render
    return blocks.CURRENT.render()
boofuzz/blocks/request.py:155: in render
    self._rendered += item.render()
boofuzz/blocks/block.py:232: in render
    self._rendered += item.render()
E   TypeError: cannot concatenate 'str' and 'int' objects
------------------------------------ Captured stdout -------------------------------------
PRIMITIVE MUTATION COUNTS (SIZES):
	delim:  44	(3979)
	string: 1441	(65927421)
	byte:   112
	word:   140
	dword:  140
	qword:  140
	sizer:  140
```

Running the fuzz session provided in the issue works with both approaches.

jtpereyda:
See line 121 in test_blocks.py
```
        s_random(0, 5, 10, 100, name="random", fuzzable=False)
```
With the new change, the first argument would need to be `"0"` instead of `0`.

SR4ven:
@jtpereyda thanks for the hint! Now it works.

kleest:
With the latest changes, both the tests and my use case work fine. Thanks!

jtpereyda:
Thanks all for the PR and fix!

PR #295, opened 2019-06-26
More python 3 compatibility

aedrax:
I've come across more locations where default values need to be forced as a byte type for compatibility.

### DIFF
--- a/boofuzz/primitives/random_data.py
+++ b/boofuzz/primitives/random_data.py
@@ -72,7 +72,7 @@ class RandomData(BasePrimitive):
             length = self.min_length + self._mutant_index * self.step
 
         # reset the value and generate a random string of the determined length.
-        self._value = b""
+        self._value = six.binary_type(b"")
         for i in xrange(length):
             self._value += six.int2byte(random.randint(0, 255))
 
--- a/boofuzz/primitives/random_data.py
+++ b/boofuzz/primitives/random_data.py
@@ -72,7 +72,7 @@ class RandomData(BasePrimitive):
             length = self.min_length + self._mutant_index * self.step
 
         # reset the value and generate a random string of the determined length.
-        self._value = six.binary_type(b"")
+        self._value = b""
         for i in xrange(length):
             self._value += six.int2byte(random.randint(0, 255))
 
--- a/utils/pdml_parser.py
+++ b/utils/pdml_parser.py
@@ -59,13 +59,14 @@ class ParsePDML(ContentHandler):
 
     # noinspection PyMethodMayBeStatic
     def get_string(self, parsed):
-
-        parsed = parsed.replace("\t", "")
-        parsed = parsed.replace("\r", "")
-        parsed = parsed.replace("\n", "")
-        parsed = parsed.replace(",", "")
-        parsed = parsed.replace("0x", "")
-        parsed = parsed.replace("\\x", "")
+        # until this becomes more universal, utf-8 is assumed
+        parsed = six.binary_type(parsed, "utf-8")
+        parsed = parsed.replace(b"\t", b"")
+        parsed = parsed.replace(b"\r", b"")
+        parsed = parsed.replace(b"\n", b"")
+        parsed = parsed.replace(b",", b"")
+        parsed = parsed.replace(b"0x", b"")
+        parsed = parsed.replace(b"\\x", b"")
 
         value = six.binary_type(b"")
         while parsed:
@@ -78,12 +79,12 @@ class ParsePDML(ContentHandler):
 
             value += six.int2byte(hex_pair)
 
-        value = value.replace("\t", "")
-        value = value.replace("\r", "")
-        value = value.replace("\n", "")
-        value = value.replace(",", "")
-        value = value.replace("0x", "")
-        value = value.replace("\\x", "")
+        value = value.replace(b"\t", b"")
+        value = value.replace(b"\r", b"")
+        value = value.replace(b"\n", b"")
+        value = value.replace(b",", b"")
+        value = value.replace(b"0x", b"")
+        value = value.replace(b"\\x", b"")
 
         return value
 
--- a/utils/pdml_parser.py
+++ b/utils/pdml_parser.py
@@ -69,7 +69,7 @@ class ParsePDML(ContentHandler):
         parsed = parsed.replace(b"0x", b"")
         parsed = parsed.replace(b"\\x", b"")
 
-        value = six.binary_type(b"")
+        value = b""
         while parsed:
             pair = parsed[:2]
             parsed = parsed[2:]
--- a/utils/pdml_parser.py
+++ b/utils/pdml_parser.py
@@ -8,6 +8,7 @@ from xml.sax.handler import feature_namespaces
 
 import six
 
+from boofuzz import helpers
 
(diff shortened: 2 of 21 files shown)


boofuzz/primitives/random_data.py
```diff
@@ -72,7 +72,7 @@ def mutate(self):
             length = self.min_length + self._mutant_index * self.step
 
         # reset the value and generate a random string of the determined length.
-        self._value = b""
+        self._value = six.binary_type(b"")
```
cq674350529: Actually, using `b""` is enough.

utils/pdml_parser.py
```diff
@@ -67,7 +67,7 @@ def get_string(self, parsed):
         parsed = parsed.replace("0x", "")
         parsed = parsed.replace("\\x", "")
 
-        value = ""
+        value = six.binary_type(b"")
```
cq674350529: The following `value.replace()` statements should be updated correspondingly. Otherwise, `TypeError` will occur in py3 env.

boofuzz/primitives/random_data.py
aedrax: True, but for consistency and explicitness is why I went with using the binary_type

utils/pdml_parser.py
aedrax: Good call I'll update that

utils/pdml_parser.py
```diff
@@ -59,15 +59,16 @@ def endElement(self, name):
...
-        parsed = parsed.replace("\r", "")
-        parsed = parsed.replace("\n", "")
-        parsed = parsed.replace(",", "")
-        parsed = parsed.replace("0x", "")
-        parsed = parsed.replace("\\x", "")
-
-        value = ""
+        # until this becomes more universal, utf-8 is assumed
+        parsed = six.binary_type(parsed, "utf-8")
+        parsed = parsed.replace(b"\t", b"")
```
cq674350529: In py2, using `six.binary_type(parsed, "utf-8")` will cause `TypeError`. Maybe use the buildin `helpers.str_to_bytes()` instead.

aedrax: I added that helper, do you currently use any of this utility scripts? I personally don't and often forget about them

cq674350529: Me too. These scripts are copied from `Sulley` and only used for special cases.

driechersRC:
When running with python3 I seem to run out of memory easily. I have over 1 Gig of free memory and I get this error when I try to send a message just over 2Mb long. Is there anything that can be done to reduce the footprint in the python3 compatibility code? I'm sure it is not the most efficient in python2 either but I only run into the crash with python3.
```
  File "../boofuzz/boofuzz/sessions.py", line 590, in fuzz
  File "../boofuzz/boofuzz/sessions.py", line 695, in _main_fuzz_loop
  File "../boofuzz/boofuzz/sessions.py", line 1424, in _fuzz_current_case
    self.transmit_fuzz(target, self.fuzz_node, path[-1], callback_data=callback_data)
  File "../boofuzz/boofuzz/sessions.py", line 1103, in transmit_fuzz
    data = node.render()
  File "../boofuzz/boofuzz/blocks/request.py", line 154, in render
    self._rendered += item.render()
  File "../boofuzz/boofuzz/blocks/block.py", line 231, in render
    self._rendered += item.render()
  File "../boofuzz/boofuzz/primitives/base_primitive.py", line 72, in render
    value = self._render(self._value)
  File "../boofuzz/boofuzz/primitives/string.py", line 322, in _render
    value = helpers.str_to_bytes(value)
  File "../boofuzz/boofuzz/helpers.py", line 466, in str_to_bytes
    result = six.binary_type().join(temp)
MemoryError
```

driechersRC:
Looks like there may be some more in fuzz_logger_text. It may be worth testing with each of the different fuzz loggers.

```
  File "../boofuzz/boofuzz/sessions.py", line 695, in _main_fuzz_loop
  File "../boofuzz/boofuzz/sessions.py", line 1394, in _fuzz_current_case
    current_num_mutations=self.fuzz_node.num_mutations())
  File "../boofuzz/boofuzz/fuzz_logger.py", line 69, in open_test_case
    fuzz_logger.open_test_case(test_case_id=test_case_id, name=name, index=index, *args, **kwargs)
  File "../boofuzz/boofuzz/fuzz_logger_text.py", line 61, in open_test_case
    msg_type='test_case')
  File "../boofuzz/boofuzz/fuzz_logger_text.py", line 79, in _print_log_msg
    file=self._file_handle)
TypeError: a bytes-like object is required, not 'str'

During handling of the above exception, another exception occurred:

  File "../boofuzz/boofuzz/sessions.py", line 590, in fuzz
  File "../boofuzz/boofuzz/sessions.py", line 719, in _main_fuzz_loop
    "Unexpected exception! {0}".format(traceback.format_exc()))
  File "../boofuzz/boofuzz/fuzz_logger.py", line 37, in log_error
    fuzz_logger.log_error(description=description)
  File "../boofuzz/boofuzz/fuzz_logger_text.py", line 45, in log_error
    msg_type='error')
  File "../boofuzz/boofuzz/fuzz_logger_text.py", line 79, in _print_log_msg
    file=self._file_handle)
TypeError: a bytes-like object is required, not 'str'
```

jtpereyda:
Has anyone given these changes a sanity check functional test on Py2 and/or Py3? @aedrax ?

@driechersRC is that TypeError a result of this PR or was it already existing?

For future reference, we will use the standard language feature `b''` type instead of `six.binary_type`. Python 2.5 compatibility isn't a goal for us. It's functionally fine either way though so this is fine.

jtpereyda:
@driechersRC Regarding the MemoryError, that may be worth opening an issue for. Without knowing the root cause or difference in amount of memory used, it's hard to say whether someone will take the time to investigate and fix. In general though, it seems like boofuzz could use less memory than it does.

driechersRC:
@jtpereyda The TypeError was testing off of  d1324f61162397749216a06b642805606117f181 which was master on Jun 25. I just created #303 per your request.

aedrax:
@jtpereyda I have been using it on my own system without any issues, but that's about as rigorous of a test I've done beyond stuff in tox. Also, I'll update to move away from six.binary_type

jtpereyda:
I'll merge after the six.binary_type updates. @driechersRC TypeError should be tracked, so another issue may be appropriate if it's not being addressed right away.

aedrax:
@jtpereyda the six.binary_type updates are in, and it should be good to merge in now. There is one reference to it still because it's actually looking at the type.

SR4ven:
@aedrax friendly reminder that there are a few tiny code style violations in the code.
```python
./utils/pdml_parser.py:13:1: E302 expected 2 blank lines, found 1
./unit_tests/test_socket_connection.py:343:40: E127 continuation line over-indented for visual indent
./unit_tests/test_socket_connection.py:431:23: E222 multiple spaces after operator
./unit_tests/test_socket_connection.py:585:23: E222 multiple spaces after operator
```
Code style isn't a hard requirement so fixing this is optimal I guess.

aedrax:
oh sorry about that, @SR4ven what are you using for style testing?

jtpereyda:
Looking good -- thank you @aedrax @SR4ven @driechersRC !

PR #298, opened 2019-07-04
Push Tornado, unpin Flask, add favicon

### LINKED ISSUE #208: un-pin Flask

SR4ven:
Tornado verion 5.x is the highest one that still supports Python 2.7, so let's use that.
The only downside is that when not providing a favicon, a warning is displayed. I couldn't find a way to suppress just this one message, so I simply added a favicon.

Flask is now unpinned too, which won't change anything as the version stays the same. No idea why it was pinned in the first place. Fixes #208


jtpereyda:
Thank you for the PR! We should find an image with explicit copyright license at some point.

PR #300, opened 2019-07-11
Fix broken TLS

keisentraut:
Fuzzing a SSL/TLS enabled server did not work anymore. I got this error:
```
  File "fuzz.py", line 26, in <module>
    session.fuzz()
  File "/usr/lib/python2.7/site-packages/boofuzz/sessions.py", line 598, in fuzz
  File "/usr/lib/python2.7/site-packages/boofuzz/sessions.py", line 703, in _main_fuzz_loop
  File "/usr/lib/python2.7/site-packages/boofuzz/sessions.py", line 1419, in _fuzz_current_case
    self._open_connection_keep_trying(target)
  File "/usr/lib/python2.7/site-packages/boofuzz/sessions.py", line 1457, in _open_connection_keep_trying
    target.open()
  File "/usr/lib/python2.7/site-packages/boofuzz/sessions.py", line 87, in open
    self._target_connection.open()
  File "/usr/lib/python2.7/site-packages/boofuzz/socket_connection.py", line 171, in open
    ssl_sock = ssl.SSLContext.wrap_socket(self._sock)
TypeError: unbound method wrap_socket() must be called with SSLContext instance as first argument (got _socketobject instance instead)
```
The solution was simple: Use ```ssl.wrap_socket``` instead of ```ssl.SSLContext.wrap_socket```.

In order to reproduce, please use this script:
```
#!/usr/bin/env python2
from boofuzz import *

session = Session(
    target=Target(
        connection=SocketConnection(
            host="127.0.0.1",
            port=12345,
            proto='ssl', 
            ),
    ),
)
s_initialize("A")
s_byte(0x41)
session.fuzz()
```
and run it against a test server with
```
openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365 -nodes
openssl s_server -accept 12345 -key key.pem -cert cert.pem
```


jtpereyda:
Thank you for the PR @keisentraut !

The [docs](https://docs.python.org/3/library/ssl.html) say the `wrap_socket` method is deprecated in favor of the `SSLContext.wrap_socket` approach. So this fix is more functional but not quite recommended.

Based on the error, it seems like we need to create a context object. See https://docs.python.org/3/library/ssl.html#socket-creation

If we can use the new approach and make it work without too much trouble, that would be best. Cheers!

keisentraut:
I've now rewritten it and you can pass an arbitray SSLContext now (instead of server_key/server_cert only) like before.

I wasn't sure what to do when the user does neither give a SSLContext nor a server_hostname.
The only option would be to accept all server certificates, but I do not like APIs with unsafe defaults.
Therefore, it will now fail with an assert(...). If you want to accept invalid remote certificates (which is probably common for fuzzing), then you can simply pass an appropriate SSLContext.

I've added examples how to use SSL/TLS, too.

boofuzz/socket_connection.py
```diff
@@ -93,8 +93,10 @@ def __init__(self,
         self.l2_dst = l2_dst
         self._udp_broadcast = udp_broadcast
         self.server = server
-        self.keyfile = keyfile
-        self.certfile = certfile
+        self.sslcontext = sslcontext
+        self.server_hostname = server_hostname
+        assert(not self.server or self.sslcontext, "parameter sslcontext is required when server=True.")
+        assert(self.sslcontext or self.server, "You did neither specify a sslcontext nor server_hostname. Can't proceed without making you vulnerable to MITM attacks. If this is what you want, then specify a sslcontext.")
```
jtpereyda: 1. Looks like these checks should only apply when the protocol is `ssl`
2. Use `raise ValueError...` to match current approach. See example a few lines below in the code: `            raise ValueError("__init__() argument port required for protocol {0}".format(self.proto))`
3. Be explicit about what is being checked. E.g. `self.server is not None and self.sslcontext is None`
4. The error explanations are a bit confusing. Try rewording them for clarity. It is helpful to use the same format for each so that the semantic difference is obvious. You might also match the ValueError example I cited. E.g., `__init__() argument sslcontext is required when server == True`.
5.  Should the second assert `assert(self.sslcontext or self.server, ...` be checking `self.server_name` instead of `self.server`?

boofuzz/socket_connection.py
```diff
@@ -165,13 +167,23 @@ def open(self):
...
+                        self.sslcontext.check_hostname = True
+                        self.sslcontext.verify_mode = ssl.CERT_REQUIRED
+                    else:
+                        # Only option left would be to accept any certificates.
+                        # This would left us vulnerable to man-in-the-middle attacks.
+                        # We won't allow that and the user has to remove the assert below.
+                        assert(False)
+                        self.sslcontext.check_hostname = False
+                        self.sslcontext.verify_mode = ssl.CERT_NONE
+            self._sock = self.sslcontext.wrap_socket(self._sock, server_side=self.server, server_hostname=self.server_hostname)
```
jtpereyda: Let's keep `if self.server: ... else: ...` approach. I like that it makes very clear what the behavior of the client is and what the behavior of the server is.

For this PR, are you trying to address the server behavior or client behavior?

boofuzz/socket_connection.py
```diff
@@ -165,13 +167,23 @@ def open(self):
...
+                if not self.server and not self.sslcontext:
+                    self.sslcontext = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+                    if self.server_hostname:
+                        self.sslcontext.check_hostname = True
+                        self.sslcontext.verify_mode = ssl.CERT_REQUIRED
+                    else:
+                        # Only option left would be to accept any certificates.
+                        # This would left us vulnerable to man-in-the-middle attacks.
+                        # We won't allow that and the user has to remove the assert below.
+                        assert(False)
```
jtpereyda: Let's avoid requiring code modification just to use the framework. Sounds like the user already needs to make their own context anyway.

boofuzz/socket_connection.py
keisentraut: This PR is for both: SSL server and client fuzzing. I do not need to care about the value of ```self.server```, because the ```wrap_socket``` method is exactly the same for both cases. See simplified code in next commit.

boofuzz/socket_connection.py
keisentraut: 1. yes
2. ValueError is of course the better way, I currently need to fuzz a TLS server/client and did some "hacking"... changed it, see below
3. done (flake8 does report this, too)
4. I'm not native english speaker, hopefully it's clearer now!
5. correct, that was a mistake/bug

boofuzz/socket_connection.py
keisentraut: Yes, the user will need to make his own context, if he wants to accept any certificates. This is also explained in the ```fuzz-ssl-server.py``` example.

keisentraut:
I'm confused about how to proceed with the ```stickler-ci``` warning. I merged your master into my tls-fix branch, but it still doesn't work. I assume it's not the fault of my pull request, but rather ongoing effort at your side?

SR4ven:
@keisentraut you are safe to ignore this check for now.
Stickler is a code style checker, but it doesn't have a config file yet. This will be fixed when #305 is merged.

keisentraut:
From my side, this code is now ready to get merged then. Please review again! 
To clarify again: It works for both server- and client-side SSL/TLS fuzzing, please see the files ```examples/fuzz-ssl-server.py``` and ```examples/fuzz-ssl-client.py```. 

@SR4ven: I could theoretically write a unit test, but how would one test a SSL/TLS SocketConnection without a defined endpoint to connect to?

SR4ven:
@keisentraut I didn't mention a unit test here but in #302 as it's a new primitive.

But having a unit-test for this too wouldn't hurt either I guess.
A possible endpoint could be something like [this](https://blog.anvileight.com/posts/simple-python-http-server/) or [this](https://gist.github.com/dergachev/7028596).

keisentraut:
To be honest, I don't really know how to write a unit test because the whole folder is cluttered with unused code. However, the current diff is pretty straightforward and there wasn't a SSL/TLS unit test before, too (otherwise, you wouldn't have had this bug at all).

If you insist on a unit test, all you'd need to do is to run the ```openssl``` commands in the two SSL/TLS examples and then run the two examples itself!

keisentraut:
Please tell me if something's missing to get this merged, thx.

starek4:
@jtpereyda to be honest, it would be great to merge this already. I am using the latest master version of Boofuzz in my project for master thesis, because of several issues which were fixed and not released yet. And TLS testing is quite important for me. Thanks for any new comment about what should be done in this pull request to merge it into the master branch 👍

jtpereyda:
Thank you everybody for the work on this one, and my apologies for the delay. I'm going ahead with the merge even though I'm not sure about all the details yet. I suspect some of the details could change in the future, but at least we are moving forward.

Thank you @keisentraut @SR4ven @starek4

jtpereyda:
@starek4 looking forward to seeing your thesis when it comes out :)

keithmendozasr:
@starek4 I ran into the same issue in Windows. What worked for me was to pass a send_timeout and recv_timeout arguments with 6000.0. I have a feeling that Window's setsockopt() implementation for SO_RCVTIMEO and SO_SNDTIMEO is either not a struct timeval as defined by the Posix standard; or whatever _second_to_second_microsecond_struct() returns is not translating nicely to the Windows implementation of struct timeval.

starek4:
Interesting, I think this can be also related to https://github.com/jtpereyda/boofuzz/issues/281 (just for the reference).

Many thanks for the workaround!

keisentraut:
@starek4: Sry for the late answer. I didn't test on Windows, only on Arch and Ubuntu Linux.

starek4:
@keisentraut no problem, already fixed by https://github.com/jtpereyda/boofuzz/pull/340  👍

PR #301, opened 2019-07-11
.gitignore: add boofuzz-results

keisentraut:
```boofuzz-results/``` should be in ```.gitignore```, too.


jtpereyda:
Fantastic. Thank you for the PR @keisentraut !

Issue #144, opened 2017-04-29
Make s_bytes


jtpereyda:
The only way to create a field of an odd number of bytes is to use s_bit_field. Create an s_bytes or something that lets you specify an arbitrary byte length.

Issue created based on this mailing list question: https://groups.google.com/d/msg/boofuzz/RfTg3biZOIY/aMK7U8rBAAAJ

jtpereyda:
Issue #131 seems to be a similar request

SR4ven:
Implemented in #302

PR #302, opened 2019-07-11
add primitive s_bytes: fuzz arbitrary binary data

keisentraut:
This was requested in issue #144. 

The code is kind of cumbersome because of the cumbersome API of BasePrimitive.
However, I've tested it on Python 2 and Python 3 and it seems to work.

The fuzzing strategy for an arbitrary length, binary string is:
- replace binary value with list of static binary strings
- replace each 1 byte block in original value with list of static 1byte strings
- replace each 2 byte block in original value with list of static 2byte strings
- replace each 4 byte block in original value with list of static 4byte strings

I've compiled the list of interesting strings based on [the wikipedia list of magic debug values](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_debug_values) and some others based on my gut feeling. Please feel free to change them.

Minimal sample to test it:
```
#!/usr/bin/env python
from boofuzz import *

s_initialize("A")
s_bytes(b"0000000000000000000")
session.fuzz()
```
and run ```while [ True ]; do nc 127.0.0.1 12345 | hexdump -C; sleep 0.1; done``` in another terminal.


jtpereyda:
@keisentraut Thank you, very cool!

The code is a bit messy which could come back to us later, but it's fine for now.

The API is indeed pretty cumbersome :( Hopefully that can change one day.

jtpereyda:
Oh, please add a new feature note to CHANGELOG.rst. Thanks!

SR4ven:
@keisentraut good job!
Have you considered writing a unit-test? Doesn't have to be in this PR of course, but it would be nice to have.

Also please have a look at the code style checker [here](https://travis-ci.org/jtpereyda/boofuzz/jobs/557380360). There a few syntax related PEP8 violations in your code which should be really easy to fix.

@jtpereyda what do you think about making tox fail on new code style violations? That would make it more obvious to the contributors that something is not completely right with the code. Also if a test fails because of this, you can always look at the logs, check the unit-tests and still merge it.

boofuzz/__init__.py
```diff
@@ -621,6 +622,27 @@ def s_byte(value, endian=LITTLE_ENDIAN, output_format="binary", signed=False, fu
     byte = primitives.Byte(value, endian, output_format, signed, full_range, fuzzable, name)
     blocks.CURRENT.push(byte)
 
+def s_bytes(value, size=-1, padding=six.binary_type(b"\x00"), fuzzable=True, max_len=-1, name=None):
+    """
+    Pushes a bytes field onto the current block stack.
```
cq674350529: Great job! Maybe using `b""` directly, as discussed in other place.

> For future reference, we will use the standard language feature b'' type instead of six.binary_type. Python 2.5 compatibility isn't a goal for us. It's functionally fine either way though so this is fine.

boofuzz/primitives/bytes.py
```diff
@@ -0,0 +1,248 @@
...
+            b'\x7F\xFF\xFF\xFF',
+            b'\xFF\xFF\xFF\x7F',
+            b'\xFF\xFF\xFF\xFF',
+            ] + [i for i in _magic_debug_values if len(i) == 4]
+    _len_fuzz_strings_4byte = len(_fuzz_strings_4byte)
+
+
+    def __init__(self, value, size=-1, padding=six.binary_type(b"\x00"), fuzzable=True, max_len=-1, name=None):
+        """
+        Primitive that fuzzes a binary byte string with arbitrary length.
```
cq674350529: Maybe using `b""` directly .

boofuzz/primitives/bytes.py
```diff
@@ -0,0 +1,248 @@
...
+        return num
+
+    def _render(self, value):
+        """
+        Render string value, properly padded.
+        """
+
+        if isinstance(value, six.text_type):
+            value = helpers.str_to_bytes(value)
+
```
cq674350529: It seems the condition will be checked again in `helpers.str_to_bytes()`. Maybe using `value = helpers.str_to_bytes(value)` without the condition directly.

keisentraut:
Resolved most of the issues above.

I'm not sure what the ```max_len``` and ```size``` parameters are supposed to do. If I simply skip testcases in the ```mutate(...)``` function as I do now, it's hard to calculate a correct number for ```num_mutations``` without calculating all mutations first.

But if I skip them, the ```num_mutations``` is now only an upper limit, therefore, boofuzz might stop after 156 of 189 test cases. That's ugly, but I don't know how to resolve this. What's your opinion on this?

PS: The 80 characters limit of pycodestyle is IMHO too short, would need to break almost every line.
And we are not using IBM punch cards anymore where this 80 chars limit originally came from...

SR4ven:
@keisentraut the current max line lenght is set to 120. 80 is indeed a bit short but it is still the GitHub review window width. 120 seems to be a common standard.
https://github.com/jtpereyda/boofuzz/blob/0b1a0d705afd00f794e3caf105f5f363c01edd5c/setup.cfg#L11

There are still some [warnings](https://travis-ci.org/jtpereyda/boofuzz/jobs/557785448) present, but this repo doesn't enforce code style at the moment, so fixing them is optimal I guess.
```
4     E123 closing bracket does not match indentation of opening bracket's line
5     E126 continuation line over-indented for hanging indent
9     E226 missing whitespace around arithmetic operator
4     E501 line too long (125 > 120 characters)
3     F401 '.bytes.Bytes' imported but unused
```

keisentraut:
@SR4ven: ok, will fix it. The problem was I used```pycodestyle``` instead of ```flake``` and this did not have any warnings anymore (besides >80 chars)

jtpereyda:
@keisentraut 
> I'm not sure what the max_len and size parameters are supposed to do. If I simply skip testcases in the mutate(...) function as I do now, it's hard to calculate a correct number for num_mutations without calculating all mutations first.
>
> But if I skip them, the num_mutations is now only an upper limit, therefore, boofuzz might stop after 156 of 189 test cases. That's ugly, but I don't know how to resolve this. What's your opinion on this?

`size` makes every mutation the same length. `max_len` limits the maximum length of any given mutation.

`max_len` implies truncating or skipping test cases that are too long. Check out `string.py` for an example -- in that case everything is computed beforehand and oversized strings are removed.

In your case, you could simply truncate oversized strings. This narrows down the possibilities and would make it `num_mutations` relatively simple. The disadvantage is the risk of repeating test cases. Yet another advantage is that you can still leverage useful strings that might be too long. Thoughts?

As an aside, the expectation that every primitive/block will report its number of mutations ahead of time places a burden on implementation. I'm not sure if the information gain of knowing how many test cases you have is worth the cost of implementation.

jtpereyda:
@SR4ven Regarding fail-on-style-violation -- I agree we would be helped by raising the visibility of warnings. If it's possible to fail only on new violations, that might be an OK way. There might be other CI services out there that will help out in a smoother way. Might be worth a separate issue to talk over this point.

keisentraut:
Please review again (and merge if appropriate), thanks.

unit_tests/primitives.py
```diff
@@ -97,6 +98,28 @@ def s_mirror_tests():
...
+        tmp = req.names['bytes_max_len'].render()
+        assert(len(tmp) <= 5)
+
+    s_initialize("test_bytes_size")
+    s_bytes(b"1234567", name="bytes_size", size=7, padding=b"A")
+    req = s_get('test_bytes_size')
+    while s_mutate():
+        tmp = req.names['bytes_size'].render()
+        assert(len(req.names['bytes_size'].render()) == 7)
+
```
cq674350529: It seems the `tmp` is useless .

keisentraut: fixed, please see next commit

keisentraut:
Please tell me if something's missing for this to get merged, thx.

SR4ven:
@jtpereyda seems to be very busy at the moment, if there're any outstanding requests, you will get to know eventually. :D

@keisentraut, I just noticed that this PR does not fully meet the [black ](https://github.com/psf/black)code style, which is understandable as we only just implemented this. Also, there is a problem with the auto fixer when the PR does not allow edits from maintainers https://github.com/jtpereyda/boofuzz/issues/304#issuecomment-520495330.
To fix this, install `black` either with pip directly or with `pip install -e .[dev]` in your boofuzz directory and execute `black .`

It would be nice if you could check your other PR too.

jtpereyda:
This looks great! High quality PR especially with the UTs. My apologies for the delay and thank you for your patience @keisentraut and reviewers!

Issue #304, opened 2019-07-14
Stickler-CI and Black for a better code style consistency


SR4ven:
Currently most of the code is PEP8 compliant, but upcoming PRs might unknowingly violate the style guide. Running `tox` will notify about this, but will still pass the test and as nobody looks at a passed test log, the author might not be aware of any issues.

To raise the visibility of code style violations, we should add some kind of CI service or test.
I did some research and found [Stickler-CI](https://stickler-ci.com/) and [Black](https://github.com/python/black). Sticker uses the new GitHub-Checks API, so style violations will be displayed similar as code review comments which I find is a very fancy and productive way. See example [here](https://github.com/markstory/lint-review/pull/251/checks) and [here](https://github.com/markstory/lint-review/pull/251/files#annotation_22986266).

Sticker offers multiple code style checkers, of which I found Black to be both easy to use and to offer the best result. The tool is not only able check for style violations, but also fix them.
In combination with Stickler we have the option to automatically push a commit with code style fixes after a PR has been opened. This is optional of course, but might help saving time because the author doesn't have to refactor his code in most cases where no manual changes are required.
On top we could include Black in tox so that when you run the unit-tests, your code gets restyled automatically. There are [editor integrations](https://black.readthedocs.io/en/stable/editor_integration.html) too.

Black does not only help to meet PEP8, but to maintain a general code style consistently e.g. list formatting. Check the upcoming PR for examples, in my opinion it increases readability greatly and if not, you can always disable it with `# fmt: off` and `# fmt: on`.

I will open a PR with all necessary config files and initial style changes so that you @jtpereyda only have to setup Stickler-CI.
Things we need to discuss:
- Automatically commit fix when a PR is created? (+Less work -not everyone will like a bot messing with the code)
- Automatically run Black with tox?
- Use multiple linters in Stickler like flake8?

SR4ven:
@jtpereyda I'm not exactly sure whether to add a black style check to tox or not.
As you already mentioned, it might make changes pre-commit and mess with your style, without having the option to revert. We still have Stickler-CI which will auto-commit a fix after pushing anyway.

On top, we wouldn't want travis to run the black style check in tox as we have Stickler for that (keep style and other tests in separate checks). Travis exports a `CI` env-var which could be used for that, but checking this isn't really nice in tox as we need to use bash and cmd for that.

SR4ven:
[...]

PR #305, opened 2019-07-14
Added Stickler and Black configuration

SR4ven:
Preparations for #304 

TODO:
- [x] Port existing code to Black standard

Optimal features:
- [x] Automated commit with code style fixes
- [ ] Tox integration


setup.cfg
```diff
@@ -5,7 +5,10 @@ zip_ok = False
 python-file-with-version = boofuzz/__init__.py
 
 [flake8]
-ignore = F403, F405, W503
+ignore = F403, F405, W503, E203
```
jtpereyda: Why ignore E203?

jtpereyda:
@SR4ven This is fantastic. I'll set up Stickler.

The auto-commit for fixes sounds really cool. Let's do it.

The tox integration makes me a bit more nervous as it could edit pre-commit files, but I think it's still worth a shot!

setup.cfg
SR4ven: See https://github.com/python/black#slices

SR4ven:
@jtpereyda I see you already enabled Stickler, but it doesn't have a config file yet, so it fails the whole check.
We have two options now:
1. You merge this PR in the current state without any optimal features or updating existing code. I could take care of that in a seperate PR.
    I couldn't find any info about whether Stickler is checking only the diffs, whole touched files or the whole repo. Without updating the existing code this could possibly result in tests failing because a not modified file is not style conform.
2. I can definetly finish updating the existing code and enable the auto commit feature today.
    I need to think of a way to have Travis ignore Black in tox first before I can add that. I'll do this in another PR.

If getting current CI tests to pass is not that urgent, option 2 might be the safer one.
On the other hand it would be nice to know if Stickler is smart enough to only check the diff. We could easily find that out with option 1. Even if Stickler check all code everytime, it's not too much of a problem as the whole repo should follow the same style guide, which would be ensured that way.
Making the existing code Black conform means a lot of changes, but isn't that much work actually.

SR4ven:
@jtpereyda if you agree with all style changes, we are ready to merge.

jtpereyda:
@SR4ven Sorry for the long delay on this -- merging now!

PR #309, opened 2019-07-24
Fixed concatenation of bytes in PED-RPC

kleest:
On Python3, socket.recv returns bytes. Concatenating bytes to str results
in TypeError.

```
TypeError: can only concatenate str (not "bytes") to str
```


jtpereyda:
@kleest Thank you for the PR!

PR #312, opened 2019-08-03
Updated gitignore

SR4ven:
After updating my system to Ubuntu 19.04, Python3.7 and pip 19.2.1 I noticed a `pip-wheel-metadata` folder appearing. Might be because of tox.
So I used the circumstances to rewrite the .gitignore using this [template](https://github.com/github/gitignore/blob/master/Python.gitignore).

There are some exclusions that are probably not need at the moment, but I guess it doesn't hurt having them and it makes this a little more future proof.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,58 @@
-*.pyc
-archived_fuzzies/*.session
-docs/_build/
+# Byte-compiled / optimized / DLL files
+__pycache__/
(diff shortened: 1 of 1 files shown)


jtpereyda:
@SR4ven Looking good, thank you. Good call switching to a template and heading off a bunch of future issues.

Issue #256, opened 2019-03-10
Add ability to write only failures to disk


just1fi3d:
Would like the ability to have only cases with failures written to disk when using IFuzzLoggers. This would help when reading through the data after fuzzing to reduce time reading through unnecessary data.

SR4ven:
For the database logger this is already implemented. If you set `fuzz_db_keep_only_n_pass_cases = 1` it will only write failed test cases to the db/disk.
For the other loggers such an option doesn't exist yet but there might be a console GUI soon that would make the console view a lot more user friendly.
Or do you need this feature for all loggers in general?

just1fi3d:
I was looking for this for all of the loggers.  Primarily I'm just looking for an easy way of consuming the crash data without having to scan through all of the successful cases too. Ideally I would be able to then output that to disk in some easily consumable format so I could further parse (i.e. csv) however thats not a hard requirement if that info is easily accessible via the gui in the future.

I would want a way to save that data off somehow though.

jtpereyda:
The FuzzLoggerCsv class may help you out -- https://boofuzz.readthedocs.io/en/latest/user/logging.html#csv-logging (I just realized the docs say something about pcaps, but that is misleading)

But it sounds like a way to get only failures output to the other log formats is a needed feature.

The GUI highlights failures on the main index page, which is meant to make it easy to see failure data. Though there's no export right now, so all you'd be able to get is the text output.

just1fi3d:
> The FuzzLoggerCsv class may help you out 
Ya, this is what I have been using so far.

One thing I don't care for about the GUI is that if I step away while fuzzing is running, the web server shuts down after fuzzing is complete. Is there a flag I can pass somewhere to leave it running? That might be a workable solution for me.

SR4ven:
You can use `boo open <database file>` to reopen the webinterface of a completed test. The db file is created in a subdir called boofuzz-results in the current working dir.
If that doesn't work you need to add the boofuzz install dir to the PATH var (for Linux that is). Or just install with `sudo pip install boofuzz`

jtpereyda:
See https://github.com/jtpereyda/boofuzz/issues/166

To do manually, you can do something like the following to the end of your script.
```
while True:
    sleep(.1)
```
this will allow you to Ctrl+C to quit still.

just1fi3d:
> You can use boo open <database file>

[...]

PR #313, opened 2019-08-03
Fixed and documented boo open

SR4ven:
`boo open` was broken in Python3.7, works now.
```python
AttributeError: 'sqlite3.Cursor' object has no attribute 'next'
```
#256 indicated that `boo open` is not documented, so I added a short description to the Quickstart page.

When building the docs locally, I noticed that tox fails when the docs/_build dir still exists. Added a ignore for it to the MENIFEST.in to fix this.

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -65,7 +65,7 @@ current workdir. You can reopen the webinterface on any of those databases at an
     boo open <run-*.db>
 
 To do cool stuff like checking responses, you'll want to use ``post_test_case_callbacks`` in
-:class:`Session <boofuzz.Session>`. You may also want be interested in :ref:`custom-blocks`.
+:class:`Session <boofuzz.Session>`. You may also be interested in :ref:`custom-blocks`.
 
 Remember boofuzz is all Python, so everything is there for your customization.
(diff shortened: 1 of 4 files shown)


docs/user/quickstart.rst
```diff
@@ -58,8 +59,13 @@ After that, you are ready to fuzz: ::
...
 
-To do cool stuff like checking responses, you'll want to use :meth:`Session.post_send <boofuzz.Session.post_send>`.
-You may also want be interested in :ref:`custom-blocks`.
+The log data of each run will be saved to a SQLite database located in the **boofuzz-results** directory at your
+current workdir. You can reopen the webinterface on any of those databases at any time with::
+
+    boo open <run-*.db>
+
+To do cool stuff like checking responses, you'll want to use ``post_test_case_callbacks`` in
+:class:`Session <boofuzz.Session>`. You may also want be interested in :ref:`custom-blocks`.
```
jtpereyda: This was my mistake, but while we're here, might as well change:

```
You may also want be interested in :ref:`custom-blocks`.
```
To
```
You may also be interested in :ref:`custom-blocks`.
```

PR #314, opened 2019-08-03
Fixed receive function in fuzz_logger_curses

SR4ven:
The curses logger uses the wrong `msg_type` for received messages.
https://github.com/jtpereyda/boofuzz/blob/473190604da004ca952616005861c4c0abfa3cf6/boofuzz/helpers.py#L63-L70


jtpereyda:
Thank you for yet another fix @SR4ven !

PR #322, opened 2019-08-16
Disabled stickler fixers

SR4ven:
Temporarily disable stickler fixers as missing push permissions on the fork will result in false positive checks.
See https://github.com/jtpereyda/boofuzz/issues/304#issuecomment-520495330

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -1,10 +1,10 @@
 fixers:
-  enable: true
+  enable: false
 
 linters:
   black:
     config: ./pyproject.toml
-    fixer: true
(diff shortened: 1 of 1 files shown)


SR4ven:
I have just received an e-mail from the Stickler-CI team saying that they will take a look into this.

So maybe give them some more time to fix this before you merge.

jtpereyda:
@SR4ven Roger, will keep open for now.

SR4ven:
The bug was fixed by the Stickler team.
Disabling the fixers is no longer necessary.

PR #323, opened 2019-08-19
Add s_numbers primitive

secenv:
The idea is to use a list of numbers that may cause errors in, say, web applications, or even find bugs and vulnerabilities such as division by zero, integer overflows and similar.

I took most of these from the [Big List of Naughty Strings](https://github.com/minimaxir/big-list-of-naughty-strings) repo, and then some [data type limits](
https://en.wikipedia.org/wiki/Magic_number_(programming)#Data_type_limits).

Similarly to https://github.com/jtpereyda/boofuzz/pull/302, you can test this primitive by running `while [ True ]; do nc -w 1 127.0.0.1 12345; echo; done` in one console and the following code in python:
```from boofuzz import *
s_initialize("A")
s_numbers("0")
session.fuzz()
```

### DIFF
--- a/boofuzz/primitives/numbers.py
+++ b/boofuzz/primitives/numbers.py
@@ -1,18 +1,8 @@
-import six
-
 from .base_primitive import BasePrimitive
-from .. import helpers
 
 
 class Numbers(BasePrimitive):
-    def __init__(self,
-        value,
-        max_len=-1,
-        padding="",
-        signed=False,
-        fuzzable=True,
-        name=None,
-    ):
+    def __init__(self, value, max_len=-1, padding="", signed=False, fuzzable=True, name=None):
         """
         Primitive that cycles through a library of "bad" numbers, represented as strings.
 
@@ -21,7 +11,7 @@ class Numbers(BasePrimitive):
         @type  max_len:       int
         @param max_len:       (Optional, def=-1) Max length of strings returned, leave -1 for any length
         @type  padding:       str
-        @param padding:       (Optional, def="") Return strings left-padded with this string, for use with max_len. Default is " "
+        @param padding:       (Optional, def="") String for left-padding, for use with max_len. Default is " "
         @type  signed:        bool
         @param signed:        (Optional, def=False) Make size signed vs. unsigned
         @type  fuzzable:      bool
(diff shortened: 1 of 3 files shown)


secenv:
@jtpereyda @SR4ven any inputs?

SR4ven:
First of all, thanks for the PR!
I'll add some ideas in the review.

Also, please make sure that the code is [black](https://github.com/psf/black) conform as the style check currently fails.
https://boofuzz.readthedocs.io/en/latest/user/contributing.html#pull-request-checklist
Don't forget to add an entry to the changelog either.

boofuzz/primitives/numbers.py
```diff
@@ -0,0 +1,150 @@
...
+        fuzzable=True,
+        name=None,
+    ):
+        """
+        Primitive that cycles through a library of "bad" numbers, represented as strings.
+
+        @type  value:         str
+        @param value:         Default number value
+        @type  size:          int
+        @param size:          (Optional, def=-1) Static size of this field, leave -1 for dynamic.
```
SR4ven: This seems to be used for padding only. Truncating should be possible too.

boofuzz/primitives/numbers.py
```diff
@@ -0,0 +1,150 @@
...
+                "-,",
+                "-Infinity",
+                "-1#IND",
+            ]
+
+            if signed:
+                for num in signed_nums:
+                    self._fuzz_library.append(num)
+
+            bits = 64
```
SR4ven: Fixed field size conflicts with
https://github.com/jtpereyda/boofuzz/blob/650d41300375dd05e7894175ffe377ba613abc3d/boofuzz/primitives/numbers.py#L22

Too long values should either be truncated later on or not be generated in the first place.

boofuzz/primitives/numbers.py
```diff
@@ -0,0 +1,150 @@
...
+            while bits != 0:
+                self._fuzz_library.append(str(1 << bits))
+                self._fuzz_library.append(str((1 << bits) + 1))
+                self._fuzz_library.append(str((1 << bits) - 1))
```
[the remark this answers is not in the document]
secenv: Actually, I was going to send another PR, for generating integers, based on BitField. We could simply remove those lines and make them part of another module. But my idea was to make this module independent of BitField.

boofuzz/primitives/numbers.py
secenv: Yeah, it was some quick code that I didn't want to think much of, but I guess I could use something like `bits = 64 if size == -1 else size`.

boofuzz/primitives/numbers.py
secenv: Truncating them could work, though it wouldn't make much sense IMO.
I left these params so that they could be used in the future. Didn't want to decide on how to use them...

secenv:
BTW, I've never sent a PR before, so please help me here. Do I only have to apply changes in my repo?

SR4ven:
> 
> 
> BTW, I've never sent a PR before, so please help me here. Do I only have to apply changes in my repo?

Yes, you need to push commits to the `Numbers` branch of your fork in order to make changes to this PR.

boofuzz/primitives/numbers.py
SR4ven: Imagine a case where this field is somewhere in the middle of the frame and because it is not truncated correctly everything behind it gets shifted backwards. This would corrupt the whole structure of the frame. Sometimes a field has to have a fixed size.

If it doesn't make sense to truncate values they should be filtered out, so that only values with the same length or shorter (with padding) are added.

boofuzz/primitives/numbers.py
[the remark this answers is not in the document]
SR4ven: Sorry, I still can't come up with a use case where you would want to keep the values here separated from those in `bit_field`.
What was your idea behind this? Do you have an example?

SR4ven:
`value` is usually the default value and is used when sending not fuzzed and intact data. It is set by `super(Numbers, self).__init__()` as far as I can see.

I find the name `s_numbers` a little misleading, it suggests a binary output format, while it's actually a list of (ascii encoded?) strings.
Wouldn't it be easier to just append those strings to the `s_string` primitive?Though I do have to agree that this could be misleading too.

How about a new/modified primitive like `s_string`, which contains all of the https://github.com/minimaxir/big-list-of-naughty-strings strings?
If we reuse the existing `s_string`, an option to enable the extra test cases should be added.
```python
s_string("Default", extended_list=True)
```

jtpereyda:
Pending discussion in #324

SR4ven:
Closing due to inactivity.
Feel free to reopen/comment!

PR #324, opened 2019-08-27
Add s_integers primitive and Integers class

secenv:
The idea behind this module is to make it simple to get integers in their decimal representation.

To test, execute this script:
```
from boofuzz import *

s_initialize("A")
s_integers(0)
session.fuzz()
```

and then run  `nc 127.0.0.1 12345` in another system console.

### DIFF
--- a/boofuzz/primitives/integers.py
+++ b/boofuzz/primitives/integers.py
@@ -24,4 +24,4 @@ class Integers(BitField):
         self.bits = bits
         self.max_num = max_num
 
-        super(Integers, self).__init__(value, width=bits, signed=signed, name=None, output_format="ascii")
+        super(Integers, self).__init__(value, width=bits, signed=signed, name=name, output_format="ascii")
--- a/boofuzz/primitives/integers.py
+++ b/boofuzz/primitives/integers.py
@@ -1,0 +1,30 @@
+import six
+import struct
+
+from boofuzz.primitives.bit_field import BitField
+
+
+class Integers(BitField):
+    def __init__(self, value, bits=128, max_num=None, signed=True, name=None, *args, **kwargs):
+        """
+        The Integers primitive represents integers of arbitrary value.
+
+        @type  value:          int
+        @param value:          Default integer value
+        @type  bits:           int
+        @param bits:           Width of integer numbers, in bits. Default: 128. Ignored if max_num is set
+        @type  max_num:        int
+        @param max_num:        Maximum number to iterate up to. Takes precedence over bits
+        @type  full_range:     bool
+        @param full_range:     (Optional, def=False) If enabled the field mutates through *all* possible values
+        @type  fuzzable:       bool
+        @param fuzzable:       (Optional, def=True) Enable/disable fuzzing of this primitive
+        @type  name:           str
+        @param name:           (Optional, def=None) Specifying a name gives you direct access to a primitive
+        """
+        self.output_format = "ascii"
+        self.signed = signed
+        self.bits = bits
+        self.max_num = max_num
+
+        super(Integers, self).__init__(value, width=bits, signed=signed, name=None, output_format="ascii")
--- a/boofuzz/primitives/integers.py
+++ b/boofuzz/primitives/integers.py
@@ -1,6 +1,3 @@
-import six
-import struct
-
 from boofuzz.primitives.bit_field import BitField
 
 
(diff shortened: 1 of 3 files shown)


boofuzz/primitives/integers.py
```diff
@@ -0,0 +1,27 @@
...
+        @type  max_num:        int
+        @param max_num:        Maximum number to iterate up to. Takes precedence over bits
+        @type  full_range:     bool
+        @param full_range:     (Optional, def=False) If enabled the field mutates through *all* possible values
+        @type  fuzzable:       bool
+        @param fuzzable:       (Optional, def=True) Enable/disable fuzzing of this primitive
+        @type  name:           str
+        @param name:           (Optional, def=None) Specifying a name gives you direct access to a primitive
+        """
+        self.output_format = "ascii"
```
SR4ven: Assigned but unused.

boofuzz/primitives/integers.py
```diff
@@ -0,0 +1,27 @@
+from boofuzz.primitives.bit_field import BitField
+
+
+class Integers(BitField):
+    def __init__(self, value, bits=128, max_num=None, signed=True, name=None, *args, **kwargs):
```
SR4ven: 128 bit as the default width? Seems a little overkill.

boofuzz/primitives/integers.py
```diff
@@ -0,0 +1,27 @@
...
+        @param fuzzable:       (Optional, def=True) Enable/disable fuzzing of this primitive
+        @type  name:           str
+        @param name:           (Optional, def=None) Specifying a name gives you direct access to a primitive
+        """
+        self.output_format = "ascii"
+        self.signed = signed
+        self.bits = bits
+        self.max_num = max_num
+
+        super(Integers, self).__init__(value, width=bits, signed=signed, name=name, output_format="ascii")
```
SR4ven: Why is `output_format` fixed to "ascii"? Wouldn't it be better to let the user decide which format to use?

SR4ven:
What exactly is the use case of this primitive?

In both of the following cases the exact same output is being generated.
```python
s_integers(0)

s_bit_field(0, width=128, output_format="ascii", signed=True)
```
So can't we simply use `s_bit_field`?

Additionally the `s_integers` doesn't adapt the `name` argument correctly.
```python
s_integers(0, name="Test")
```
Results in
```console
Test Case: 140: A.True.140
```
while it should be
```console
Test Case: 140: A.Test.140
```

jtpereyda:
@secenv This is very cool! The actual fuzz strategy has gotten little attention, so this PR is very much welcome.

I think merging this primitive with `bit_field.py` has a few advantages, as @SR4ven suggested:

1. Less duplicated code: You'll end up reusing some of the same code that BitField already has.
2. Take advantage of multiple encodings: BitField has both binary and ASCII encodings built in.
3. Used by more people: By building all of the fuzzing strategy into one existing primitive, your changes will be available to more people. It will increase the default effectiveness of the numeric types' fuzzing strategy.

BitField is used by the Byte/Word/DWord/QWord types, which means it will also be easy to specify a maximum size.

The idea with a fuzzing framework is to allow for quick prototyping and effective testing, which means we should try to make the default data types as effective as possible.

@secenv thoughts?

secenv:
>What exactly is the use case of this primitive?

The idea was to have a type and a primitive that make it easy to use ascii-encoded integers instead of arrays of bytes. `s_integers` is basically a simplification of the s_bit_field primitive, restricted to outputting ASCII-encoded integer numbers, similar to how byte/word/dword/qword "restrict" BitField. But, yes, it is not necessary on itself.

IMO, it's a bit difficult to see that you can actually do this with BitField, which is why I added these types in boofuzz. Then I thought "why not share it?", modified it a bit and send the PRs.

secenv:
Oh, and BTW, the primary use I thought for this was against web apps, webservers and similar servers that receive ascii-encoded integers instead of network programs that receive a bunch of bytes.

SR4ven:
Closing due to inactivity.
Feel free to reopen/comment!

zhouzoe:
I just added a new primitive(named s_numerical to make muti-fields mutation), and I added some codes in that three files(metioned above) just like what @secenv did . However, when I try to use this primitive, I got a "name 's_numerical' is not defined" error. I confirmed further that when I import the file numerical which is in the catalogue Primitives I also got a "No module named 'boofuzz.primitives.numerical" error.
Why did I get such a error ??? Anyone has  some aolution ???

jtpereyda:
@zhouzoe best to open another issue for a new question like that. See also: http://www.catb.org/~esr/faqs/smart-questions.html#before

PR #325, opened 2019-08-27
Modify BitField to add (+/-) near powers of two...

secenv:
... and surrounding numbers, as @SR4ven said in #323.

### DIFF
--- a/boofuzz/primitives/bit_field.py
+++ b/boofuzz/primitives/bit_field.py
@@ -107,13 +107,32 @@ class BitField(BasePrimitive):
             else:
                 # try only "smart" values.
                 self.add_integer_boundaries(0)
-                self.add_integer_boundaries(self.max_num // 2)
-                self.add_integer_boundaries(self.max_num // 3)
-                self.add_integer_boundaries(self.max_num // 4)
-                self.add_integer_boundaries(self.max_num // 8)
-                self.add_integer_boundaries(self.max_num // 16)
-                self.add_integer_boundaries(self.max_num // 32)
-                self.add_integer_boundaries(self.max_num)
+                self.add_integer_boundaries(self.max_num // 2, 10)
+                self.add_integer_boundaries(self.max_num // 3, 10)
+                self.add_integer_boundaries(self.max_num // 4, 10)
+                self.add_integer_boundaries(self.max_num // 8, 10)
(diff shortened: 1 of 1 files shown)


boofuzz/primitives/bit_field.py
```diff
@@ -106,34 +107,59 @@ def __init__(
...
+                self.add_integer_boundaries(self.max_num // 2, 10)
+                self.add_integer_boundaries(self.max_num // 3, 10)
+                self.add_integer_boundaries(self.max_num // 4, 10)
+                self.add_integer_boundaries(self.max_num // 8, 10)
+                self.add_integer_boundaries(self.max_num // 16, 10)
+                self.add_integer_boundaries(self.max_num // 32, 10)
+                self.add_integer_boundaries(self.max_num, 10)
+
+                power_of_two = math.log(self.max_num, 2)
+                bits = width if not self.max_num else int(math.floor(power_of_two))
```
jtpereyda: `power_of_two = math.log(self.max_num, 2)` will fail if  `self.max_num` is `None`.  Make this conditional situation more robust.

Also, for `bits = width if not self.max_num else int(math.floor(power_of_two))`, I'm not clear why the existence of max_num means you should use width.

Breaking this out into a proper if/then statement might make the whole thing look cleaner.

boofuzz/primitives/bit_field.py
```diff
@@ -106,34 +107,59 @@ def __init__(
...
+                    # check if width is a power of 2. if it isn't, calculate and add numbers near the next
+                    # lower power of two
+                    power_of_two = math.log(width, 2)
+                    if power_of_two != math.floor(power_of_two) and power_of_two > 0:
+                        tmp = 1 << int(power_of_two)
+                        self.add_integer_boundaries(tmp, 1)
+                        if signed:
+                            self.add_integer_boundaries(-tmp, 1)
+                    bits = bits // 2
+                    tmp = 1 << bits
```
jtpereyda: I suspect this whole chunk could be cleaned up a bit -- perhaps more helpful variable names or a modification of the layout. If you have a go I can make a second pass.

jtpereyda:
@secenv Thank you for the PR! I left some review comments for your consideration.

SR4ven:
Closing due to inactivity.
Feel free to reopen/comment!

Issue #326, opened 2019-08-30
pip install shouldn't use sudo


jeffythedragonslayer:
In the documentation, Installing from source on Ubuntu, "sudo pip install ." fails

Removing sudo works.

SR4ven:
What error message do you get?

`sudo` in combination with pip is a security risk anyways.
`pip install . --user` should be the safest way to install boofuzz for a single user.

I'll open a PR when I find some time.
Thanks for the info @jeffythedragonslayer.

jeffythedragonslayer:
Attached [sudo_pip_error.txt](https://github.com/jtpereyda/boofuzz/files/3563058/sudo_pip_error.txt)

sudo -H as it suggested also didn't work, and even more bizarre is that I already had setuptools version 41.2.0 so idk why it's complaining.

jtpereyda:
Has anyone replicated this error?

I'm a bit ambivalent on the best install strategy to recommend. Using virtualenv will work without sudo, but without virtualenv I would anticipate sudo wouldn't work, right?

SR4ven:
@jeffythedragonslayer your installation problem with `sudo` seems to be caused by connection problems as your log contains `Failed to establish a new connection: [Errno -2] Name or service not known`.

> 
> 
> Has anyone replicated this error?
> 
> I'm a bit ambivalent on the best install strategy to recommend. Using virtualenv will work without sudo, but without virtualenv I would anticipate install-without-sudo would fail, right?

@jtpereyda you are right, thanks for pointing this out.
The installation without `sudo` will fail because of missing write access to `/usr/local/lib/python3.7/dist-packages/`.
I'm currently on Ubuntu 19.04 which doesn't come with a reference to pip for the root user by default (`command not found`).
Installing with `sudo python3 -m pip install boofuzz` works fine, so the installation method is not broken, but not best practice either.

The one way to install pip packages, that has always worked for me, is using the `--user` option.
That installs the package for the current user and not system wide, which should be sufficient for the most use cases.
https://pip.pypa.io/en/stable/user_guide/#user-installs
https://stackoverflow.com/questions/42988977/what-is-the-purpose-pip-install-user/42989020

In my opinion we should recommend the user-install method and maybe mention the system wide installation only as a fallback.
On top, there are many statements about `sudo pip install` being a bad practice.
https://askubuntu.com/questions/802544/is-sudo-pip-install-still-a-broken-practice

[...]

PR #327, opened 2019-09-01
Remove sudo from install instructions

SR4ven:
Adressing #326 + some syntax errors.

Tested in a virtualenv under Ubuntu 19.04 where installing without `sudo` worked fine in every case.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ Fixes
 - Fixed process monitor handling of backslashes in Windows start commands
 - Fixed and documented `boo open`
 - Fixed receive function in `fuzz_logger_curses`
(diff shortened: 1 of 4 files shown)


jeffythedragonslayer:
Should it be pip or pip3?

SR4ven:
> Should it be pip or pip3?

Just `pip` should be fine for now, as it will always refer to the default python interpreter.
From Ubuntu 19.04 on that is Python 3.7 so `pip3` is not needed.

I now added `--user` to all install commands.
Another approach would be to only mention the `--user` option once in a note about permission errors.

jtpereyda:
Security win, I like it. Thanks @jeffythedragonslayer @SR4ven

PR #334, opened 2019-10-15
add crc16/dnp support and avoid CSV output decode error

JsHuang:


boofuzz/fuzz_logger_csv.py
```diff
@@ -56,7 +56,7 @@ def log_recv(self, data):
         self._print_log_msg(["recv", len(data), self._format_raw_bytes(data), data.decode()])
 
     def log_send(self, data):
-        self._print_log_msg(["send", len(data), self._format_raw_bytes(data), data.decode()])
+        self._print_log_msg(["send", len(data), self._format_raw_bytes(data), repr(data)])
```
jtpereyda: @JsHuang What is the impact of this change?

JsHuang: For none ascii data, data.decode() will throw exception

SR4ven: I agree that in case we send non ascii data, the logger will raise an uncaught exception. Git blame shows that the decoding was added when we were working on Python 3 compatibility.
I'd even go further and change the `log_recv()` function too, just in case the target answers with non ascii formatted data.

I did some testing to see what the difference between those two methods are.
`decode()` will interpret all input including line breaks. The result of this is, that the scv file will have line breaks inside the data field. This seems to be allowed as long as the field is surrounded by quotation marks, which it currently is.
But then we have the problem that quotation marks inside the data stream will mess with this so I don't think it's a good idea to decode the data.
On the pro side, decoding the data does of course improve the readability greatly, but `repr()` seems to be the better choice to guarantee compatibility.

jtpereyda: Thanks all; makes sense.

SR4ven:
@JsHuang I applied all necessary changes related to the CSV logger.

In the current state, the new checksum does not work with Python 3. This is because the CRC16_DNP class requires a string type as input, but we supply a bytes type value. For Python 2 that doesn't make a difference, but for Python 3 the decoding is missing.
At the moment I can't think of a way to reliably decode any kind of bytes to achieve an equivalent of how Python 2 handles this (where no encoding/decoding is happening).

I will open a new PR to get the CSV logger fix merged faster. If you continue working on this PR, please remove everything unrelated to the new checksum.

SR4ven:
Closing because of inactivity.

See #355, it is possible to pass a custom function to s_checksum.

Issue #296, opened 2019-06-28
OSError: [Errno 98] Address already in use


starek4:
I use `SocketConnection` with `tcp` protocol.

When I run a bunch of tests, sometimes (for me usually after between 15 000 - 30 000 test cases) I got the following error:

```
Traceback (most recent call last):
  File "fuzzer/src/wfuzz.py", line 43, in <module>
    main()
  File "fuzzer/src/wfuzz.py", line 39, in main
    fuzzer.fuzz()
  File ".../fuzzer/src/fuzzer.py", line 52, in fuzz
    self._session.fuzz()
  File ".../env/lib/python3.6/site-packages/boofuzz/sessions.py", line 590, in fuzz
    self._main_fuzz_loop(self._iterate_protocol())
  File ".../env/lib/python3.6/site-packages/boofuzz/sessions.py", line 695, in _main_fuzz_loop
    self._fuzz_current_case(*fuzz_args)
  File ".../env/lib/python3.6/site-packages/boofuzz/sessions.py", line 1412, in _fuzz_current_case
    self._open_connection_keep_trying(target)
  File ".../env/lib/python3.6/site-packages/boofuzz/sessions.py", line 1450, in _open_connection_keep_trying
    target.open()
  File ".../env/lib/python3.6/site-packages/boofuzz/sessions.py", line 79, in open
    self._target_connection.open()
  File ".../env/lib/python3.6/site-packages/boofuzz/socket_connection.py", line 159, in open
    self._sock.connect((self.host, self.port))
OSError: [Errno 98] Address already in use
```

It happens on both Windows and WSL and different versions of Python. No other application is connecting to the target port. Any ideas?

starek4:
When I have both server and fuzzer on the same machine, everything works fine.

starek4:
Well, after the weekend, I cannot reproduce this issue again. So I will believe that it was some environmental problem and hope that it never shows again. Closing this issue for now and when I see this behavior again, I will try to investigate it more and share with you guys what I've found.

jtpereyda:
I think I've seen that sort of thing before, but it was also transient for me. I remember something about the OS sometimes holding onto TCP connections. Could also be another program that happened to be trying to use the same port.

It also depends on your configuration -- if you're simply connecting to another TCP server, this would be rather unexpected.

starek4:
Well, it happens again 😢 

There is indeed a problem that TCP connections are not terminated. They are stuck in the state: **TimeWait**

[...]

PR #337, opened 2019-10-23
Enable "open_connection_keep_trying" for servers as well

ColBFritz:
I had an issue where I am fuzzing as a server, waiting for the target client to connect to the server. Sometimes the client hits a retry limit and gives up so the server is left waiting indefinitely for a connection to use for fuzzing.

I used recv_timeout and send_timeout on the SocketConnection used by the Target, which lets the server give up on that connection attempt. However, this raises an uncaught EAGAIN exception, killing the fuzzing session. Plus, I could not find a satisfactory built-in method to restart the target _only when needed_ for this connection attempt being abandoned.

So I found the open_connection_keep_trying feature that was added for clients (#268 ) and modified the connection code when running as a server to tie into the same feature. Now I can define a function to restart my target which gets called when the server times out waiting for a client to connect.

### DIFF
--- a/boofuzz/socket_connection.py
+++ b/boofuzz/socket_connection.py
@@ -167,7 +167,8 @@ class SocketConnection(itarget_connection.ITargetConnection):
                     self._serverSock.listen(1)
                     self._sock, addr = self._serverSock.accept()
                 except socket.error as e:
-                    # When connection timeout expires, tear down the server socket so we can re-open it again after restarting target
+                    # When connection timeout expires, tear down the server socket so we can re-open it again after
+                    # restarting the target.
                     self._serverSock.shutdown(socket.SHUT_RDWR)
                     self._serverSock.close()
                     if e.errno in [errno.EAGAIN]:
--- a/boofuzz/socket_connection.py
+++ b/boofuzz/socket_connection.py
@@ -175,8 +175,8 @@ class SocketConnection(itarget_connection.ITargetConnection):
                     self._sock, addr = self._serverSock.accept()
                 except socket.error as e:
                     # When connection timeout expires, tear down the server socket so we can re-open it again after
-                    # restarting the target. Can't just call self.close() because self._sock isn't as expected for it.
-                    self._serverSock.close()
+                    # restarting the target.
+                    self.close()
                     if e.errno in [errno.EAGAIN]:
                         raise exception.BoofuzzTargetConnectionFailedError(str(e))
                     else:
--- a/boofuzz/socket_connection.py
+++ b/boofuzz/socket_connection.py
@@ -163,8 +163,17 @@ class SocketConnection(itarget_connection.ITargetConnection):
 
             if self.proto == "tcp" or self.proto == "ssl":
                 self._serverSock = self._sock
-                self._serverSock.listen(1)
-                self._sock, addr = self._serverSock.accept()
+                try:
+                    self._serverSock.listen(1)
+                    self._sock, addr = self._serverSock.accept()
+                except socket.error as e:
+                    # When connection timeout expires, tear down the server socket so we can re-open it again after restarting target
+                    self._serverSock.shutdown(socket.SHUT_RDWR)
+                    self._serverSock.close()
+                    if e.errno in [errno.EAGAIN]:
+                        raise exception.BoofuzzTargetConnectionFailedError(str(e))
+                    else:
+                        raise
 
         # Connect is needed only for TCP protocols
         elif self.proto == "tcp" or self.proto == "ssl":
--- a/boofuzz/socket_connection.py
+++ b/boofuzz/socket_connection.py
@@ -159,7 +159,14 @@ class SocketConnection(itarget_connection.ITargetConnection):
 
         if self.server:
             self._sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-            self._sock.bind((self.host, self.port))
+            try:
+                self._sock.bind((self.host, self.port))
+            except socket.error as e:
+                # Handle when socket is in TIME_WAIT state.
+                if e.errno == errno.EADDRINUSE:
+                    raise exception.BoofuzzOutOfAvailableSockets()
+                else:
+                    raise
 
             if self.proto == "tcp" or self.proto == "ssl":
                 self._serverSock = self._sock
@@ -168,7 +175,7 @@ class SocketConnection(itarget_connection.ITargetConnection):
                     self._sock, addr = self._serverSock.accept()
                 except socket.error as e:
                     # When connection timeout expires, tear down the server socket so we can re-open it again after
-                    # restarting the target. Can't call self.close() because we don't have a valid self._sock.
+                    # restarting the target. Can't just call self.close() because self._sock isn't as expected for it.
                     self._serverSock.close()
                     if e.errno in [errno.EAGAIN]:
                         raise exception.BoofuzzTargetConnectionFailedError(str(e))
--- a/boofuzz/socket_connection.py
+++ b/boofuzz/socket_connection.py
@@ -168,8 +168,7 @@ class SocketConnection(itarget_connection.ITargetConnection):
                     self._sock, addr = self._serverSock.accept()
                 except socket.error as e:
                     # When connection timeout expires, tear down the server socket so we can re-open it again after
-                    # restarting the target.
-                    self._serverSock.shutdown(socket.SHUT_RDWR)
+                    # restarting the target. Can't call self.close() because we don't have a valid self._sock.
                     self._serverSock.close()
                     if e.errno in [errno.EAGAIN]:
                         raise exception.BoofuzzTargetConnectionFailedError(str(e))


boofuzz/socket_connection.py
```diff
@@ -160,8 +160,18 @@ def open(self):
...
                 self._serverSock = self._sock
-                self._serverSock.listen(1)
-                self._sock, addr = self._serverSock.accept()
+                try:
+                    self._serverSock.listen(1)
+                    self._sock, addr = self._serverSock.accept()
+                except socket.error as e:
+                    # When connection timeout expires, tear down the server socket so we can re-open it again after
+                    # restarting the target.
+                    self._serverSock.shutdown(socket.SHUT_RDWR)
```
SR4ven: Why are we using `shutdown()` here but nowhere else?
https://github.com/jtpereyda/boofuzz/blob/e3c79dc0257f10ee4bfaa7273b87d702d3a9d030/boofuzz/socket_connection.py#L116-L125
Should we use it everywhere or nowhere?

According to https://docs.python.org/3/library/socket.html#socket.socket.close using `shutdown()` would close the connection immediately, while `close()` only decrements the handle count and closes the socket when no handles are left. See https://stackoverflow.com/questions/409783/socket-shutdown-vs-socket-close

This might be related to https://github.com/jtpereyda/boofuzz/issues/296

ColBFritz: I can't comment on where else shutdown() might be needed, but I added it here specifically to get past errors for the port already being in use. I don't remember the specific error raised but I think it might have been the "address already in use" message from #296 which you mention. In general, if you're done with a connection for one test, and are closing it or letting it get garbage collected, but the next test trys to establish a connection on the same port, it probably is a good idea to call shutdown().

ColBFritz: Just checking in case I misunderstood something: is there something you wanted changed in this pull request before it's ready?

SR4ven: It's fine by me, I'm just unsure whether it would be a benefit to add the socket shutdown to the default close function. If we want to make sure the connection is closed immediately, the shutdown method might help.

> **Note:** close() releases the resource associated with a connection but does not necessarily close the connection immediately. If you want to close the connection in a timely fashion, call shutdown() before close().

From https://docs.python.org/3/library/socket.html#socket.socket.close

Either way it's up to @jtpereyda to merge this.

jtpereyda: I didn't see anything to the effect that shutdown is safer in my usual sockets reference: http://beej.us/guide/bgnet/html/#close-and-shutdownget-outta-my-face

And I don't think anyone's encountered this error before with boofuzz (server fuzzing is a newer use case though). @ColBFritz I wonder if your error might have been related to some of the recent sockets related fixes. Would you be willing to switch it back to close (after merging in updates from master)? If the same errors happen again, please share them here so we can check it out. Thanks!

SR4ven: I did some more research regarding the shutdown function and from what I read, it is considered best practice to use it. Though, it is almost never really required.
See https://docs.python.org/2/howto/sockets.html#disconnecting

As I understood it, `shutdown()` immediately sends a FIN but it does not free/close the socket.
Calling `close()` without a previous shutdown might not send the FIN right away but instead will try to finish sending/receiving whatever is left before closing the socket. As I mentioned before, the handles count is decremented with `close()` so that the kernel can close the socket when no handles are left.

So as long as we don't call `close()` while another thread is still sending/receiving data it shouldn't make a difference.

@ColBFritz keep in mind that calling `shutdown()` will fail with ENOTCONN if the client has already closed the connection. In that case we should catch the exception and call `close()`.

jtpereyda: I found that FIN vs RST thing (second answer [here](https://stackoverflow.com/questions/4160347/close-vs-shutdown-socket)). Good point about the ENOTCONN issue.

If no one objects and it doesn't cause any problems, it seems like switching back to close for now makes sense. If that causes errors, we can reassess.

Maybe it would make sense to switch everything to shutdown then close, but this would be easiest as a separate change.

ColBFritz: @jtpereyda thanks for the feedback, I will give it another try without shutdown() in a few days. Are "the recent sockets related fixes" you refer to #342 ? I believe I had been getting EADDRINUSE as that pull request handles so it will probably work. However, I see that it adds a minimum 5 second delay between test cases if that error is raised, whereas with shutdown it can proceed without any delay. That could greatly increase the amount of time taken to fuzz, given thousands of mutations; although in this case, it only is an issue when the target is already being restarted which _should_ be rare, and can take several seconds anyway. Still, it seems strange to just expect an error and catch it to wait for the system to clean up the socket, rather than avoiding it by explicitly shutting down the socket.

@SR4ven good catch for ENOTCONN. If the performance is acceptable without calling shutdown as jtpereyda suggests, then no specific handling is needed for that, correct?

starek4: @ColBFritz calling shutdown before closing socket did not solve the problem with EADDRINUSE for me. If your EADDRINUSE errors are solved by calling shutdown, we can call it before every close.

SR4ven: @ColBFritz you're right about the error handling. If we don't use shutdown, we don't need to care about ENOTCONN.

@starek4 I found pretty good definition of your problem [here](https://knowledgebase.progress.com/articles/Article/Can-the-time-a-socket-spends-in-TIMED-WAIT-state-be-reduced).
In that case there is no other option to speed up the fuzzing, as the kernel will take care of the closed sockets and we cannot influence that process.
According to the article, it doesn't make a difference if the socket was shutdown or closed.

@ColBFritz in the initial comment on this PR you were talking about an uncaught EAGAIN exception, while #342 fixed EADDRINUSE. Please correct me if I'm wrong but I don't think those are related. So I recommend testing this PR on top of the current master without shutdown. Please report back with the results.

We can still add shutdown to our socket closing function if needed, but we should do that consistently and in a separate PR.

ColBFritz: Thanks, and sorry for the delay getting back.

Rebasing onto the latest master, and removing shutdown(), I got no errors in over an hour of running, with several restarts triggered during that time.

If I also remove the close(), I get "socket.error: [Errno 98] Address already in use". I had assumed it would hit that retry code instead of throwing an uncaught exception, but it looks like #342 was only for client code, not server code.

For comparison, I ran my original code but with the shutdown() call removed, for over a half hour. I see several restarts, but still no socket errors. I worry that I might hit a corner case where close() isn't enough if I leave it running for several hours, because I'm pretty sure I saw this error before when I did not have a shutdown() call. However from the links in this discussion, even calling shutdown() may not be enough, and we could run into an issue with TIME_WAIT similar to what was fixed in #342, but on the server side.

Rather than calling shutdown(), I am adding code similar to #342 around the bind() call. I think it will work and it's probably necessary but I don't know how to force the situation where I can test it.

SR4ven: > If I also remove the close(), I get "socket.error: [Errno 98] Address already in use". I had assumed it would hit that retry code instead of throwing an uncaught exception, but it looks like #342 was only for client code, not server code.

#342 was indeed only for client side connections. Currently there is no exception handling for opening server side connections.

> Rather than calling shutdown(), I am adding code similar to #342 around the bind() call. I think it will work and it's probably necessary but I don't know how to force the situation where I can test it.

Sounds like the best option. Catching `EADDRINUSE` as in the following example should do it I guess.
```python
try:
    bind...
except socket.error as e:
    if e.errno == errno.EADDRINUSE:
        raise exception.BoofuzzOutOfAvailableSockets()
```
To test this you could try to increase your `TIME_WAIT` value so that the kernel closes connections the less frequent. Combined with a high test case frequency you might be able to provoke the Address already on use error.

boofuzz/socket_connection.py
```diff
@@ -159,12 +159,28 @@ def open(self):
...
             if self.proto == "tcp" or self.proto == "ssl":
                 self._serverSock = self._sock
-                self._serverSock.listen(1)
-                self._sock, addr = self._serverSock.accept()
+                try:
+                    self._serverSock.listen(1)
+                    self._sock, addr = self._serverSock.accept()
+                except socket.error as e:
+                    # When connection timeout expires, tear down the server socket so we can re-open it again after
+                    # restarting the target. Can't just call self.close() because self._sock isn't as expected for it.
```
SR4ven: Without having tested it, I currently don't see why we can't use `self.close()`. Does `self._sock.close()` cause an exception?

boofuzz/socket_connection.py
```diff
@@ -159,12 +159,28 @@ def open(self):
...
+                    self._serverSock.close()
+                    if e.errno in [errno.EAGAIN]:
```
starek4: When the process cannot allocate a new socket (for example because of `TIME_WAIT` problem), the error number in the socket is `errno.EADDRINUSE`. Or is this behaviour different for `bind` method? If not, I would suggest handling `errno.EADDRINUSE` too.

Like:
```python
if e.errno in [errno.EAGAIN, errno.EADDRINUSE]:
    raise exception.BoofuzzTargetConnectionFailedError(str(e))
```

Or maybe I just missing something. Feel free to correct me, I am quite a beginner in this field. Thanks.

SR4ven:
@starek4 did you do some testing for the additional error handling that you suggested as you closed your review?

Reading the docs about [socket.accept()](https://docs.python.org/3/library/socket.html#socket.socket.accept), it could make sense to catch `EADDRINUSE` as accept() appears to open a new socket object.
This seems to be rather difficult to test, but I guess it's better to just catch the exception than not to.

starek4:
@SR4ven no, I did not test it. I misunderstood the latest changes, that is why I deleted the last comment. Sorry about that.

Well, I thought that only `socket.bind` could cause `EADDRINUSE`, and this is already handled by @ColBFritz . I am not sure if `socket.accept` could trigger it too, especially when we are using `SO_REUSEADDR` option. But this additional check will not cause any troubles, so we can add it there, just to be sure to handle all situations 👍

SR4ven:
Thanks for the info @starek4 

I took a look at some other open source server socket applications and couldn't find a single one that catches `EADDRINUSE` for `socket.accept`. So I tend to follow these examples and just leave it as it is. We can always add that error handling when this specific error comes up.

boofuzz/socket_connection.py
ColBFritz: So I tested and actually self.close() works fine, so I changed it to that. I had assumed that since we are handling an error in listen() or accept() that self._sock would still be the same object as self._serversock, and further assumed the double close on that object would cause issues.

I guess it works fine so to help keep things consistent in the future using self.close() is a better idea, to avoid having close code in multiple places.

boofuzz/socket_connection.py
ColBFritz: I was unable to trigger the error handling code with my specific fuzzer I've been using to debug this, even after setting tcp_fin_timeout (mentioned in [the article you linked earlier](https://knowledgebase.progress.com/articles/Article/Can-the-time-a-socket-spends-in-TIMED-WAIT-state-be-reduced)) to several minutes, although I briefly saw my connection in TIME_WAIT state a few times in the netstat -a output.  I don't want to monkey with that setting too much, as this test runs on a shared resource, and this is a global setting. But, the restart code still seems to work fine, worst case with the new exception handler is that we catch an exception for a different problem and retry for a few minutes before informing the user of a connection error.

SR4ven:
Looks good to me now.
Thanks for the work on this PR @ColBFritz!

PR #338, opened 2019-10-23
Allow running with older/missing dependencies

ColBFritz:
I am running my fuzzers on an older system which is difficult to install new packages on, and could not run the latest boofuzz code until I made the following updates:

 - Ignore import failures for fuzz_logger_curses, in case its dependencies are not available, if the console logger is not used.
 - Fall back to compatible version of attr, if ib() does not take a type argument.

### DIFF
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -134,11 +134,13 @@ __all__ = [
 ]
 
 if exclude_curses_logger:
-    __all__.remove('FuzzLoggerCurses')
+    __all__.remove("FuzzLoggerCurses")
 
-__version__ = '0.1.5'
+__version__ = "0.1.5"
 
 # REQUEST MANAGEMENT
+
+
 def s_get(name=None):
     """
     Return the request with the specified name or the current request if name is not specified. Use this to switch from
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -12,12 +12,7 @@ from .event_hook import EventHook
 from .exception import MustImplementException, SizerNotUtilizedError, SullyRuntimeError
 from .fuzz_logger import FuzzLogger
 from .fuzz_logger_csv import FuzzLoggerCsv
-exclude_curses_logger = False
-try:
-    from .fuzz_logger_curses import FuzzLoggerCurses
-except ImportError:
-    exclude_curses_logger = True
-    pass  # allow continuing if fuzz_logger_curses is never used
+from .fuzz_logger_curses import FuzzLoggerCurses
 from .fuzz_logger_text import FuzzLoggerText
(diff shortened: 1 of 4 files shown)


SR4ven:
Thanks for the PR and making us aware of these issues.

What exact import errors did you get concerning fuzz_logger_curses? We already have an import guard for curses, so it must be something different right?
https://github.com/jtpereyda/boofuzz/blob/e3c79dc0257f10ee4bfaa7273b87d702d3a9d030/boofuzz/fuzz_logger_curses.py#L6-L9
I'm asking because I think we should fix this in the fuzz_logger_curses class itself if possible. That would keep the importing files shorter/cleaner.

ColBFritz:
Sorry for the force-push. I didn't notice my committer and author fields differed.

The issue I was having was not curses, but "backports-shutil-get-terminal-size". It's not available on the system I'm running on and getting any new packages added to the repository is a time-consuming process.

SR4ven:
Wouldn't it be easier to add an import guard around the backports then?

The reason we have an import guard around curses is that we cannot install it with pip. All other packages are installed automatically, so there shouldn't be any problems when installing boofuzz with pip.
Adding the mentioned include guard shouldn't be an issue, though your problem seems to be quite unique.

ColBFritz:
How about this instead, then? I reverted the fuzz_logger_curses import guards, and put the import guard where you suggested right in fuzz_logger_curses itself. Instead of just failing I added a fallback that just assigns a static window size, with a warning raised to the user about the missing dependency. I used UserWarning instead of ImportWarning because the default warning filter ignores ImportWarning so nobody would see a message with that type.

ColBFritz:
FYI: Testing on my system, the console GUI does seem to work with these changes, although it gets corrupted from time to time and won't exit with CTRL+C. But I think those issues are unrelated to my changes and just the result of the very specific package versions installed on this system and the way I'm accessing it (through a Windows implementation of X Windows, connecting to xterm, running tmux, and fuzzing inside all of that...too much to bother debugging for a feature I don't really plan to use very often).

ColBFritz:
@SR4ven I tweaked the warning messages as suggested, except instead of "may not resize" I used "will not resize" because no effort is made to do so. I initialized curses to None as suggested to force failures on use, but did not make the warning OS dependent. While curses _ought to_ be available on any *nix system, a failure to import curses on _any_ system (perhaps they have a very minimal custom install or something) should show the warning so the user is not confused why an internal BooFuzz feature is failing.

jtpereyda:
Thank you for the PR @ColBFritz , and for the review @SR4ven !

On the check for os = nt vs. catch the import exception issue, I agree with y'all's conclusion, and would note that this is a classic instance where the Python EAFP philosophy applies well. Cheers!

PR #340, opened 2019-10-30
Fix setsockopt timeout on Windows

### LINKED ISSUE #281: Fuzzing HTTP server on Windows - TimeoutError: [WinError 10060]

SR4ven:
In #300 it was noticed, that `setsockopt()` does not work correctly on Windows.
The reason for this is, that Windows uses a different format for the timeout values than Linux.
Windows uses a DWORD in milliseconds, while Linux uses a byte-like-object of seconds and microseconds.
See https://docs.microsoft.com/en-us/windows/win32/winsock/sol-socket-socket-options

Might fix #281, too? @starek4

On top I applied some fixes for possible reference-before-assignment warnings.

### DIFF
--- a/boofuzz/socket_connection.py
+++ b/boofuzz/socket_connection.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 
 import errno
 import math
+import os
 import socket
 import ssl
 import struct
@@ -14,12 +15,17 @@ from . import exception, helpers, ip_constants, itarget_connection
 ETH_P_IP = 0x0800  # Ethernet protocol: Internet Protocol packet, see Linux if_ether.h docs for more details.
 
 
-def _seconds_to_second_microsecond_struct(seconds):
-    """Convert floating point seconds value to second/useconds struct used by socket library."""
-    microseconds_per_second = 1000000
-    whole_seconds = int(math.floor(seconds))
-    whole_microseconds = int(math.floor((seconds % 1) * microseconds_per_second))
-    return struct.pack("ll", whole_seconds, whole_microseconds)
+def _seconds_to_sockopt_format(seconds):
+    """Convert floating point seconds value to second/useconds struct used by UNIX socket library.
+    For Windows, convert to whole milliseconds.
+    """
+    if os.name == "nt":
+        return int(seconds * 1000)
+    else:
+        microseconds_per_second = 1000000
+        whole_seconds = int(math.floor(seconds))
+        whole_microseconds = int(math.floor((seconds % 1) * microseconds_per_second))
+        return struct.pack("ll", whole_seconds, whole_microseconds)
 
 
 class SocketConnection(itarget_connection.ITargetConnection):
@@ -104,6 +110,7 @@ class SocketConnection(itarget_connection.ITargetConnection):
             if self.sslcontext is None and self.server_hostname is None:
                 raise ValueError("SSL/TLS requires either sslcontext or server_hostname to be set.")
 
+        self._serverSock = None
         self._sock = None
         self._udp_client_port = None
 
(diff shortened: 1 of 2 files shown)


starek4:
Great work, @SR4ven!

I tried an example from https://github.com/jtpereyda/boofuzz/issues/281 on Boofuzz before this patch and from time to time it failed. Now, after this patch, I am unable to get the error even after many repeats of this test suite. So it seems that it solved the issue. I can finally get rid of the ugly workaround, which I have implemented in [wapifuzz](https://github.com/ysoftdevs/wapifuzz/blob/master/fuzzer/src/post_test_case_callback.py#L22).

I also test SSL fuzzing and it also solves the issue, which I describe here https://github.com/jtpereyda/boofuzz/pull/300#issuecomment-548114002.

starek4:
Just for the record, it does **not** solve https://github.com/jtpereyda/boofuzz/issues/296.

boofuzz/socket_connection.py
```diff
@@ -241,6 +246,8 @@ def send(self, data):
         Returns:
             int: Number of bytes actually sent.
         """
+        num_sent = 0
```
jtpereyda: Just curious, were these changes (this and `data = b""` and `self._serverSock = None`) requested by Stickler or some other tool?

jtpereyda:
For those that have tested, did you test using 64 bit Python on Windows? I'm not clear on what `setsockopt` when given an int for the third argument. I'm just a little nervous it would throw in a 64 bit value instead of 32 bit. The docs aren't explicit on what happens when you give an int https://docs.python.org/2.7/library/socket.html

jtpereyda:
Btw, this is very cool. This is the kind of infrastructure type work that makes tools useful for more and more people.

boofuzz/socket_connection.py
SR4ven: Those were warnings in PyCharm.
Looking closer at the functions, all cases seem to be covered so that a reference before assignment should be impossible. PyCharm might not be smart enough to see this, but I guess it doesn't hurt to just initialise the return value. Just to be on the safe side and to appease PyCharm :D.

About `_serverSock`, it was initialised outside of the init function, with PyCharm complaining about this too.

SR4ven:
I was testing with Python 3.7 64bit on Windows 10 64 bit and it worked perfectly fine.

I didn't test with large values which would only fit into a 64bit Int.

starek4:
I was testing on Windows 10 64-bit with Python 3.7.2 32-bit.

jtpereyda:
Well, the int thing makes me nervous but that's just speculation. In any case this is a solid improvement. Thanks @SR4ven and @starek4 for the code and review!

jtpereyda:
Great, good to know it works with both 32 and 64 bit Python. I'm glad to have this shortcoming filled in. Thanks @SR4ven and @starek4 for the code and review!

@SR4ven Could you reach out to me at my email (viewable on my profile page https://github.com/jtpereyda)? Github doesn't allow messages or anything so sorry for the in-PR shout out.

PR #341, opened 2019-11-01
Add Python 3.5/3.8 CI-checks

SR4ven:
As Python 3.8 is now released as stable we should run our tests against it.
At the same time I fixed some upcoming deprecation warnings with `thread.isAlive()`.

With the [python_requires ](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires)option, pip will refuse to install boofuzz on any unsupported Python version. We can't support versions <3.5 because there will be no [Flask ](https://pypi.org/project/Flask/)version for it.

I noticed that some package classifiers for pip are outdated, so I updated those too. I narrowed the supported Python versions down to the major versions, or should we list the minor versions too?
According to https://github.com/jtpereyda/boofuzz/blob/master/LICENSE.txt we are using GPLv2 and not MIT.


jtpereyda:
LGTM

PR #342, opened 2019-11-01
Fix for #296, socket exhaustion

starek4:
See https://github.com/jtpereyda/boofuzz/issues/296#issuecomment-548774369 for more details.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,7 +23,7 @@ Fixes
 - Fixed receive function in `fuzz_logger_curses`
 - Installing boofuzz with `sudo` is no longer recommended, use the `--user` option of pip instead
 - Fixed setting socket timeout options on Windows
-- If all sockets are exhausted, repeatedly try fuzzing for 2 minutes before failing
+- If all sockets are exhausted, repeatedly try fuzzing for 4 minutes before failing
 
 v0.1.5
 ------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,7 @@ Fixes
 - Fixed receive function in `fuzz_logger_curses`
 - Installing boofuzz with `sudo` is no longer recommended, use the `--user` option of pip instead
 - Fixed setting socket timeout options on Windows
+- If all sockets are exhausted, repeatedly try fuzzing for 2 minutes before failing
 
 v0.1.5
 ------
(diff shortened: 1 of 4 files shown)


starek4:
I just refactor the code a bit. Ready for verification.

jtpereyda:
Seems like a good workaround overall, given that a fix that avoids blocking would take some more time and effort.

One suggestion on the code: You might give a wait time just a bit over 2 minutes in order to avoid throwing an exception in some corner case where the timing is off.

Did you check Linux MSL time?

starek4:
MSL for Linux is usually lower than for Windows. [Wiki ](https://en.wikipedia.org/wiki/Maximum_segment_lifetime)says that usually 60s (30s for FreeBSD).

I personally tried these distributions:
- Ubuntu 19.04, desktop version: 60s
- Fedora 28, desktop version: 60s
- Debian GNU/Linux 9.9 (stretch), minimal server installation: 60s

I do not expect that any standard Linux would have this value greater than is recommended by RFC. So I think that something a bit more than 2 minutes should work great.

SR4ven:
The values you checked seems to be the TIME_WAIT value, which is 2 * MSL according to https://stackoverflow.com/a/35000966 and https://en.wikipedia.org/wiki/Maximum_segment_lifetime

I checked with `cat /proc/sys/net/ipv4/tcp_fin_timeout` on Ubuntu 19.04 which returned 60s. So MSL would be 30s.

In the code you are retrying 50 times, waiting 5s each time. That's 250 seconds or about 4.2 minutes. Am I missing something?

starek4:
Oh, sorry. Yes, these values on Linux are `TIME_WAIT`, not MSL. My bad.

But on Windows (Windows 2000 server), the `TIME_WAIT` is 4 minutes (240s).
Src: https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-2000-server/cc938217(v=technet.10)?redirectedfrom=MSDN

On newer Windows, `TcpTimedWaitDelay` can be a bit lover, according to [this ](https://docs.microsoft.com/en-us/biztalk/technical-guides/settings-that-can-be-modified-to-improve-network-performance) article.

But, the maximum `TIME_WAIT` which I found is 4 minutes. I add a more 10 seconds because of corner cases, which @jtpereyda mentions.

starek4:
And I just want to add, that for my environment, usually 3 iterations (so ~15s) is enough for an additional ~15k test cases. This repeats every ~15k test cases.

SR4ven:
Thanks for the clarification and the work on this issue @starek4!

PR #345, opened 2019-11-06
added error handling to byte decoding in csv logger

dorpvom:
This fixes issues arising from *non-decodable* bytes in the fuzzing payload. While the `_format_raw_bytes` function can be injected from outside, it does not seem that the second decode can be skipped.
A regression test for this would be simple if wanted. I'd just have to first acquaint myself with the test syntax and tooling.

### DIFF
--- a/boofuzz/fuzz_logger_csv.py
+++ b/boofuzz/fuzz_logger_csv.py
@@ -53,10 +53,10 @@ class FuzzLoggerCsv(ifuzz_logger_backend.IFuzzLoggerBackend):
         self._print_log_msg(["error", "", "", description])
 
     def log_recv(self, data):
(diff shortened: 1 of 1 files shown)


SR4ven:
Thanks for the PR @dorpvom!

This problem was addressed in #334, but isn't merged yet because of some pending changes.
The author solved the issue by simply replacing `data.decode()` with `repr(data)`. Take a look at the discussion here https://github.com/jtpereyda/boofuzz/pull/334#discussion_r341819629.

Or is there any reason to stick with `decode()` that we missed?

dorpvom:
Thanks, that looks find. I'd also say repr is sufficient. I'll close this one.

PR #346, opened 2019-11-08
Update documentation links and add sphinx linkcheck to tox

SR4ven:
Some links in the documentation were outdated, so I updated them and added a link checker to tox.
This should give us a warning in case of broken links in the future.

I shorted the link to the gitter page by the stats/reference arguments. If those were included for a reason, I'll add them back.


jtpereyda:
Very cool; I like the link checker.

PR #347, opened 2019-11-08
Improve appearance of docs

SR4ven:
Suggestion for a different documentation theme.

[Preview](https://boofuzz.readthedocs.io/en/sphinx_rtd_theme/)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -82,7 +82,7 @@ todo_include_todos = False
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = "alabaster"
+html_theme = "sphinx_rtd_theme"
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,6 +19,7 @@
 # import os
 # import sys
 # sys.path.insert(0, os.path.abspath('.'))
+from datetime import datetime
 
 
 # -- General configuration ------------------------------------------------
@@ -46,7 +47,7 @@ master_doc = "index"
 
 # General information about the project.
 project = u"boofuzz"
-copyright = u"2017, Joshua Pereyda"
+copyright = u"{}, Joshua Pereyda".format(datetime.now().year)
 author = u"Joshua Pereyda"
 
 # The version info for the project you're documenting, acts as replacement for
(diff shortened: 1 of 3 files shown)


jtpereyda:
IDK, I kinda like the other theme better as I feel the sphinx_rtd_theme is a bit overused. I first noticed the Alabaster theme while reading the [Requests](https://requests.readthedocs.io/en/master/) docs and have liked it ever since. I like the heading font in Alabaster (thinner letters that look less like a **Bold** font) and the minimalist left-hand side bar. I also like the default code theme -- gray background instead of yellow-green; Consolas font instead of the Sphinx theme's Courier-esque look. Definitely open to more discussion or thoughts on this though!

SR4ven:
You've got point there, the Requests doc looks very sleek and minimalistic. The difference to our docs is, that they have lots of custom, handwritten text and code examples, which leads to said appearance.

Our docs on the other hand mostly contain content generated by autodoc, which IMHO look a bit cluttered in Alabaster. E.g. parameter and return definition font is bigger than the actual function/class name. [Example](https://boofuzz.readthedocs.io/en/latest/user/connections.html#rawl3socketconnection)

You're also right with sphinx_rtd_theme being overused, so I looked around for some alternatives [here](https://sphinx-themes.org/). The only one that I found appealing was [Insegel](https://insegel.readthedocs.io/en/latest/). Thoughts?

However, there's no need to switch themes, if we don't find a better alternative.

jtpereyda:
I like Insegel more than the sphinx_rtd_theme. Do you think it addresses some of your other concerns?

I see what you mean about the function/class definition looking weird in Alabaster. I think sphinx_rtd_theme does a better job there. Not sure about Insegel.

SR4ven:
I did some testing with Insegel and ran into a few bugs. Headings with preceding labels are displayed twice and long code examples completely break the website. I opened issues about this on the Insegel repro, but until this is fixed we should stay with a different theme.

I still think that sphinx_rtd has the best overall readability, but I also agree that it's vastly overused.
While looking for alternatives, I noticed some settings for Alabaster that might help us to improve the appearance.
For my taste the autodoc content is too cluttered with not enough contrast for e.g. class names (we could introduce headings for each class to fix this). Anyways, I found a way to increase the page width, which makes the theme much more appealing to me. Maybe this is the best way to go right now.

[Preview](https://boofuzz.readthedocs.io/en/sphinx_rtd_theme/)
At the moment the page is 1200px wide, I think most devices support that.

jtpereyda:
@SR4ven just visited the requets docs again -- what we need is a sweet logo :P

Issue #348, opened 2019-11-15
SSLError exception not properly handled


kmendoza-bt:
As of python 3.3 ssl.SSLError exception is not a subtype of socket.error anymore. If the target connection is using SSL, and the ssl module throwns an exception the fuzzer is terminated.

PR #349, opened 2019-11-16
Add handling of ssl.SSLError exception.

kmendoza-bt:
Proposed fix for #348. I didn't want to add another exception to boofuzz; hence, why I chose to throw an exception.BoofuzzTargetConnectionFailedError exception when socket_connect.send or socket_connect.recv encounters an ssl.SSLError exception.

One case where I've seen this issue crop up is when the system I was fuzzing encountered a stack exhaustion exception. The remote end triggered an SSL bad length error. A way that I feel this could be replicated is to write a program that uses OpenSSL BIO and have a filter that would deliberately "chop" the data that the SSL BIO source wants to have sent to the TCP connection.


SR4ven:
Thank you for the PR @kmendoza-bt!

The exception handling for SSL errors looks reasonable, but there are a few things we should discuss.
1. While we're at it, wouldn't it make sense to add the SSL error handling to `socket_connection.open()` too? I don't have a lot of experience with SSL sockets in Python, so I can't tell if this is really needed. Either way, I guess it won't hurt having some additional exception handling.
2. There might be a rare case, when using the post_send callback function, where the log message could be misleading.
https://github.com/jtpereyda/boofuzz/blob/74dd08c783c2301081751fe793c94d8b36c62730/boofuzz/sessions.py#L1526-L1527
If an SSL error comes up when calling `send()` or `recv()` on the target inside a post_send callback, it will be caught here and the fail log message will be overwritten with a constant. In this case the SSL error message won't be displayed.
Fixing this would require a new exception or a different error handling in the pre/post/restart callback functions. As I don't expect this problem to appear often, I'd be fine with postponing it for another time.

kmendoza-bt:
@SR4ven,
Thank you for reviewing my PR.

> 1. While we're at it, wouldn't it make sense to add the SSL error handling to `socket_connection.open()` too? I don't have a lot of experience with SSL sockets in Python, so I can't tell if this is really needed. Either way, I guess it won't hurt having some additional exception handling.

Looking at the documentation for [ssl.SSLContext.wrap_socket()](https://docs.python.org/3.7/library/ssl.html#ssl.SSLContext.wrap_socket) I would say it's best to catch the SSLError exception; because it appears that the handshake will immediately happen since by the time it's called in socket_connection.open() the socket.connect() has already taken place.

> 2. There might be a rare case, when using the post_send callback function, where the log message could be misleading.
>    https://github.com/jtpereyda/boofuzz/blob/74dd08c783c2301081751fe793c94d8b36c62730/boofuzz/sessions.py#L1526-L1527
>    
>    If an SSL error comes up when calling `send()` or `recv()` on the target inside a post_send callback, it will be caught here and the fail log message will be overwritten with a constant. In this case the SSL error message won't be displayed.
>    Fixing this would require a new exception or a different error handling in the pre/post/restart callback functions. As I don't expect this problem to appear often, I'd be fine with postponing it for another time.

Given this information, I think it's best to have a separate exception for SSL-related issues. Should we have 1 exception for all SSL-related issues--like what I did with the PR; or should we have exceptions to distinguish whether the exception occurred during connect/reconnect or during send/receive? I would also need to be pointed where else in the code base should I look to determine if the new exception should be caught there or allowed to percolate up the call stack.

I'll make the necessary changes once I get your response, and as time permits.

SR4ven:
> Looking at the documentation for ssl.SSLContext.wrap_socket() I would say it's best to catch the SSLError exception; because it appears that the handshake will immediately happen since by the time it's called in socket_connection.open() the socket.connect() has already taken place.

Sounds good to me.

> Given this information, I think it's best to have a separate exception for SSL-related issues. Should we have 1 exception for all SSL-related issues--like what I did with the PR; or should we have exceptions to distinguish whether the exception occurred during connect/reconnect or during send/receive? I would also need to be pointed where else in the code base should I look to determine if the new exception should be caught there or allowed to percolate up the call stack.

I took a closer look at the error handling and I now think that we should both reuse `BoofuzzTargetConnectionFailedError` and add a new exception for SSL related issues.
`BoofuzzTargetConnectionFailedError` should be used for any exceptions while opening connections to the target, no matter if SSL related or not. To add more details, we should add more details about the root of the exception when raising it, so that we can log it when it's being caught. For that we could either use predefined strings from `constants.py` or a stack trace of the exception.

Example for raising
```python
try:
except socket.error as e:
    if e.errno in [errno.ECONNREFUSED, errno.EINPROGRESS, errno.ETIMEDOUT]:
        raise exception.BoofuzzTargetConnectionFailedError(str(e))
```
Example for catching
```python
    except exception.BoofuzzTargetConnectionFailedError as e:
        self._fuzz_data_logger.log_fail(str(e))
```
The problem is, that there is no proper handling for `BoofuzzTargetConnectionFailedError` yet. Only the `_post_send` method catches it at the moment. On top, the whole exception handling for the callback functions in not consistent. I'm not exactly sure how to deal with that yet, but for now it might be sufficient to just adopt the one from `_post_send` for `_pre_send` and `_restart_target`.

For SSL related errors we should add a new exception `BoofuzzTargetSSLError` (name is not fixed), that is used for SSL errors during sending/receiving. This exception needs to be caught at the same points as `BoofuzzTargetConnectionReset`. The affiliated checks for `_ignore_connection_issues_when_sending_fuzz_data` and `_check_data_received_each_request` to decide whether or not to mark the test as failed should be adopted.

You don't need to take care of this all, the SSL related part would be fine for this PR.
If you need any help/details let us know.

kmendoza-bt:
@SR4ven I updated my PR to incorporate the changes you recommended above. Let me know if there are more changes we need to incorporate.

boofuzz/sessions.py
```diff
@@ -1162,6 +1171,10 @@ def transmit_fuzz(self, sock, node, edge, callback_data):
             else:
                 self._fuzz_data_logger.log_info(msg)
             pass
+        except exception.BoofuzzTargetConnectionFailedError as e:
+            self._fuzz_data_logger.log_fail(str(e))
+        except exception.BoofuzzSSLError as e:
+            self._fuzz_data_logger.log_fail(str(e))
```
starek4: ## Just minor stylish issue
When the block in exception handling is the same for more types of exception, you can catch multiple of them in one line.

In this example:
```python
except (exception.BoofuzzTargetConnectionFailedError, exception.BoofuzzSSLError ) as e:
    self._fuzz_data_logger.log_fail(str(e))
```

Src: https://docs.python.org/3/tutorial/errors.html#handling-exceptions

boofuzz/sessions.py
```diff
@@ -1117,6 +1120,10 @@ def transmit_normal(self, sock, node, edge, callback_data):
                 self._fuzz_data_logger.log_fail(msg)
             else:
                 self._fuzz_data_logger.log_info(msg)
+        except exception.BoofuzzTargetConnectionFailedError as e:
+            self._fuzz_data_logger.log_fail(str(e))
+        except exception.BoofuzzSSLError as e:
+            self._fuzz_data_logger.log_fail(str(e))
```
starek4: The same as [here](https://github.com/jtpereyda/boofuzz/pull/349/files/febb8dd2297e31d9c69e3be30e651b71e430b558#r348180471) 😉

boofuzz/sessions.py
kmendoza-bt: Fixed in Commit fc997eb

boofuzz/sessions.py
kmendoza-bt: Fixed in Commit fc997eb

kmendoza-bt: Fixed in Commit fc997eb

boofuzz/sessions.py
```diff
@@ -1162,6 +1169,8 @@ def transmit_fuzz(self, sock, node, edge, callback_data):
             else:
                 self._fuzz_data_logger.log_info(msg)
             pass
+        except (exception.BoofuzzTargetConnectionFailedError, exception.BoofuzzSSLError) as e:
+            self._fuzz_data_logger.log_fail(str(e))
```
SR4ven: Same as comment in `transmit_normal()`.

Catching `BoofuzzTargetConnectionFailedError` not needed; maybe add an option to log SSL errors on info level.

boofuzz/sessions.py
```diff
@@ -1117,6 +1120,8 @@ def transmit_normal(self, sock, node, edge, callback_data):
                 self._fuzz_data_logger.log_fail(msg)
             else:
                 self._fuzz_data_logger.log_info(msg)
+        except (exception.BoofuzzTargetConnectionFailedError, exception.BoofuzzSSLError) as e:
+            self._fuzz_data_logger.log_fail(str(e))
```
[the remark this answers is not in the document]
kmendoza-bt: @SR4ven invalid/fuzzed payload should not cause an SSL error; however, I believe it should be treated as a test case failure. Given the code you referenced, I feel it might be worth giving the user the option to ignore SSL errors and trigger a restart of the target.

In all cases an SSL error would indicate that an EOT or socket-related error has occurred; however, the SSL API is not in a state where it has the complete encrypted data to complete the decryption. This could be caused by a connection loss before the expected encrypted data was received; or, a failed SSL-related operation (such as a session renegotiation).

In my particular case, the SSL error was thrown by ssl.recv() because my target threw a .Net stack exhaustion error. Given the particular SSL "reason" I got, it looks like the issue was the .Net SSL API ran of stack space in the process of encrypting the data that the target was attempting to send back.

boofuzz/sessions.py
kmendoza-bt: Please see my comment for transmit_normal() above.

boofuzz/sessions.py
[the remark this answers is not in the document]
SR4ven: Thanks for the detailed info @kmendoza-bt.

Let's add a session option `ignore_connection_ssl_errors` to log SSL errors on info level so that the test case doesn't fail. Should be false by default.
You can take this code/option for reference:
https://github.com/jtpereyda/boofuzz/blob/c41860f30ed93cfb52a72bc8d103104c2451860e/boofuzz/sessions.py#L1085-L1090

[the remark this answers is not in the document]
kmendoza-bt: Fixed in commit b2b3cec

boofuzz/sessions.py
kmendoza-bt: Fixed in commit b2b3cec

kmendoza-bt:
Ok, I think I royally cooked this PR. I'm going to create a new branch and clean up this PR.

SR4ven:
You could also revert the recent commits and then resolve the conflicts directly on GitHub.

keithmendozasr:
@SR4ven latest changes should address the last review comments.

kmendoza-bt:
@SR4ven latest update includes the new attribute and a blurb on CHANGELOG.rst.

SR4ven:
I quickly applied these changes myself.
If you agree with them, we can merge the PR.

kmendoza-bt:
@SR4ven your last change looks good to me.

SR4ven:
Thanks for the work on this PR @kmendoza-bt!

PR #350, opened 2019-11-19
Fix csv logger data decoding

SR4ven:
Fix for an error when calling `decode()` on non ASCII bytes data while logging sent/received data with the csv logger class.

This fix was split off from #334 to speed up the merging process. Thanks to @JsHuang

A review would be much appreciated @jtpereyda @JsHuang @dorpvom @starek4


SR4ven:
Thanks for the review @starek4!

jtpereyda:
Thanks @SR4ven and @starek4 !

Issue #351, opened 2019-11-20
boofuzz crashes in python2


secenv:
Script and logs shown below. I'm not sure if it's related to the 's_from_file' primitive. It works well when boofuzz is installed through pip3.

```
$ echo -e 'a\nb' > test.txt
$ cat test.py
from boofuzz import *

session = Session( target=Target( connection=SocketConnection( host="127.0.0.1", port=12345, proto='tcp', server=True ),),)
s_initialize("A")
s_from_file("0", filename="test.txt") # same happens with 'encoding="utf8"'
session.connect(session.root, s_get("A"))
session.fuzz()

$ python test.py
[2019-11-20 14:48:57,901]     Info: Web interface can be found at http://localhost:26000
[2019-11-20 14:48:57,903] Test Case: 1: A.no-name.1
[2019-11-20 14:48:57,903]     Info: Type: FromFile. Default value: '0'. Case 1 of 2 overall.
[2019-11-20 14:48:57,903]     Info: Opening target connection (127.0.0.1:12345)...
[2019-11-20 14:48:59,064]     Info: Connection opened.
[2019-11-20 14:48:59,064]   Test Step: Fuzzing Node 'A'
[2019-11-20 14:48:59,064]     Info: Sending 2 bytes...
[2019-11-20 14:48:59,085]     Error!!!! Unexpected exception! Traceback (most recent call last):
                                File "/home/user/.local/lib/python2.7/site-packages/boofuzz/sessions.py", line 703, in _main_fuzz_loop
                                  self._fuzz_current_case(*fuzz_args)
                                File "/home/user/.local/lib/python2.7/site-packages/boofuzz/sessions.py", line 1431, in _fuzz_current_case
                                  self.transmit_fuzz(target, self.fuzz_node, path[-1], callback_data=callback_data)
                                File "/home/user/.local/lib/python2.7/site-packages/boofuzz/sessions.py", line 1114, in transmit_fuzz
                                  self.targets[0].send(data)
                                File "/home/user/.local/lib/python2.7/site-packages/boofuzz/sessions.py", line 157, in send
                                  self._fuzz_data_logger.log_send(data[:num_sent])
                                File "/home/user/.local/lib/python2.7/site-packages/boofuzz/fuzz_logger.py", line 71, in log_send
                                  fuzz_logger.log_send(data=data)
                                File "/home/user/.local/lib/python2.7/site-packages/boofuzz/fuzz_logger_text.py", line 52, in log_send
[...]

PR #352, opened 2019-11-22
Fix s_from_file decoding in Python 2

SR4ven:
Proposed fix for #351 

The problem here is, that file.read() returns a unicode type instead of string. This causes issues when logging the data at a later point.

Fixes:
- Switched to byte wise reading of files for s_from_file to bypass decoding.
- Marked the `encoding` parameter of this primitive as depreciated. It wasn't even implemented before.

TODO:

- [X] Remove tailing newline operator \n from data when reading the file


SR4ven:
Ready to merge from my side.
A review would be very welcome @secenv or anyone else!

SR4ven:
These changes were manually tested on Windows 10 1909 and Ubuntu 19.04 with Python 2 and 3.

Merging to include this fix in the upcoming release.

PR #353, opened 2019-11-29
added crc32 (Castagnoli) as checksum algorithm

keisentraut:
I needed another version of the CRC32 algorithm (Castagnoli variant). This simple pull request adds it. In order to do this, I've used a new external module"crc32c", is this ok?


SR4ven:
Looks good, thank you @keisentraut!

I'll add an import guard for the crc32c package, so that it won't fail importing on systems without support for this.
If you're ok with the changes, we can merge.

keisentraut:
Of course, I'm fine, feel free to merge!

SR4ven:
After having added the new algorithm to the documentation, I think we're good to go.

Thanks for sharing this PR @keisentraut!

PR #355, opened 2019-12-01
Update documentation of s_checksum

SR4ven:
Recently two PRs about new checksum algorithms were opened. There is a possibility to use a custom checksum algorithm with `s_checksum`, which is currently not visible in the online documentation.
Even though it's nice to have a larger variety of checksum algorithms implemented, my guess is that people don't know about this feature, hence opening PRs.

This PR correctly advertises this feature and fixes some other minor copy-paste mistakes in the documentation.

Every review is welcome.

### DIFF
--- a/boofuzz/blocks/checksum.py
+++ b/boofuzz/blocks/checksum.py
@@ -35,7 +35,8 @@ class Checksum(primitives.BasePrimitive):
     Args:
         block_name (str): Name of target block for checksum calculations.
         request (s_request): Request this block belongs to.
-        algorithm (Union[str, function], optional): Checksum algorithm to use. (crc32, adler32, md5, sha1, ipv4, udp)
+        algorithm (str, function, optional): Checksum algorithm to use. (crc32, adler32, md5, sha1, ipv4, udp)
+            Pass a function to use a custom algorithm. This function has to take and return byte-type data.
         length (int, optional): Length of checksum, auto-calculated by default.
             Must be specified manually when using custom algorithm.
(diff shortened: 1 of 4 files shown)


boofuzz/blocks/checksum.py
```diff
@@ -121,9 +122,7 @@ def _should_render_fuzz_value(self):
         return self._fuzzable and (self._mutant_index != 0) and not self._fuzz_complete
 
     def _get_dummy_value(self):
-        if self._length:
-            return self._length * "\x00"
-        return self.checksum_lengths[self._algorithm] * "\x00"
+        return self._length * "\x00"
```
SR4ven: We can simplify this as
https://github.com/jtpereyda/boofuzz/blob/c81f0d9232929fbbe534bb3c850a02e6b6f6ad2f/boofuzz/blocks/checksum.py#L77-L78
will assure that `self._length` will be updated with the correct value before `_get_dummy_value` is called.

If we use a custom function for the checksum algorithm and don't specify it's length, the dummy value will be empty. This is wrong but doesn't have an effect on fuzzing as far as I know. It is now also clearly documented, that `length` needs to be set when using a custom algorithm.

keisentraut:
As the author of one of those two pull requests, I can confirm your suspicion that I did not know about this feature.

SR4ven:
Thanks for the info @keisentraut.
Do the changes in this PR make it clear enough, that custom algorithms are supported?

SR4ven:
No critical changes here. Merging now to include into upcoming release.

PR #357, opened 2019-12-05
Init GitHub Actions

SR4ven:
I see the main advantage of GitHub Actions, in the possibility to test on multiple operating systems.
It is of course also very nicely integrated into GitHub, too.

This PR also contains a fix for Python 3.8 on Windows related to tornado, see https://github.com/tornadoweb/tornado/issues/2608
If we don't merge this PR, we need to apply this fix in another PR.
Additionally I made some preparations for testing on macOS in the socket unit-tests, as macOS doesn't support raw sockets like Linux does. There is still a problem with timeouts, so those tests will be added at another time.

The deployment to PyPI will no longer be triggered by tagged commits, but by creating a release on GitHub now. This was suggested in the default workflow and we could of course change it, but I think the new way would grant us some more flexibility.
For the deployment process to work, the secrets for `PYPI_USERNAME` and `PYPI_PASSWORD` need to be set on the GitHub project settings page @jtpereyda.

Python 3.5 on Windows is exuded from the tests because some dependency for building the netifaces package is missing in the toolchain.

Thoughts about this?

TODO:

- [X] Add changelog entry for tornado fix
- [x] Set PyPI login secrets


SR4ven:
The Actions are not triggered for this PR because the changes are not in master yet. At least that's what I guess.
It does work on my fork:
https://github.com/SR4ven/boofuzz/commit/3e0ccbcde4c06d550150735b90074db297fcc0a8/checks?check_suite_id=344086328

.github/workflows/test.yml
```diff
@@ -0,0 +1,43 @@
...
+    - uses: actions/checkout@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install pip and tox
+      run: python -m pip install --upgrade pip tox
+    - name: Install vcpython27
+      if: matrix.os == 'windows-latest' && matrix.python-version == '2.7'
+      run: choco install vcpython27
```
jtpereyda: choco, love it

unit_tests/test_socket_connection.py
```diff
@@ -425,7 +425,7 @@ def test_udp_broadcast_client(self):
         self.assertEqual(data_to_send, server.received)
         self.assertEqual(received, server.data_to_send)
 
-    @pytest.mark.skipif(sys.platform == "win32", reason="Raw sockets not supported on Windows.")
+    @pytest.mark.skipif(sys.platform in ["win32", "darwin"], reason="Raw sockets not supported on Windows.")
```
jtpereyda: should be `not supported on Windows or Mac OS`.

Or alternatively, two separate `skipif`s with their own message.

.github/workflows/test.yml
SR4ven: Won't be needed for long though (hopefully). #370 will make it obsolete.

SR4ven:
We should keep Travis active for some time until we're sure that the Actions perform as expected.

However, with both services we should avoid pushing a new release. Not that it would mess anything up, but I want to test the deploy action on GitHub.

PR #358, opened 2019-12-06
Split SocketConnection into subclasses

mistressofjellyfish:
Packet sockets may be bound to an interface, restricting the recieved packets to come from a specific interface. Additionally, recieving on raw sockets can be filtered by protocol (like it's done already for
raw-l3). This commit makes this functionality available to the user.

TODO:
- [x] Update documentation consistently
- [x] Update changelog
- [ ] Write unit-test for raw socket receiving (optimal but would be really nice to have)
- [x] Update examples folder (optimal, could be done in another PR)


boofuzz/socket_connection.py
```diff
@@ -148,7 +149,10 @@ def open(self):
             if self._udp_broadcast:
                 self._sock.setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, True)
         elif self.proto == "raw-l2":
-            self._sock = socket.socket(socket.AF_PACKET, socket.SOCK_RAW)
+            raw_proto = self.ethernet_proto if self.ethernet_proto else 0
+            self._sock = socket.socket(socket.AF_PACKET, socket.SOCK_RAW, socket.htons(raw_proto))
+            if self.bind:
+                self._sock.bind(self.bind)
```
SR4ven: We can either use `socket.bind()` while opening or `socket.sendto()` later. Using both is redundant.

However, I haven't quite figured out what the right format of the tuple inside bind() would be for AF_PACKET.
The [docs](https://docs.python.org/3/library/socket.html#socket-families) at AF_PACKET might help here. Maybe something like `bind(self.host, raw_proto)` works. Also I don't think we should reuse the `bind` argument which was meant to be used for UDP receptions.
Have you fully tested this all? Currently, I don't really have the time to do so.

Right now boofuzz does not support receiving on raw sockets at all. While sending, there is no use for filtering by ethertype or MAC. If we manage to get a proper raw packet filter working, we could start to implement the receiving function. Otherwise I don't really see a point in these changes here or am I missing something?

boofuzz/socket_connection.py
```diff
@@ -98,7 +99,7 @@ def __init__(
         self._recv_timeout = recv_timeout
         self._send_timeout = send_timeout
         self.proto = proto.lower()
-        self.ethernet_proto = ethernet_proto
+        self.ethernet_proto = ETH_P_IP if ethernet_proto is None and proto == "raw-l3" else ethernet_proto
```
SR4ven: How about this instead?
```python
if ethernet_proto is None:
    if proto == "raw-l2":
        self.ethernet_proto = 0
    elif proto == "raw-l3":
        self.ethernet_proto = ETH_P_IP
else:
    self.ethernet_proto = ethernet_proto
```
Just a code style question but I think we could gain some readability here.

boofuzz/socket_connection.py
mistressofjellyfish: Nope, you've assessed the situation perfectly. My current recieving setup is a bit... janky to say the least and I need to work on that. I plan to add support for recieving on raw sockets as well, however this may be taking another week or two depending on my workload.

Personally, I'd stay with `socket.bind()` over `socket.sendto()`, especially with receiving in mind. The Tuple inside bind() -- for the AF_PACKET socket -- selects an interface and a protocol (which, for some reason, you have to supply to `socket()` as well). This, to me, seems to stem from some python weirdness, in raw C you would have to supply the protocol in the sockaddr struct to bind as well. Reading `man 7 packet` helped me getting a better understanding of that. For example, to send and capture a multitude of 802.11 frames on my wifi adapter ath0 I'd use `("ath0", ETH_P_ALL)` with `ETH_P_ALL = 0x0003` taken from [<net/if_ether.h>](https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_ether.h)

As for testing: I haven't tested the patch per se apart from ensuring that the socket opens like i'd expect it to. However, in a little script to test stuff opening a socket like this works like expected, so I see no reason why it shouldn't do what I expect here.

As for the bind argument: I don't really see your concern here, apart that I forgot to update the documentation. The bind argument behaves with either protocol exactly how I would expect it to and one is able to guess from the name what happens behind the scenes.

boofuzz/socket_connection.py
mistressofjellyfish: I'll change that, no problem. However, in the grand scheme, it may be a good idea to split the whole Target class and get rid of such argument conflicts and if-if-else-else ladders entirely.

Keep in mind though that this would introduce some breaking changes [0], as it would also mean that the `proto` argument is then surplus - you're directly instanciating a, say, `UdpTarget` instead of `Target(proto=udp` and so on. Having that cleanly OOPed would also make the adding (and testing) of new targets easier as it'd spare me to edit the `Target` class if I wanted to add support for Token Ring or CAN or whatever; and implementing a new target class would probably have less possibilities for side effects than now.

[0] this could always be kinda remedied by naming the new base something like `BaseTarget` and having `Target` as a shim to select the correct implementation while printing deprecated warnings.

boofuzz/socket_connection.py
SR4ven: Sounds good, no hurry.

Agreed, we should use `socket.bind()`, but consistently.
As for the tuple, `self._sock.bind(self.host, raw_proto)` should work fine then. Why would we want to use the `bind` argument for that? Wouldn't the information in that argument be redundant with `host` and `ethernet_proto`?

Ok, we don't really need any special testing in the current state, but once the receiving is implemented, we have to verify that the packet filter works as intended.

boofuzz/socket_connection.py
SR4ven: Good point about splitting the Target class, this would indeed make things easier. Although, I'm not sure if the Target class is the right one to split as it also contains other arguments than just the connection.
Wouldn't it be better to split and rename the SocketConnection class instead so that we can do something like this?
```python
sess = Session(
    target=Target(
        connection=TCPConnection(host=127.0.0.1, port=1234),
        netmon=...),

sess = Session(
    target=Target(
        connection=RawL2Connection(interface="lo", ethernet_proto='0x0800'),
        netmon=...),
```
Either way, we'd have to find a way to keep it backwards compatible.

mistressofjellyfish: Ah, sure, I've meant the SocketConnection class. That's what you get when you run low on sleep.

But yeah, we could totally do that and keep the SocketConnection constructor as-is as kind of a factory for the splitted classes, thus deferring the break until sufficient time has passed. Seeing as the SocketConnection currently is only 350-odd lines, I'll give that a go on Monday.

mistressofjellyfish:
@SR4ven I still have to fix a number of style issues and some tests that rely on SocketConnection being SocketConnection.

While I was at it, I implemented some receiving support into the RawL2/RawL3 classes, which unfortunately introduced some breakage with a semi-private API, as all of the Raw classes now accept a parameter to set their message/frame size. On the raw level, it only makes sense to receive a whole packet anyways, so recv() on those connections returns exactly one of those. That implies however that the MAX_PAYLOADS dict is now removed.

However, I see that as a good thing, because the values there were wrong for anything but bog-standard ethernet in any case as it ignored the existence of Jumboframes or anything other than 802.3, really. [Wikpedia](https://en.wikipedia.org/wiki/Maximum_transmission_unit#IP_MTUs_for_common_media) has a nice overview.

I'll rebase and squash the commits once I sorted the test and style issues. In the meantime, please review.

boofuzz/raw_l3_socket_connection.py
```diff
@@ -0,0 +1,121 @@
...
+        self.ethernet_proto = ethernet_proto
+        self.l2_dst = l2_dst
+        self.packet_size = packet_size
+
+    def open(self):
+        self._sock = socket.socket(socket.AF_PACKET, socket.SOCK_DGRAM)
+
+        super().open()
+
+        self._sock.bind((self.interface, self.ethernet_proto))
```
SR4ven: We can bind the socket before setting sockopts with super().open()
See raw socket sniffer [example](https://docs.python.org/3/library/socket.html#example).

boofuzz/raw_l3_socket_connection.py
```diff
@@ -0,0 +1,121 @@
...
+        Returns:
+            Received data
+        """
+
+        data = b""
+
+        try:
+            data = self._sock.recv(self.packet_size)
+
+            if max_bytes > 0 and max_bytes < self.packet_size:
```
SR4ven:
```suggestion
            if 0 < max_bytes < self.packet_size:
```

boofuzz/raw_l3_socket_connection.py
```diff
@@ -0,0 +1,121 @@
...
+            data = b""
+        except socket.error as e:
+            if e.errno == errno.ECONNABORTED:
+                raise_(
+                    exception.BoofuzzTargetConnectionAborted(socket_errno=e.errno, socket_errmsg=e.strerror),
+                    None,
+                    sys.exc_info()[2],
+                )
+            elif e.errno in [errno.ECONNRESET, errno.ENETRESET, errno.ETIMEDOUT]:
+                raise_(exception.BoofuzzTargetConnectionReset(), None, sys.exc_info[2])
```
SR4ven:
```suggestion
                raise_(exception.BoofuzzTargetConnectionReset(), None, sys.exc_info()[2])
```

boofuzz/socket_connection.py
```diff
@@ -1,34 +1,38 @@
...
+    ethernet_proto=None,
+    l2_dst=b"\xFF" * 6,
+    udp_broadcast=False,
+    server=False,
+    sslcontext=None,
+    server_hostname=None,
+):
     """ITargetConnection implementation using sockets.
 
     Supports UDP, TCP, SSL, raw layer 2 and raw layer 3 packets.
```
SR4ven: Add depreciation warning to docs.

boofuzz/ssl_socket_connection.py
```diff
@@ -0,0 +1,86 @@
...
+        recv_timeout (float): Seconds to wait for recv before timing out. Default 5.0.
+        server (bool): Set to True to enable server side fuzzing.
+        sslcontext (ssl.SSLContext): Python SSL context to be used. Required if server=True or server_hostname=None.
+        server_hostname (string): server_hostname, required for verifying identity of remote SSL/TLS server
+    """
+
+    def __init__(
+        self, host, port, send_timeout=5.0, recv_timeout=5.0, server=False, sslcontext=None, server_hostname=None
+    ):
+        super().__init__(host, port, send_timeout, recv_timeout, server)
```
SR4ven: Initialisation of `self.sslcontext` and `self.server_hostname` missing.

boofuzz/ssl_socket_connection.py
```diff
@@ -0,0 +1,86 @@
...
+            max_bytes (int): Maximum number of bytes to receive.
+
+        Returns:
+            Received data.
+        """
+        try:
+            super().recv(max_bytes)
+        except ssl.SSLError as e:
+            # If an SSL error is thrown the connection should be treated as lost
+            raise_(exception.BooFuzzSSLError(e.reason))
```
SR4ven:
```suggestion
            raise_(exception.BoofuzzSSLError(e.reason))
```

boofuzz/ssl_socket_connection.py
```diff
@@ -0,0 +1,86 @@
...
+        Send data to the target. Only valid after calling open!
+
+        Args:
+            data: Data to send.
+
+        Returns:
+            int: Number of bytes actually sent.
+        """
+
+        num_bytes = 0
```
SR4ven:
```suggestion
        num_sent = 0
```

And add `return num_sent` at the end.

boofuzz/tcp_socket_connection.py
```diff
@@ -0,0 +1,135 @@
...
+                    raise exception.BoofuzzTargetConnectionFailedError(str(e))
+                else:
+                    raise
+
+        try:
+            self._sock.connect((self.host, self.port))
+        except socket.error as e:
+            if e.errno == errno.EADDRINUSE:
+                raise exception.BoofuzzOutOfAvailableSockets()
+            elif e.errno in [errno.ECONNREFUSED, errno.EINPROGRES, errno.ETIMEDOUT]:
```
SR4ven:
```suggestion
            elif e.errno in [errno.ECONNREFUSED, errno.EINPROGRESS, errno.ETIMEDOUT]:
```

boofuzz/tcp_socket_connection.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        Receive up to max_bytes data from the target.
+
+        Args:
+            max_bytes (int): Maximum number of bytes to receive.
+
+        Returns:
+            Received data.
+        """
+        data = b""
```
SR4ven: `return data` is missing

boofuzz/udp_socket_connection.py
```diff
@@ -0,0 +1,193 @@
...
+        """Opens connection to the target. Make sure to call close!
+
+        Returns:
+            None
+        """
+
+        self._sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+
+        if self.bind:
+            # TODO: should this be try/except'd?
```
SR4ven: Depends, whether or not we can run into `EADDRINUSE`. We had this problem for TCP sockets #296.
I'd say leave it as it is for now and keep the TODO in.

boofuzz/udp_socket_connection.py
```diff
@@ -0,0 +1,193 @@
...
+            self._sock.bind(self.bind)
+
+        if self.broadcast:
+            self._sock.setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, True)
+
+        super().open(self)
```
SR4ven:
```suggestion
        super().open()
```

boofuzz/udp_socket_connection.py
```diff
@@ -0,0 +1,193 @@
...
+            data = b""
+        except socket.error as e:
+            if e.errno == errno.ECONNABORTED:
+                raise_(
+                    exception.BoofuzzTargetConnectionAborted(socket_errno=e.errno, socket_errmsg=e.strerror),
+                    None,
+                    sys.exc_info()[2],
+                )
+            elif e.errno in [errno.ECONNRESET, errno.ENETRESET, errno.ETIMEDOUT]:
+                raise_(exception.BoofuzzTargetConnectionReset(), None, sys.exc_info[2])
```
SR4ven:
```suggestion
                raise_(exception.BoofuzzTargetConnectionReset(), None, sys.exc_info()[2])
```

docs/user/connections.rst
```diff
@@ -14,9 +14,44 @@ ITargetConnection
     :undoc-members:
     :show-inheritance:
 
-SocketConnection
-================
-.. autoclass:: boofuzz.SocketConnection
```
SR4ven: Keep `SocketConnection` in the docs until we completely remove it.
Move it to the end of the list and add a notice about the depreciation.

boofuzz/base_socket_connection.py
```diff
@@ -0,0 +1,59 @@
...
+    def close(self):
+        """
+        Close connection to the target.
+
+        Returns:
+            None
+        """
+        self._sock.close()
+
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
mistressofjellyfish: Shouldn't all these members just tickle through from the baseclass? At least in my ipython shell, this works as expected:

    In [1]: from boofuzz import BaseSocketConnection                                                                    
    In [2]: class t(BaseSocketConnection): 
 
      ...:     def __init__(self): 
      ...:         super().__init__(5.0, 5.0) 
      ...:                                                                                                             
    In [3]: s = t()                                                                                                     
    ---------------------------------------------------------------------------
    TypeError                                 Traceback (most recent call last)
    <ipython-input-3-e764f24fca13> in <module>
    ----> 1 s = t()
    TypeError: Can't instantiate abstract class t with abstract methods info, open, recv, send
    In [4]:  

As for recv(), personally, I'd rather see this implemented in the respective classes and not in the base class as this would prompt developers to write some docs on why recv() is unimplemented thus making leaving it unimplemented a conscious decision.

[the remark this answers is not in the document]
SR4ven: Sorry my bad, I completely forgot about `ITargetConnection`.
You are right, keep it as it is.

docs/user/connections.rst
mistressofjellyfish: Unfortunately sphinx doesn't have a block for notices, so I settled for a warning.

mistressofjellyfish:
Okay, so I finally got around to fix almost all of the remaining issues and made every test working again. Yay!

I've not gone forward in moving all the connection classes  to their own respective folder. As this would -- and should -- move the `ITargetConnection` and `SerialConnection` classes aswell, I feel that it would be more appropriate to do that in another PR.

I see what I can do rebasing my PR.

mistressofjellyfish:
Seems like the CI likes to run again now, but I think I introduced some mess into the history (dunno what exactly happened here -- sorry). Regardless, that's fixable by carefully squashing commits together, so it shouldn't be too much of a problem. I'll leave it as-is right now unless you want me to do some cleanup @SR4ven.

CHANGELOG.rst
```diff
@@ -8,12 +8,25 @@ v0.1.6
...
+- Pushed Tornado to 5.x and unpinned Flask
+- Added a favicon
+- new primitive `s_bytes` which fuzzes an arbitrary length binary value (similiar to `s_string`)
+- We are now using `Black` for code style standardization
+- Added support for Python 3.8
+- added crc32c as checksum algorithm (Castagnoli)
+>>>>>>> 661f6ecc41079291bcd8d398537fc74fe5e1d513
+- Rewrote and split the SocketConnection class into individual classes per socket type.
+- Added support for receiving on raw Layer 2 and Layer 3 connections
+- Layer 2 and Layer 3 connections may now use arbitrary payload / MTU sizes
```
SR4ven: Looks like some problems while merging.
Version 0.1.6 has been released, so these changes should go under `Upcoming`

boofuzz/ssl_socket_connection.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def open(self):
+        super(SSLSocketConnection, self).open()
+
+        # If boofuzz is the SSL client and user did not give us a SSLContext,
+        # then we just use a default one.
+        if self.server is False and self.sslcontext is None:
+            self.sslcontext = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+            self.sslcontext.check_hostname = True
+            self.sslcontext.verify_flags = ssl.CERT_REQUIRED
```
SR4ven:
```suggestion
            self.sslcontext.verify_mode = ssl.CERT_REQUIRED
```
Took me some time debugging this :)

boofuzz/tcp_socket_connection.py
```diff
@@ -0,0 +1,144 @@
...
+            data = self._sock.recv(max_bytes)
+        except socket.timeout:
+            data = b""
+        except socket.error as e:
+            # Issue 348, PR 361: SSLError handling on python < 3.3
+            # Before Python 3.3, SSLError used to be a subtype of socket.error.
+            # To ensure it is properly handled in SSLSocketConnection, we have to
+            # re-raise it here if we're handling a ssl.SSLErorr.
+            if type(e) is ssl.SSLError:
+                raise
```
SR4ven: I don't think this is needed. We already cover this case by `else: raise`

unit_tests/test_socket_connection.py
```diff
@@ -553,10 +551,11 @@ def test_raw_l2_oversized(self):
 
         uut = SocketConnection(host="lo", proto="raw-l2")
```
SR4ven: We should switch to the new classes here too. Else we get misleading depreciation warnings while running the tests.

starek4: Imho, `SocketConnection` still should be tested at least for some scenarios, even if it is deprecated. Until it will be removed, we still need to know, if it is working properly. I believe that deprecation warnings when running tests are not a problem.

boofuzz/udp_socket_connection.py
starek4: As UDP is connectionless, there is no `TIME_WAIT` interval or something similar to that. So no additional waiting after closing a socket is performed. I believe that `EADDRINUSE` can occur only when we use accidentally use the same port number twice without closing them. The workaround with waiting does not make sense for me here.

I think we are good to go without trying to catch `EADDRINUSE`. I suggest to remove `# TODO: should this be try/except'd?` comment.

starek4:
Splitting the `SocketConnection` class is really a good idea. The code is much more readable now 👍 

I am also really glad that we keep its previous interface working and marking it as deprecated so users have time to reimplement their tools.

boofuzz/tcp_socket_connection.py
mistressofjellyfish: I‘m not too sure either; however I wanted to play it safe when merging. I couldn’t figure out a quick unit test for this case, so I decided to go the explicit route and raise as early as possible while documenting why.

If we have a testcase that demonstrates that this isn‘t a problem with the new structure, we should remove that bit of course.

boofuzz/udp_socket_connection.py
SR4ven: Thanks for the input @starek4, makes sense.
Let's remove the TODO here.

boofuzz/tcp_socket_connection.py
SR4ven: I did some manual testing and both `ssl.SSLError` and `socket.error` both inherit from `OSError`. According to [this](https://docs.python.org/3/library/socket.html#socket.error), `socket.error` is just a depreciated alias for `OSError`. So catching `socket.error` will always also catch `ssl.SSLError`.
I verified this by some manual testing on Python 2.7 and 3.7 with a debugger.

I also ran some tests with
```python
if type(e) is ssl.SSLError:
    raise
```
and found out that the type of my provoked ssl error was of the type `SSLCertVerificationError` which doesn't match `ssl.SSLError`.

In conclusion, it's safe to remove that code.

unit_tests/test_socket_connection.py
SR4ven: Sounds reasonable. I also found a way to ignore the depreciation warnings for `SocketConnection` for PyCharm and pytest, so that's not a problem anymore.

unit_tests/test_socket_connection.py
```diff
@@ -595,40 +604,51 @@ def test_raw_l3(self):
...
 
         t = threading.Thread(target=functools.partial(server.receive_until, expected_server_receive))
         t.daemon = True
         t.start()
 
         # When
         uut.open()
-        send_result = uut.send(data=raw_packet)
+        send_result = uut.send(data=raw_packet_send)
+        _ = uut.recv(10000)  # Receiving the sent package. Will not work on a busy socket!
```
SR4ven: This seems to be a common problem with raw sockets.
Could be fixed by filtering received packages by their MAC-address.

unit_tests/test_socket_connection.py
```diff
@@ -245,8 +246,9 @@ def receive_until(self, expected):
                         if not self.stay_silent:
                             self.server_socket.sendto(self.data_to_send, addr)
                         break
-                except socket.timeout:
-                    break
+                except Exception as e:
+                    self.exception = e
+                    raise
```
SR4ven: I noticed that there was an exception while executing the `sendto()` part the whole time which remained undetected because pytest only checks for exceptions in the main thread. Hence this workaround where we need to check if `exception` is None in every test.
If anyone knows a better solution to make pytest thread aware, please let us know.

mistressofjellyfish:
> Looking at the tests for raw l2/3 connections, I noticed that they don't really work as intended. [...]
> The main problem seems to be that we receive our own outgoing packets because we only have a protocol filter, but no MAC-filter.

Well, we're using the loopback interface for testing. It's no surprise that we receive the packet we've just sent. That should not happen with other raw sockets, where the packet gets actually sent over the line / air / whatever as sent packets won't get dropped into the kernels receive queue.

SR4ven:
Right, I completely forgot that we test on loopback... but that problem can be easily solved by simply "assigning" a different MAC address to the MiniTestServer (which works on raw-l2 so it can mimic any MAC).
For this to work we'd also need some filtering in the receive function, so that we only catch packets which come from our target instead of any that match the ether_type.
I quickly put together some code that seems to work well so far. It replaces `data = self._sock.recv(self.packet_size)` in RawL3SocketConnection.recv

```python
elapsed_time = 0
start_time = time.time()
while elapsed_time < self._recv_timeout:
    self._sock.setsockopt(
        socket.SOL_SOCKET,
        socket.SO_RCVTIMEO,
        base_socket_connection.seconds_to_sockopt_format(self._recv_timeout - elapsed_time),
    )
    received, addr = self._sock.recvfrom(self.packet_size)
    print(addr)
    # See https://docs.python.org/3/library/socket.html at AF_PACKET adress tupel
    # https://docs.huihoo.com/doxygen/linux/kernel/3.7/if__packet_8h.html
    if (
        (list(addr)[2] == 0 and list(addr)[4] == self.l2_dst)  # Unicast from target MAC
        or list(addr)[2] == 1  # Bradcast
        or list(addr)[2] == 2  # Multicast
    ):
        data = received
        break
    elapsed_time = time.time() - start_time
```

However, writing unit tests and the MAC filter seems to require quite some extra work. We should perhaps make these changes in a different PR to keep things simple.
That would mean reverting 7cddcfd95b091921b1f18e557fbcff7be55d767c and merging this PR first, if that's fine by you @mistressofjellyfish.

mistressofjellyfish:
Hm... I don't think that this is a particularly elegant way to address this. My main concern is that I then have to provide a MAC to filter by. While I don't have a convincing argument on why I would have a scenario where I would want to have a catch-all, I don't have an argument why such a scenario could not exist.

Also, while this is rather trivial to implement for Layer 3 (iff Layer 2 is Ethernet), it's a whole different story on Layer 2, which in the current form should support stuff like 802.11 or 802.5.

So, in the end, I'd much rather write documentation about this issue and encouraging people to either pass a filter function or provide their own subclass of one of the raw classes. With the new design, that is a fairly straight forward to do after all, and we could always just do something like this:

```python
def recv(self):
    elapsed_time = 0
    start_time = time.time()
    while elapsed_time < self._recv_timeout:
        self._sock.setsockopt(
            socket.SOL_SOCKET,
            socket.SO_RCVTIMEO,
            base_socket_connection.seconds_to_sockopt_format(self._recv_timeout - elapsed_time),
        )
        received, addr = self._sock.recvfrom(self.packet_size)
    
        # See https://docs.python.org/3/library/socket.html at AF_PACKET adress tupel
        # https://docs.huihoo.com/doxygen/linux/kernel/3.7/if__packet_8h.html
        if self.accept_packet(addr, received):
            data = received
            break
        elapsed_time = time.time() - start_time

def accept_packet(self, addr, received):
    return True
```

So anyone that wishes to filter just subclasses the appropriate Raw class and implements the accept_packet function.

With regards to the tests - because we know what is going on, I see no harm in doing a double-receive there, as long as it is documented why.

SR4ven:
>Hm... I don't think that this is a particularly elegant way to address this. My main concern is that I then have to provide a MAC to filter by. While I don't have a convincing argument on why I would have a scenario where I would want to have a catch-all, I don't have an argument why such a scenario could not exist.

While I agree that a catch-all scenario might come in handy at some point, I think the case where one would like to receive from just the target's MAC seems to be more common.
IMHO we should provide a built in MAC filter, which can be overwritten by a custom function or completely disabled. Though this PR will make it much easier to implement things like a MAC-filter, my guess is that average user can't or doesn't want to deal with such stuff.
After all, boofuzz is supposed to simplify fuzzing :)

>Also, while this is rather trivial to implement for Layer 3 (iff Layer 2 is Ethernet), it's a whole different story on Layer 2, which in the current form should support stuff like 802.11 or 802.5.

As I don't know too much about 802.11 or 802.5, I can't tell why a MAC-filter could problematic. Aren't we somewhat bound to 802.3 because of the AF_PACKET socket anyway?
However, no need to commit to a solution yet, but as before, I'd tend to at least implement the option to pass a function to check the packet.

>So anyone that wishes to filter just subclasses the appropriate Raw class and implements the accept_packet function.

While this would work just fine, why not use a callback function instead? Boofuzz already offers support for pre/post/restart callbacks etc. so that procedure should be know to users.

>With regards to the tests - because we know what is going on, I see no harm in doing a double-receive there, as long as it is documented why.

The problem I see are sporadic failures on busy loopback interfaces. No idea how frequent this could happen, but if it does on CI tests, it would be really annoying as I can't retrigger them. There is an option to retry failed tests (https://github.com/pytest-dev/pytest-rerunfailures), but if possible I'd like to avoid such stuff.
However, fixing the unit-tests is something we need to worry about in another PR.

>Edit: If we are not going the minmally invasive route, I think we should merge this PR now and discuss further in a second PR in order to avoid future merge conflicts. After all, this is a major change from an implementation perspective; and other contributors shouldn't base their changes on the SocketConnection class if it is clear that we're going to merge this PR eventually.

I'll revert the changes I made to the unit-tests and review again. I hope that @jtpereyda has the time to lay his eyes on this before we merge, it's a pretty big change after all.

unit_tests/test_socket_connection.py
SR4ven: Will be dealt with in another PR.

unit_tests/test_socket_connection.py
SR4ven: Will be dealt with in another PR.

mistressofjellyfish:
> While I agree that a catch-all scenario might come in handy at some point, I think the case where one would like to receive from just the target's MAC seems to be more common.
> IMHO we should provide a built in MAC filter, which can be overwritten by a custom function or completely disabled. Though this PR will make it much easier to implement things like a MAC-filter, my guess is that average user can't or doesn't want to deal with such stuff.
> After all, boofuzz is supposed to simplify fuzzing :)

I think you know the userbase better than I do; however I'd much rather enable stuff than disable it - it is always much harder to get rid of than adding functionality. Passing a validation function is a practical solution, however I'd still argue the default should be just to accept everything while providing some default filters.

To some degree, we could also just make all filters implement some Interface and have one of the filters passed when instantiating the socket connection, like so:

```python
foo = RawL3SocketConnection(interface="lo", packet_filter=EthernetMacFilter("aa:bb:cc:dd:ee:ff")
```

And for the default argument, we just use a `AcceptAllFilter()` or whatever.
 
> As I don't know too much about 802.11 or 802.5, I can't tell why a MAC-filter could problematic. Aren't we somewhat bound to 802.3 because of the AF_PACKET socket anyway?
> However, no need to commit to a solution yet, but as before, I'd tend to at least implement the option to pass a function to check the packet.
At least for 802.11, it's an AF_PACKET socket, and it should be for 802.5 as well, as we're just changing the Layer 2 protocol. The Problem whit implementing that "naively" is that the offset where the target and source mac adress resides in these protocols is different than in 802.3 ethernet. In any case, those filters depend on the User knowing which Layer 2 protocol to use.

 
> > So anyone that wishes to filter just subclasses the appropriate Raw class and implements the accept_packet function.
> While this would work just fine, why not use a callback function instead? Boofuzz already offers support for pre/post/restart callbacks etc. so that procedure should be know to users.
See above. I don't have any hard feelings there, as long as the solution that gets implemented is easily extendable without modifying boofuzz source code. 

> > With regards to the tests - because we know what is going on, I see no harm in doing a double-receive there, as long as it is documented why.
> 
> The problem I see are sporadic failures on busy loopback interfaces. No idea how frequent this could happen, but if it does on CI tests, it would be really annoying as I can't retrigger them. There is an option to retry failed tests (https://github.com/pytest-dev/pytest-rerunfailures), but if possible I'd like to avoid such stuff.
> However, fixing the unit-tests is something we need to worry about in another PR.

Hm, that's indeed a valid concern I haven't thought about. In any case, introducing more code there won't fix the underlying issue of the loopback interface and still yields code paths that can't be tested. Is it possible to set up a virtual ethernet thingy in the CI? [RedHat](https://developers.redhat.com/blog/2018/10/22/introduction-to-linux-interfaces-for-virtual-networking/#veth) has a nice description. That would give us a 100% private interface for testing, which actually has two ends so we circumvent this problem entirely. At least for the raw tests, I think that this will be closer to a practical application as well.

SR4ven:
>I think you know the userbase better than I do; however I'd much rather enable stuff than disable it - it is always much harder to get rid of than adding functionality. Passing a validation function is a practical solution, however I'd still argue the default should be just to accept everything while providing some default filters.

Sounds reasonable.
Whether to use an interface for classes or callback is something we can discuss during implementation.

>At least for 802.11, it's an AF_PACKET socket, and it should be for 802.5 as well, as we're just changing the Layer 2 protocol. The Problem whit implementing that "naively" is that the offset where the target and source mac adress resides in these protocols is different than in 802.3 ethernet. In any case, those filters depend on the User knowing which Layer 2 protocol to use.

In the example I posted earlier, the source mac was extracted from the address tuple returned by recvfrom(). So as long as this function is aware of the underlying layer, I see no problem here. It even works for layer 2 ethernet.

>Hm, that's indeed a valid concern I haven't thought about. In any case, introducing more code there won't fix the underlying issue of the loopback interface and still yields code paths that can't be tested. Is it possible to set up a virtual ethernet thingy in the CI? RedHat has a nice description. That would give us a 100% private interface for testing, which actually has two ends so we circumvent this problem entirely. At least for the raw tests, I think that this will be closer to a practical application as well.

Virtual ethernet would fix that problem, but the issue here might be the implementation. First I'm not sure if the CI env allows it (most likely does) and second this would mean that dev pcs that run local tests would have to provide those interfaces too.
Currently I'd tend to make this one test optional and rely on other tests using mac filters. If a test with a mac filter passes, how likely is it that the receive all fails...
On top all CI tests run in either a dedicated VM or docker container, so technically there shouldn't be much traffic on loopback, it's just that we cannot guarantee it.

boofuzz/raw_l3_socket_connection.py
```diff
@@ -0,0 +1,122 @@
...
+    ):
+        super(RawL3SocketConnection, self).__init__(send_timeout, recv_timeout)
+
+        self.interface = interface
+        self.ethernet_proto = ethernet_proto
+        self.l2_dst = l2_dst
+        self.packet_size = packet_size
+
+    def open(self):
+        self._sock = socket.socket(socket.AF_PACKET, socket.SOCK_DGRAM)
```
SR4ven: Why don't we open the socket with ethernet proto like we do on raw l2?
https://github.com/jtpereyda/boofuzz/blob/c303290dcd79da2af0ffc24d56ca8203f03bad0d/boofuzz/raw_l2_socket_connection.py#L48
Might be redundant because we bind the socket later.
I ran some tests and everything works as intended. Should we still add it for consistency?

boofuzz/raw_l2_socket_connection.py
```diff
@@ -0,0 +1,130 @@
...
+            elif e.errno in [errno.ECONNRESET, errno.ENETRESET, errno.ETIMEDOUT, errno.EPIPE]:
+                raise_(exception.BoofuzzTargetConnectionReset(), None, sys.exc_info()[2])
+            else:
+                raise
+
+        return num_sent
+
+    @property
+    def info(self):
+        return "{0}, type 0x{1:04x}".format(self.interface, self.ethernet_type)
```
SR4ven:
```suggestion
        return "{0}, type 0x{1:04x}".format(self.interface, self.ethernet_proto)
```

boofuzz/raw_l3_socket_connection.py
mistressofjellyfish: Hm, this code came verbatim from what was done in the old SocketConnection. But, as you mentioned, I suppose we should unify this.

mistressofjellyfish:
> Regarding merging, do you want to squash the commits to keep a certain history or can I simply squash everything into one commit?

Just squash it all together, that is good. Regarding email; it really doesn't matter.

boofuzz/raw_l2_socket_connection.py
```diff
@@ -0,0 +1,130 @@
...
+        interface (str): Hostname or IP adress of target system.
+        send_timeout (float): Seconds to wait for send before timing out. Default 5.0.
+        recv_timeout (float): Seconds to wait for recv before timing out. Default 5.0.
+        ethernet_proto (int): Ethernet protocol to bind to. If supplied, the opened socket
+            gets bound to this protocol, otherwise the python default of 0 is used. Must
+            be supplied if this socket should be used for receiving. For valid options,
+            see <net/if_ether.h> in the Linux Kernel documentation. Usually, ETH_P_ALL
+            (0x0003) is not a good idea.
+        mtu (int): sets the maximum transmission unit size for this connection. Defaults
+            to 1518 for standard Ethernet.
```
SR4ven: The Ethernet 2 frame consists of 14 bytes header, up to 1500 bytes data and 4 bytes CRC, totalling 1518 bytes.
However, setting this as the default value requires the NIC to not handle the CRC on it's own. While this might be possible with some NICs (check [this](https://stackoverflow.com/questions/723635/how-do-you-send-an-ethernet-frame-with-a-corrupt-fcs)), it's definitely not the default case for raw-l2 socket usage.
We should change the default value back to 1514.

Thoughts about this?

mistressofjellyfish: That's why I introduced the has_framecheck argument (with a default value of True) to substract 4 bytes if the NIC calculates the CRC. Again, 802.11 cards usually don't inject the CRC; and for 802.5 the CRC is somewhere in the middle of the frame -- although I'm not sure about how much we can actually change in the 802.5 frame (and I don't currently have token ring equipment on hand to check, even if I'd be interested).

SR4ven: My bad, I forgot about the has_framecheck argument.
Thanks for the quick reply, I'll finally merge this shortly.

SR4ven:
Many thanks to @mistressofjellyfish for the great work!

jtpereyda:
Thanks all! Nice PR @mistressofjellyfish . The big ol' if statements in the previous code were an antipattern IMO.

PR #359, opened 2019-12-09
Fix for various potential type errors

SR4ven:
I used [pytype](https://github.com/google/pytype) to identify and fix many potential type errors.
Those for which I did not know any solution, I have added exceptions.

There's one change I'm unsure about.
As far as I understand, we currently create a root node from the class `Node` during initialisation. After that, we are able to connect more nodes. Those nodes are the ones we create during the protocol definition with `s_initialize`. The base class seems to be `IFuzzable`, which implements properties and methods like `mutant`. The class `Node` doesn't implement these, that's why pytype rises following error:
```python
File "/home/dev/PycharmProjects/boofuzz/boofuzz/sessions.py", line 908, in _process_failures: No attribute 'mutant' on boofuzz.pgraph.node.Node [attribute-error]
```
This is because when pytype runs, there is only the root node present, which doesn't have the attributes that are being called. During runtime this will never be a problem, because there will always be another node, that does have these attributes.

Now I'm uncertain whether we should add ignores to all calls to attributes on nodes that are being implemented by `IFuzzable` or not. For now I added `IFuzzable` as a base class to `Node`. This also required overwriting the name property of `IFuzzable`.
Definitely not the cleanest solution, but inline exceptions for around 20 errors didn't seem much better.

Suggestions and reviews welcome!


boofuzz/pedrpc.py
```diff
@@ -42,9 +42,9 @@ def __connect(self):
         self.__disconnect()
 
         # connect to the server, timeout on failure.
+        self.__server_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.__server_sock.settimeout(3.0)
         try:
-            self.__server_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            self.__server_sock.settimeout(3.0)
             self.__server_sock.connect((self.__host, self.__port))
```
jtpereyda: @SR4ven Did pytype flag this one because more code was in the `try` than necessary? If so that's pretty neat.

examples/fuzz_trend_control_manager_20901.py
```diff
@@ -6,10 +6,14 @@
...
+from boofuzz import Session, Target, s_get
+
 # noinspection PyUnresolvedReferences
-from requests import trend  # noqa: F401
+# pytype: disable=import-error
+from request_definitions import trend  # noqa: F401
+
+# pytype: enable=import-error
 
-from boofuzz import *
```
jtpereyda: Does pytype flag `import *`s?

unit_tests/test_fuzz_logger.py
```diff
@@ -178,15 +178,15 @@ def test_error_count(self):
         self.assertEqual(["FAILURE 1"], self.logger.error_test_cases["a"])
 
         # Use a number to verify that non-string keys can work
-        self.logger.open_test_case(test_case_id=5, name=self.some_other_text, index=self.some_int)
+        self.logger.open_test_case(test_case_id="5", name=self.some_other_text, index=self.some_int)
```
jtpereyda: The comment implies that a number is intentional for this test. The idea is that if one passes in a plain number, it will get converted to a string. Maybe the code in open_test_case can be changed somehow to flag to pytype that int and str are allowed?

jtpereyda: Or maybe the number more harmful than I realize...

unit_tests/test_socket_connection.py
```diff
@@ -259,6 +259,8 @@ class TestSocketConnection(unittest.TestCase):
     hardware or network dependent.
     """
 
+    # TODO: Remove pytype ignore when SocketConnection is removed
+    # pytype: disable=attribute-error
```
jtpereyda: Why does SocketConnection mean we need the ignore?

boofuzz/pedrpc.py
SR4ven: No it's not that smart sadly.
The error here is:
```python
pedrpc.py, line 62, in __connect: No attribute 'setsockopt' on None [attribute-error]
  In Optional[socket.socket]
```
Pytype assumes that the assignment of `self.__server_sock` fails, which could be caught by the try/except. That again would let the program continue to `self.__server_sock.settimeout(None)` which would then fail because the variable is not an instance of socket.

examples/fuzz_trend_control_manager_20901.py
SR4ven: Only when scanning single files as it seems to be missing the context then.
When running it on the whole project, star imports work fine.

unit_tests/test_socket_connection.py
SR4ven: Since SocketConnection is deprecated it only passes the arguments to the appropriate class based on BaseSocketConnection.
Pytype raises following error without the attribute-error ignore:
```python
test_socket_connection.py", line 285, in test_tcp_client: No attribute 'open' on None [attribute-error]
  In Optional[Union[boofuzz.connections.raw_l2_socket_connection.RawL2SocketConnection, boofuzz.connections.raw_l3_socket_connection.RawL3SocketConnection, boofuzz.connections.tcp_socket_connection.TCPSocketConnection, boofuzz.connections.udp_socket_connection.UDPSocketConnection]]
```
I can only guess, but I think this is because of the `raise` statements in SocketConnection which would lead to a None return.

The error doesn't appear when referring to the new socket classes directly, hence the TODO.

unit_tests/test_fuzz_logger.py
SR4ven: Good point, I didn't even notice the comment... must have been late that day :D

I added a type hint to FuzzLogger and reverted the changes to test_fuzz_logger.

SR4ven:
Hmm, Actions tests failed twice already on serial connection test (2 different python versions).
My guess is that the runners are under heavy lead and our tests aren't robust enough.

boofuzz/pedrpc.py
jtpereyda: Ah, thanks -- still a nice catch.

jtpereyda:
Changes approved. Were both fails on the same exact config (Test / Python 3.8)?

SR4ven:
Thanks for the review @jtpereyda 

Both fails were on different Python versions, but always on Windows and the same test case ([TestSerialConnection.test_recv_message_separator_time](https://github.com/jtpereyda/boofuzz/blob/00cd32f3c20358654988baaa070f97430bd7ecde/unit_tests/test_serial_connection_generic.py#L352)). 
Taking a very brief look at this test, we seem to be testing timeouts. My guess is that the runners were under high load, so the task got delayed/interrupted a little which resulted in our timings being off.

Issue #193, opened 2018-06-29
Retry indefinitely on connection issues instead of failing immediately


jtpereyda:
A lot of communication errors result in a failed test case currently. It makes more sense to engage in some retry behavior, with a possible long term timeout. It is meaningless to push forward with a test if connection errors are still in place.

Add some kind of retry behavior (possibly in the Target class?). Depending on some setting:
 1. Retry indefinitely or
 2. Retry for n retries or x amount of time or
 3. Don't retry at all

jtpereyda:
Might as well retry indefinitely for most vulnerability research use cases. Failing could be useful in some scenarios like signaling test failure to an automated system.

dorpvom:
I tried addressing this in #363

PR #363, opened 2019-12-10
193 restart threshold

dorpvom:
This addresses #193 .

Default behavior (indefinite retries) unchanged, but offering two options:

- [`restart_threshold`] Stop connection attempts after n tries
- [`restart_timeout`] Stop connection attempts after n seconds

First option can be combined with `restart_sleep_time` to basically achieve the second, but for ease of use added here nevertheless.


SR4ven:
Thanks for your contribution @dorpvom!

Two things:
- #193 is about retrying on any connection related problems. This is something that could be best implemented in the Target class. Currently, this PR would only provides the ability to limit the number of retries for opening a connection to the target. Connection errors while sending or receiving are not effected by this.
- If your goal was to add a retry threshold/timeout for the restart function, you might be better off implementing that inside the `_restart_target()` function itself. There are some other calls to this function than the one in `_open_connection_keep_trying` which would ignore the threshold else wise.
But then we might have a problem with checking whether the target has restarted successfully or not, which we'd have to know to reset the counter/timer. Some restarting methods don't have any feedback, so this might become problematic (maybe simply try to open a connection and close it right after?).

dorpvom:
Hi,
maybe you can help clarify if my intention here was correct. The scenario I was facing was both

- trying to open a connection to a host/port that is not even reachable
- trying to reconnect to a host/port that has crashed and did not come back up

I encountered the latter but thought the code to handle both cases together.

The indefinite retry was problematic since I have a remote scenario without the ability to attach a debugger, thus having no idea that the service crashed (i.e. segfaulted).
Now as I am integrating my work as piece of an automation context, I need some kind of timeout on the retry and would prefer not to induce it through calling boofuzz as subprocess.

Thanks for taking the time to watch at my code.

dorpvom:
Thanks again, I'll try to finish this in the next days. Got sidetracked during christmas break ...

boofuzz/sessions.py
```diff
@@ -1505,13 +1512,26 @@ def _open_connection_keep_trying(self, target):
...
+
             while True:
                 try:
                     target.open()
                     break  # break if no exception
                 except exception.BoofuzzTargetConnectionFailedError:
                     self._fuzz_data_logger.log_info(constants.WARN_CONN_FAILED_TERMINAL)
+                    if unable_to_connect_count == self.restart_threshold:
+                        raise exception.BoofuzzTargetConnectionFailedError(
+                            "Unable to reconnect target: Reached threshold. Ending fuzzing."
```
jtpereyda: Suggest something like:

```python
"Unable to reconnect to target: Reached threshold of {0} retries. Ending fuzzing.".format(self.restart_threshold)
```

boofuzz/sessions.py
```diff
@@ -1505,13 +1512,26 @@ def _open_connection_keep_trying(self, target):
...
+                        )
+                    if self.restart_timeout and time.time() > initial_time + self.restart_timeout:
+                        raise exception.BoofuzzTargetConnectionFailedError(
+                            "Unable to reconnect target: Reached restart timeout. Ending fuzzing."
```
jtpereyda: Suggest something like:
Another similar suggestion:
```python
"Unable to reconnect to target: Reached restart timeout of {0} seconds. Ending fuzzing.".format(self.restart_timeout)
```

boofuzz/sessions.py
```diff
@@ -1505,13 +1512,26 @@ def _open_connection_keep_trying(self, target):
...
+            unable_to_connect_count = 0
+            initial_time = time.time()
+
             while True:
                 try:
                     target.open()
                     break  # break if no exception
                 except exception.BoofuzzTargetConnectionFailedError:
                     self._fuzz_data_logger.log_info(constants.WARN_CONN_FAILED_TERMINAL)
+                    if unable_to_connect_count == self.restart_threshold:
```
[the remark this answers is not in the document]
jtpereyda: If we change to `>=` (which I also prefer for safety), we'll also need another check to verify the value was actually set (with `==`, a default value of `-1` will never trigger the the threshold). The first idea that comes to mind is making the default None and doing:
```python
if self.restart_threshold and ...
```
This will keep the default approach consistent with None as a default, and make the check more explicit than it was with `==`.

jtpereyda:
Thanks for the PR @dorpvom !

SR4ven:
Please also add a line to the CHANGELOG.rst

dorpvom:
Thanks again for the feedback. I incorporated the request changes to the code. Regarding the `_restart_target` suggestion, I'd have to dig a bit deeper in the code to know where to alternatively set up the checks.

SR4ven:
While testing the feature manually, I noticed that the exception message doesn't appear in the log.
This can be easily fixed by outputting a log message through the logger and then simply re-raising the current exception.

Log before:
```
[2020-02-01 18:37:55,491]     Info: Web interface can be found at http://localhost:26000
[2020-02-01 18:37:55,492] Test Case: 1: HTTP.path.1
[2020-02-01 18:37:55,492]     Info: Type: String. Default value: 'index.html'. Case 1 of 5899 overall.
[2020-02-01 18:37:55,492]     Info: Opening target connection (127.0.0.1:81)...
[2020-02-01 18:37:55,492]     Info: Cannot connect to target; retrying. Note: This likely indicates a failure caused by the previous test case, or a target that is slow to restart.
[2020-02-01 18:37:55,492]   Test Step: Restarting target
[2020-02-01 18:37:55,492]     Info: No reset handler available... sleeping for 5 seconds
[2020-02-01 18:38:00,496]     Info: Opening target connection (127.0.0.1:81)...
[2020-02-01 18:38:00,496]     Info: Cannot connect to target; retrying. Note: This likely indicates a failure caused by the previous test case, or a target that is slow to restart.
```

Log after fix:
```
[2020-02-01 18:55:55,308]     Info: Web interface can be found at http://localhost:26000
[2020-02-01 18:55:55,308] Test Case: 1: HTTP.path.1
[2020-02-01 18:55:55,308]     Info: Type: String. Default value: 'index.html'. Case 1 of 5899 overall.
[2020-02-01 18:55:55,309]     Info: Opening target connection (127.0.0.1:81)...
[2020-02-01 18:55:55,309]     Info: Cannot connect to target; retrying. Note: This likely indicates a failure caused by the previous test case, or a target that is slow to restart.
[2020-02-01 18:55:55,309]   Test Step: Restarting target
[2020-02-01 18:55:55,309]     Info: No reset handler available... sleeping for 5 seconds
[2020-02-01 18:56:00,316]     Info: Opening target connection (127.0.0.1:81)...
[2020-02-01 18:56:00,316]     Info: Unable to reconnect to target: Reached restart timeout of 5s. Ending fuzzing.
```

I'll commit my fix, please review.

PR #364, opened 2019-12-23
Move connections into own subpackage

SR4ven:
Clean up the main package folder by moving everything related to connections into a new subpackage.

Any thoughts or suggestions? @mistressofjellyfish


mistressofjellyfish:
Looks good to me, however I would add a word or two about the move to the documentation of the Connection classes. It should not matter unless someone has gone out of their way in importing bits and pieces instead of just using `from boofuzz import SerialConnection`, but you never know for sure.

SR4ven:
Good point, thanks for the hint.
I'll add a notification to the docs.

SR4ven:
@mistressofjellyfish I changed the `warnings` into `deprecated` directives according to the [sphinx docs](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#paragraph-level-markup).
The depreciation info is less obvious now, but I guess this will do too.

[Example](https://docs.python.org/3/library/socket.html#socket.ntohs) of how it will look.

mistressofjellyfish:
Hm, your call. I deliberately went with the warning because the deprecation itself wasn't obvious enough for me with regards to the socketconnection case. For the moving part, the `deprecated` directive will surely suffice.

SR4ven:
I get your point, the depreciation is indeed not very obvious. A note should fix this; it looks like the warning but is coloured grey instead of red.

Ready to merge from my side.

mistressofjellyfish:
Agreed, ready to merge from my side as well. I mean, we could always introduce some custom css to style the deprecated notices differntly, but I guess it isn't worth the hassle (and would not translate to other forms than html docs anyways).

SR4ven:
Thanks for the review @mistressofjellyfish! Merging now.

PR #373, opened 2020-01-20
Use merge=union to reduce CHANGELOG conflicts

SR4ven:
Whenever two PRs change the same line of the changelog and one gets merged, the other one is stuck in a merge conflict.

Setting the merge strategy to [union](https://git-scm.com/docs/gitattributes#Documentation/gitattributes.txt-union) will fix this. The only downside is that the order of the lines might not be chronological, which shouldn't be a problem as we usually reorder the lines before a release anyways.

Sadly this will only affect local merges as GitHub still doesn't read the `.gitattributes` file. Strangely GitLab does support this.
https://github.com/isaacs/github/issues/487
https://github.com/olivierlacan/keep-a-changelog/issues/56

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+CHANGELOG.rst merge=union


starek4:
As this change affects only `CHANGELOG.rst`, I believe we are free to merge it. The chronological order is not important to me in the changelog.

PR #374, opened 2020-01-27
remove redundant taskkill

cq674350529:
remove redundant taskkill command, as discussed in https://github.com/jtpereyda/boofuzz/pull/269


SR4ven:
LGTM, thank you for the fix @cq674350529!

PR #377, opened 2020-02-02
Unpin ldap3; optimise setup.py and manifest.in

SR4ven:
In #249 we fixed the ldap3 version to resolve a conflict when installing boofuzz.
ldap3 is not a direct requirement of boofuzz, but from Impacket which we use in the network_monitor.

Since the dependency conflict is now resolved, we can loosen up the requirement. https://github.com/SecureAuthCorp/impacket/issues/742
@jtpereyda any plans for removing the network monitor entirely? It's been marked as deprecated for at least 4 years now and only god knows if it still works in an up to date environment.

While looking at the setup.py I noticed some inconsistencies in the package excludes. We recently renamed the requests folder and this change was not applied here yet. I also cleaned the list up, removing entries, which don't have an affect with our current folder structure.
Additionally I used a more generic approach on the package_data definition and removed the license field, which is redundant as we specify it through a classifier later on. See [sample](https://github.com/pypa/sampleproject/blob/master/setup.py).

I verified that all required files are still in place for sdist and bdist_wheel. That's where I noticed that the travis.yml was included in the sdist, while all other CI related config like stickler and GitHub workflows are not. According to the [docs](https://packaging.python.org/glossary/), sdist is only supposed to hold essential source files required for installing. An exclude in the MANIFEST.in fixes this.
I took the opportunity to cleanup and restructure the manifest, sorting it by folder names.


jtpereyda:
@SR4ven LGTM; thank you!

jtpereyda:
@SR4ven Just got a test failure notification for Python 2.7. I re-ran it for now. I'm guessing it's related to the sporadic failures previously discussed rather than to this PR.

Regarding network monitor; for a while I thought nobody was using it, but I then met somebody who was. No plans to decommission it unless it's causing problems.

jtpereyda:
We could remove the "deprecated" indication; I think it still serves some purpose though I haven't used it myself in a while.

Issue #372, opened 2020-01-16
Docs: Install from source guide seems to be broken


mistressofjellyfish:
[The docs](https://boofuzz.readthedocs.io/en/stable/user/install.html#from-source) guide you to install from souce via `pip install -e .[dev] --user`, however, on newer `pip` and `setuptools` versions, at least in a virtual environment, this throws the following error:

```
  Running setup.py develop for boofuzz
    ERROR: Command errored out with exit status 1:
     command: /home/katharina/boofuzz/env/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/katharina/boofuzz/setup.py'"'"'; __file__='"'"'/home/katharina/boofuzz/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps --user --prefix=                                                               
         cwd: /home/katharina/boofuzz/                                                                             
    Complete output (6 lines):                                                                                     
    usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]                                          
       or: setup.py --help [cmd1 cmd2 ...]                                                                         
       or: setup.py --help-commands                                                                                
       or: setup.py cmd --help                                                                                     
                                                                                                                   
    error: option --user not recognized                                                                            
    ----------------------------------------    
```

I'd suggest to drop the current documentation and replace it with a guide that uses virtual environments, dropping the `--user` entirely.

SR4ven:
With the recent changes to pip, things seemed to get a little messy.

I've seen this problem before and it's related to PEP 518 build isolation in pip. The combination with the `--user` option and the editable install now results in the error shown here.
https://github.com/pypa/setuptools/issues/1405
https://github.com/jtpereyda/boofuzz/issues/326

If we omit the `--user` option, root rights will be required for installing on linux systems. I think this is because of the entrypoint script of boofuzz.

Honestly, I'm not exactly sure what method to advertise in the docs. A guide using virtualenv would at lease be mostly independent from the system and should have the best repeatability.

jtpereyda:
@SR4ven virtualenv method is probably fine and more repeatable anyway. 🤷‍♂

SR4ven:
[...]

PR #378, opened 2020-02-06
Updated install guide

mistressofjellyfish:
As by the discussion in #372, here is my proposal for the installation instructions.

I've also gone forward and introduced a few words about py3 and py2.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -4,24 +4,79 @@ Installing boofuzz
 Prerequisites
 -------------
 
-Boofuzz requires Python. Recommended installation requires ``pip``.
+Boofuzz requires Python 2.7 or ≥3.5. Recommended installation requires ``pip``.
+To ensure forward compatibility, python 3 is recommended. As a base
+requirement, the following packages are needed:
 
-Ubuntu: ``sudo apt-get install python-pip``
+Ubuntu/Debian
+  ``sudo apt-get install python3-pip python3-venv``
+OpenSuse
+  ``sudo zypper install python3-pip python3-virtualenv``
+CentOS
+  ``sudo yum install python36 python3-pip``
+
 
 Install
 -------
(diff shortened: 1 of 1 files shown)


INSTALL.rst
```diff
@@ -4,24 +4,79 @@ Installing boofuzz
...
 
-Boofuzz requires Python. Recommended installation requires ``pip``.
+Boofuzz requires Python 2.7 or ≥3.5. Recommended installation requires ``pip``.
+To ensure forward compatibility, python 3 is recommended. As a base
+requirement, the following packages are needed:
+
+Ubuntu/Debian
+  ``sudo apt-get install python3-pip python3-venv``
+OpenSuse
+  ``sudo zypper install python3-pip python3-virtualenv``
```
SR4ven: Is there no python3-venv package for OpenSuse? If not, the venv module might already be integrated in the standard Python installation. Ubuntu and Debian seem to be an exception here, outsourcing venv in a dedicated package.
https://stackoverflow.com/questions/41573587/what-is-the-difference-between-venv-pyvenv-pyenv-virtualenv-virtualenvwrappe

Also, venv is not available for Python 2, so one has to fallback to virtualenv.
The Flask [install docs](https://flask.palletsprojects.com/en/1.1.x/installation/#virtual-environments) handle this quite nicely.

mistressofjellyfish: I'll install an OpenSuse today and test it out - I haven't used anything Suse based for like 10 years, but I'll check.

mistressofjellyfish:
So after I've spent more time than I'm willing to admit on setting up a few virtual machines, I think I've got the prerequisites down now.

Turns out, if you want to install OpenSuse, you've better have some RAM (the installer used about 9 GiB).

jtpereyda:
Thanks @mistressofjellyfish and @SR4ven !

PR #392, opened 2020-02-20
Codecov

SR4ven:
I felt like it was worth a try proposing codecov.

I tried to keep the impact of this as low as possible, so there won't be any annoying comments on PRs, just two new checks indicating the coverage diff of the PR.
One check is for the whole project, the other for the patch. I wasn't quite sure about what the patch check is good for, but it's the default setting so I left it in for now. We can always disable it later if it's not needed.

If those checks fail too often, we can also adjust some parameters or make tests always succeed so that they only indicate the coverage.

[Example PR](https://github.com/SR4ven/boofuzz/pull/3)
[PR on Codecov](https://codecov.io/gh/SR4ven/boofuzz/pull/3)

TODO (if we want this feature):

- [ ] Activate Codecov on the project (already done? https://codecov.io/gh/jtpereyda/boofuzz)
- [ ] Activate the [app](https://github.com/marketplace/codecov) (Open Source version)
- [ ] Set `CODECOV_TOKEN` secret on GitHub from https://codecov.io/gh/jtpereyda/boofuzz/settings

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
           - os: windows-latest
             python-version: 3.5
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v1
       with:
@@ -41,3 +41,7 @@ jobs:
       env:
         TOXENV: ${{ format('py{0}-{1}', steps.py_tox_version.outputs.replaced, runner.os) }}
       run: tox --skip-missing-interpreters false
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v1
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,5 +9,5 @@ recursive-include request_definitions *.html *.py
 recursive-include unit_tests *.feature *.py
 recursive-include utils *.py
 
-exclude .readthedocs.yml .travis.yml
+exclude .readthedocs.yml .travis.yml codecov.yml
 prune docs/_build
--- a/README.rst
+++ b/README.rst
@@ -13,6 +13,8 @@ boofuzz: Network Protocol Fuzzing for Humans
     :target: https://gitter.im/jtpereyda/boofuzz
 .. image:: https://img.shields.io/badge/code%20style-black-000000.svg
     :target: https://github.com/psf/black
+.. image:: https://codecov.io/gh/jtpereyda/boofuzz/branch/master/graph/badge.svg
+    :target: https://codecov.io/gh/jtpereyda/boofuzz
 
 Boofuzz is a fork of and the successor to the venerable `Sulley`_ fuzzing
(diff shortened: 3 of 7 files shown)


jtpereyda:
Would this result in a test failure for the kind of PR where someone adds a small amount of new code to an untested section? While tests are generally helpful, I don't know if the pace of this project calls for mandatory unit testing yet. I wonder if we're still in the phase where emphasizing improvements over test coverage makes sense.

SR4ven:
Thanks for the feedback @jtpereyda, you definitely have a point there. We wouldn't want to hinder contributors adding new and untested code with failing tests.
However I think it's nice to know how much and what code is covered by our tests. Speaking for myself, this encourages me to write more tests for code that I add or change.

Two things that I've changed:
- I completely disabled the commit status which shows the patch coverage. I don't see a reason to have a separate check for that on GitHub. If anyone is interested in it, it can still be found on the codecov website.
- The commit status representing the coverage change for the whole project will never fail. It's purely informational.

In conclusion, there will be one check by codecov informing about the current coverage with a delta to the parent commit. This check will never fail. (It shouldn't at least, if it does I'll add some more settings.)

I wasn't sure whether we want to have a codecov badge showing the overall test coverage in our README.rst. Currently, we're at 58%. Thoughts?

jtpereyda:
@SR4ven Thanks. It would be nice if there was a sort of "warning" status; I haven't every seen anything like that on Github personally but it would be a perfect fit for something like this that is a helpful indicator but needs to be ignored often enough for now.

Hmm, would that badge with 58% make boo look good or bad? ;)

SR4ven:
Sadly GitHub doesn't support a warning status on checks. [Docs](https://developer.github.com/v3/checks/runs/)
That would have been really neat. As far as I know only Jenkins and GitLab support such features.

I was asking myself the same question about the badge :D
Some very brief research revealed that a common coverage goal is somewhat around 70-80%, so we're not too far off.
However, I'll leave that decision to you @jtpereyda.

Are the TODOs from this PR done?

jtpereyda:
@SR4ven Let's pass on the badge for now. I see coverage a secondary goal at this point in time since:
1. There's still a lot we'd want to change and improve in boofuzz. Higher churn increases the cost of maintaining tests.
2. A lot of code is heavy on network interactions, making the tests more difficult to write. We've seen some consequences already in the sporadically failing tests we've seen recently.

There are some aspects of functionality for which unit tests provide higher value. I would emphasize behavior related to test case generation -- e.g. a well-written unit tests could have caught #399. Unit tests are especially useful there because they protect correctness where the correctness failures are likely to go unnoticed by users.

jtpereyda:
So, removing the badge is the only todo then we can merge.

SR4ven:
@jtpereyda CI reports that the `CODECOV_TOKEN` secret is not set. https://github.com/jtpereyda/boofuzz/runs/475386830#step:8:36

These are the TODOs:

> - Activate Codecov on the project (already done? https://codecov.io/gh/jtpereyda/boofuzz)
> - Activate the app (Open Source version)
> - Set CODECOV_TOKEN secret on GitHub from https://codecov.io/gh/jtpereyda/boofuzz/settings

PR #397, opened 2020-02-24
Rework tox

SR4ven:
Changes and benefits:
- Most commands in tox are now platform independent.
  Only exception is a one-line bash script on Linux, which sets `cap_net_admin` and `cap_net_raw+eip` on the tox python interpreter. This is done every time the tox env is (re-)created and requires root privileges. Subsequent runs won't need any special privileges.
  With this change we don't run pytest with sudo, so there won't be any cache/log/output/data files owned by root in your project folder. On top it's more secure, as we only grant the capabilities that are really required.
- Docs and lint checks now run in their own environment. The advantage is that we don't run these checks multiple times for every python and platform combination. It is also possible to run all tox envs in parallel using `tox -p auto`, which usually seem to work fine as long as all tests pass. As CI runs all envs in parallel, the less duplicated tests mean a reduced testing time.
  Outsourcing the docs env allows us to run it in a separate test on CI. So it is now easier to tell what went wrong without having to look at the logs.
  The lint check is no longer being run on CI. We have Stickler for that. On local dev environments executing `tox` runs black and flake8 linters. We can add pytest later.
  **Docs and Lint envs require a Python 3 interpreter now.** Else the tests will be skipped. I hope that nobody is developing boofuzz on Python 2 anymore :D
- Updated docs including some indentation fixes

Note:
Docs are not tested on travis anymore. I was planning on removing Travis when releasing v0.2.0. Are you ok with that @jtpereyda?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,3 +45,16 @@ jobs:
       uses: codecov/codecov-action@v1
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
+  docs:
+    name: Build docs
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.x
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - name: Install pip and tox
+      run: python -m pip install --upgrade pip tox
+    - name: Run tox
+      run: tox --skip-missing-interpreters false -e docs
--- a/tox.ini
+++ b/tox.ini
@@ -1,36 +1,40 @@
 [tox]
 skip_missing_interpreters = True
-envlist = py{27,35,36,37,38}-{Linux,Windows,macOS}
+envlist = py{27,35,36,37,38}-{Linux,Windows,macOS}, docs, lint
 
 [testenv]
-whitelist_externals =
-    sudo
-    make
-    cmd
 platform =
     Windows: win32
     Linux: linux
     macOS: darwin
+whitelist_externals = bash
 extras = dev
-install_command =
-    {envpython} -m pip install {opts} {packages}
-commands_pre =
-    {envpython} -m pip check
-    # Stop the test if there are Python syntax errors or undefined names.
+commands =
+    pip check
+    Linux: bash -ec "if ! getcap $(realpath {envpython}) | grep -q cap_net_admin,cap_net_raw+eip; then sudo setcap cap_net_admin,cap_net_raw+eip $(realpath {envpython}); fi"
+    pytest --cov --cov-report=xml
+    check-manifest
+commands_post =
+    - python -c "import shutil; shutil.rmtree('./boofuzz-results/')"
+
+[testenv:docs]
+basepython = python3
+extras = docs
+commands =
+    sphinx-build -TW -b dummy ./docs ./docs/_build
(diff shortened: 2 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           - os: windows-latest
             python-version: 3.5
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
```
jtpereyda: What does this do in a nutshell?

tox.ini
```diff
@@ -1,36 +1,41 @@
...
+whitelist_externals = bash
 extras = dev
-install_command =
-    {envpython} -m pip install {opts} {packages}
-commands_pre =
-    {envpython} -m pip check
-    # Stop the test if there are Python syntax errors or undefined names.
+commands =
+    pip check
+    Linux: bash -ec "if ! getcap $(realpath {envpython}) | grep -q cap_net_admin,cap_net_raw+eip; then sudo setcap cap_net_admin,cap_net_raw+eip $(realpath {envpython}); fi"
```
jtpereyda: This line is new, right? What's the gist of it?

.github/workflows/test.yml
SR4ven: Format is `user/repo@tag`.
So we're using the checkout action at tag v2 now as recommended [here](https://github.com/marketplace/actions/checkout).
It doesn't seem to be a good practice to use @master as that might result in some instabilities/inconsistencies.

tox.ini
SR4ven: From the PR comment:

> Only exception is a one-line bash script on Linux, which sets `cap_net_admin` and `cap_net_raw+eip` on the tox python interpreter. This is done every time the tox env is (re-)created and requires root privileges. Subsequent runs won't need any special privileges.
With this change we don't run pytest with sudo, so there won't be any cache/log/output/data files owned by root in your project folder. On top it's more secure, as we only grant the capabilities that are really required.

It checks if the required capabilities needed for raw socket operations are set with `getcap` and sets them with `setcap` if they aren't already set.
The capabilities are only applied to the tox env interpreter, not the global one.

jtpereyda: Ah that makes sense now. I like the security hygiene!

jtpereyda:
> Docs are not tested on travis anymore. I was planning on removing Travis when releasing v0.2.0. Are you ok with that @jtpereyda?

yep that sounds great -- there's no loss of test coverage?

SR4ven:
No, currently all tests on Travis run on Linux and those are the same we run on GitHub actions.
Test coverage is way higher on Actions as we also test on Windows.

At the moment Travis is just a redundancy in case something is wrong with GitHub Actions, but we're using it for quite a while now and everything seems fine.

jtpereyda:
OK, sounds like we can shut down Travis anytime with limited risk.

SR4ven:
> OK, sounds like we can shut down Travis anytime with limited risk.

Agreed. I'll open a PR for this later.
The only thing that hasn't been fully tested is the auto-deploy to pypi when creating a release on GitHub.
But as I only copy pasted the code for that from official sources, I'm confident that we won't run into any issues there.
Until we verified that this works as expected I wouldn't entirely remove Travis from the project.

PR #399, opened 2020-02-26
Fixed size of s_size block when output is ascii

mshaneck:
If the output format of the s_size block is ascii, the __len__ function still returns 4 bytes. This commit takes into account the output format of the s_size block. Related to issue https://github.com/jtpereyda/boofuzz/issues/197 and pull request https://github.com/jtpereyda/boofuzz/pull/198

### DIFF
--- a/boofuzz/blocks/size.py
+++ b/boofuzz/blocks/size.py
@@ -220,10 +220,10 @@ class Size(IFuzzable):
         return "<%s %s>" % (self.__class__.__name__, self._name)
 
     def __len__(self):
-        if self.format == "binary":
-            return self.length
+        if self.format == "ascii":
+            return len(str(self._calculated_length()))
         else:
-            return len(str(len(self.request.names[self.block_name])))
+            return self.length
 
     def __bool__(self):
         """
--- a/boofuzz/blocks/size.py
+++ b/boofuzz/blocks/size.py
@@ -220,7 +220,10 @@ class Size(IFuzzable):
         return "<%s %s>" % (self.__class__.__name__, self._name)
 
     def __len__(self):
-        return self.length
+        if self.format == "binary":
+            return self.length
+        else:
+            return len(str(len(self.request.names[self.block_name])))
 
     def __bool__(self):
         """
--- a/boofuzz/blocks/size.py
+++ b/boofuzz/blocks/size.py
@@ -220,10 +220,7 @@ class Size(IFuzzable):
         return "<%s %s>" % (self.__class__.__name__, self._name)
 
     def __len__(self):
-        if self.format == "ascii":
-            return len(str(self._calculated_length()))
-        else:
-            return self.length
+        return len(self._render())
 
     def __bool__(self):
         """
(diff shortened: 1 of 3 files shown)


boofuzz/blocks/size.py
```diff
@@ -220,7 +220,10 @@ def __repr__(self):
         return "<%s %s>" % (self.__class__.__name__, self._name)
 
     def __len__(self):
-        return self.length
+        if self.format == "binary":
```
SR4ven: I feel like returning `self.length` in the "default" else case might be better, as the binary format is the blocks default too.

Also, when someone defines a none existing format, we would return `self.length`. Not that we'd have to cover this edge case, but I feel like this is more straight forward than the ascii format.

However this is all about personal preference I guess, so I'm fine either way.

mshaneck: Yes, this makes sense. Although I think to keep it consistent, this same change should be made to bit_field. What do you think?

SR4ven: Good point, we should indeed change it at both points consistently.

mshaneck:
Pull request is updated, as discussed.

boofuzz/blocks/size.py
```diff
@@ -220,7 +220,10 @@ def __repr__(self):
...
+            return self.length
+        else:
+            return len(str(len(self.request.names[self.block_name])))
```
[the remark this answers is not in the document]
mshaneck: Agreed.

mshaneck:
No problem! Thanks for letting me help out.

boofuzz/blocks/size.py
```diff
@@ -220,7 +220,10 @@ def __repr__(self):
         return "<%s %s>" % (self.__class__.__name__, self._name)
 
     def __len__(self):
-        return self.length
+        if self.format == "ascii":
+            return len(str(self._calculated_length()))
+        else:
+            return self.length
```
jtpereyda: Spring-boarding off of @SR4ven 's suggestion -- Would this work too? If so, it would avoid some code reuse:
```python
    def __len__(self):
        return len(self._render())
```

The code in _render() is very slightly different -- I'm not sure if it's more accurate, but at least the logic would be in one place.

mshaneck: Yes, that looks like a good idea. It works just as good in my test cases. I updated the pull request.

SR4ven: Good catch @jtpereyda!

The same code seems to work for `bit_field`
```python
    def __len__(self):
        return len(self._render(self._value))
```

If someone else can verify that this still does what it's supposed to, I suggest applying this change.

jtpereyda: @mshaneck are you able to apply/test that change in bit_field too?

mshaneck: Yeah. Looks good. I updated the pull request.

SR4ven:
I ran some tests and everything works as expected.

Thanks again @mshaneck!

PR #401, opened 2020-02-28
Fixes for s_bytes

### LINKED ISSUE #400: total_num_mutations wrong with s_bytes?
when assign the size for s_bytes, total_num_mutations is not correct?

a simple test:

```py
from boofuzz import *


def main():
    session = Session(
        target=Target(
            connection=TCPSocketConnection("127.0.0.1", 80)
        ),
    )

    s_initialize(name="Test")
    s_bytes("\x00\x00",size=2)
    session.connect(s_get("Test"))
    session.fuzz()

if __name__ == "__main__":
    main()
```

The output:
```
[2020-02-28 14:28:32,169] Test Case: 21: Test.no-name.21
[2020-02-28 14:28:32,169]     Info: Type: Bytes. Default value: '\x00\x00'. Case 21 of 65 overall.
[2020-02-28 14:28:32,169]     Info: Opening target connection (127.0.0.1:80)...
[2020-02-28 14:28:32,169]     Info: Connection opened.
[2020-02-28 14:28:32,169]   Test Step: Fuzzing Node 'Test'
[2020-02-28 14:28:32,169]     Info: Sending 2 bytes...
[2020-02-28 14:28:32,169]     Transmitted 2 bytes: ff ff '\xff\xff'
[2020-02-28 14:28:32,170]     Info: Closing target connection...
[2020-02-28 14:28:32,170]     Info: Connection closed.

Fuzzing session completed. Keeping webinterface up on localhost:26000 
Press ENTER to close webinterface

```

The total_num_mutations is 65, but boofuzz finish fuzz at case 21.

Is this a bug?

SR4ven:
- Fixes #400 
- Properly includes values that were left out before. Renamed class var `_fuzz_library` to `_fuzz_values` as it previously seemed to collide with some other var with the same name. I'll take a close look at this when I have a proper debugger at my hand.

TODO:

- [ ] Make use of `size` and `max_len` more clear. It might be even better to have `size` and `truncate` arguments instead.
  Setting `size` currently results in too long values being dropped.
  If `max_len` is set and `size` is none, the values will be truncated.
  Setting both arguments has the same effect as setting only `size`.

  I suggest dropping `max_len` in favour of `truncate` to make using this primitive more straight forward.
  Use cases would be the following:
  - Default `truncate=False`. Setting just `size` will drop too long test cases.
  - Setting `size` and `truncate=True` will render all test cases but truncate the values.
  - Not setting `size` will render all test cases with dynamic length.
- [ ] Update docs

### DIFF
--- a/boofuzz/primitives/bytes.py
+++ b/boofuzz/primitives/bytes.py
@@ -4,7 +4,7 @@ from .. import helpers
 
 class Bytes(BasePrimitive):
     # This binary strings will always included as testcases.
-    _fuzz_library = [
+    _fuzz_values = [
         b"",
         b"\x00",
         b"\xFF",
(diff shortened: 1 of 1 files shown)


keisentraut:
I'm sorry for this, did not test it properly before creating my push request #302 . I assume you are already fixing it, if not, I can do it, too.

I remember being confused about `size` and `max_len`, too. If you change this, I would recommend changing it for the `s_string` primitive, too.

SR4ven:
Thanks for the info @keisentraut!
As far as I can tell, the main issue is already fixed by this PR. I haven't changed anything about arguments yet, because I was waiting for a second opinion.
Good point about `s_string`, maybe we could change the arguments consistently in another PR instead.

However, you're very welcome to review my changes!
I still don't get why `_fuzz_library` is an empty list in the original version. Any ideas?
I'll try to get behind this with a debugger.

keisentraut:
In the original `string` primitive, I don't understand why `fuzz_library` is set to an empty list. Instead, they create `self.fuzz_library` member in every new `string` primitive. I think you can simply remove the empty  `fuzz_library` from `s_string`. In my `bytes` primitive, those values do not change at all and therefore, we only have it once as static members.

SR4ven:
Ok, so the debugger did help a lot with this.

Turns out that `_fuzz_library` is overwritten in the constructor of `BasePrimitive`
https://github.com/jtpereyda/boofuzz/blob/0ed135ae9453a6aab4c5bdda12e9f16eccce5062/boofuzz/primitives/base_primitive.py#L35

Moving the definition of `_fuzz_library` into the constructor of `Bytes` fixed this.

Running some tests on the num_mutations still returns the wrong number.
Any ideas @keisentraut? If you prefer opening your own PR, that'd be fine too.

keisentraut:
I remembered how I wanted `s_bytes` to behave:

- When `size` is set, only testcases with the exact size should be written out. This is very useful for fixed-length protocol fields. For instance, it's not useful to output five bytes when your target protocol expects only exactly three zero bytes and some other primitives (e.g. integers) afterwards.
- When `max_len` is set, only testcases with an equal or shorter amount of bytes should be outputted. Useful for protocols where a maximum length is checked.

You can set both `size` and `max_len`, but it's not useful: If you set `size<=max_len` then `max_len` is useless. Otherwise, you will have 0 testcases which is not useful, too. I will fix it, but currently I have limited time, so if you want to have it fixed fast, then please continue with this branch ;)

SR4ven:
Should be fixed by #422

keisentraut:
I think the whole string/bytes API needs a rewrite. Obviously, ```size``` and ```max_len``` are not self-explanatory. Therefore I created an issue with a suggestion and follow-up discussions on how it could look like.

PR #402, opened 2020-03-04
Disable Travis deploy stage

SR4ven:
Preparation for v0.2.0.

Whenever we push our next release, we don't want two CI services deploying at the same time.
We could completely remove Travis at this point, but as our new deploy Action hasn't been tested yet, I'd say keep Travis around until we're certain that everything works.

I switched out the status badge and updated the docs.
Currently, there is a problem with GitHub Action badge links and the Sphinx linkcheck, that's why I added a temporary ignore list until https://github.com/sphinx-doc/sphinx/issues/7247 is fixed.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,7 +8,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


SR4ven:
True that. Good thing we have the auto fixer that takes care of this. I keep forgetting to run to run the linters... maybe I should setup some IDE integration or pre commit hook.

PR #403, opened 2020-03-04
Fix coverage scope

SR4ven:
I noticed that we currently generate a coverage report over the whole project folder.
That includes files out of the unit test scope like process_monitor.py, setup.py and the unit_tests folder itself. I figured that this doesn't make much sense, even though it boosts our coverage a little.

With this PR, only the boofuzz folder is used for coverage data.
That gives us much more accurate reports.


jtpereyda:
Great! Thanks for getting the coverage set up.

Issue #379, opened 2020-02-07
Consistency: Remove some parameters from Session class


mistressofjellyfish:
So while I was debugging another issue, I came along the enormous list of parameters in the Session class. I feel that some of them should be moved to more appropriate locations, and I'd argue that this should be a breaking change. Benefits: better documentation, easier to understand.

In particular:
- `ignore_connection_reset`, `ignore_connection_aborted`, `ignore_connection_issues_when_sending_fuzz_data`, should be moved into the Connection classes. They are only valid on Connections that *may* abort or reset anyways and are misleading - it's unclear how they apply to, say, a serial connection (where all of them basically mean the same thing).
- Likewise, `ignore_connection_ssl_errors` should be a property of SSLSocketConnection
- I'm still unsure about `reuse_target_connection`, which - conceptually - belongs IMO into the target class, but from a first glance it seems to be quite hard to move it with how the Session class currently works.

That, combined with some renaming, would clear up the **massive** constructor of the Session class.

As far as I can tell, all these parameters only affect the log level anyways, which lead to very messy code when handling these (boofuzz/sessions.py:1109ff):

```python
        except exception.BoofuzzTargetConnectionReset:
            # TODO: Switch _ignore_connection_reset for _ignore_transmission_error, or provide retry mechanism
            if self._ignore_connection_reset:
                self._fuzz_data_logger.log_info(constants.ERR_CONN_RESET)
            else:
                self._fuzz_data_logger.log_fail(constants.ERR_CONN_RESET)
```

This is repeated a few times, and only gets worse the more connection types boofuzz supports. If the fuzz data logger is passed on to a class, the interface could be extended to set the appropriate method:

```python
class SSLSocketConnection(tcp_socket_connection.TCPSocketConnection):
    def __init__(
        self, host, port, send_timeout=5.0, recv_timeout=5.0, server=False, sslcontext=None, server_hostname=None, ignore_ssl_errors=False
    ):
        super(SSLSocketConnection, self).__init__(host, port, send_timeout, recv_timeout, server)
        self._ssl_error_logger = None
        self._ignore_ssl_errors = ignore_ssl_errors

    def set_logger(self, logger):
        self._ssl_error_logger = logger.log_fail if self._ignore_ssl_erros else logger.log_info
```

Or something like this. I haven't put much thought about that into it and at the moment I can't really justify the time I would spend implementing this (I don't have a business case), but if someone wants to go ahead and move some stuff around, I'll happily provide my thoughts.

[...]

PR #405, opened 2020-03-06
Cleanup deprecated code

SR4ven:
Removing some deprecated session options that have no effect or even raise an exception will help to shrink the arguments list in the docs a little at least.
Also a small step towards #379.

Same applies to FuzzLoggerFile which doesn't implement all log functions and will prevent boofuzz from running if used. It was replaced by FuzzLoggerText which can also output to a file.


jtpereyda:
Thanks @SR4ven ! LGTM. You said we use "Squash and merge" typically for merges?

SR4ven:
At least that's what I did for the last few merges as it keeps the commit history clean and includes the PR number in the commit message. I usually remove all lines from the extended commit message which would usually contain all commits that were squashed. If there are any co-authors (two people working on the PR) then I leave that line in; if it's Stickler remove it.

That way we end up with one commit per PR, easy traceback, reverting and a reasonable number of commits.
However, if you want to stick with a merge commit, that's fine by me.

PR #406, opened 2020-03-29
Change deprecation into future warnings

SR4ven:
Using `FutureWarning` instead of `DeprecationWarning` seems to make more sense as they aren't suppressed by default and are displayed to the user.


jtpereyda:
Makes sense; thank you!

PR #408, opened 2020-03-31
Implement Session graph rendering

SR4ven:
This all started when I asked myself why `pgraph.Node` was basing `IFuzzable`. If anything, that should be the other way around. Now I ended up fixing all `render_graph_*` functions.
It's now possible to use those functions to render data that can be used to generate a picture of the current session graph.

Example usage:
```python
with open('somefile.png', 'wb') as file:
    file.write(session.render_graph_graphviz().create_png())
```

The only negative point about all this is, that everything that bases `IFuzzable` now has quite a few more members as they now also base `Node`. This makes things a little messy in the debugger and will increase memory overhead a little but nothing tragic as far as I can tell.

I also took the chance to remove the `DocStringInheritor` from `IFuzzable`. I don't see why we need it, Shpinx already has that feature. See [this](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_inherit_docstrings).


SR4ven:
Is the graph rendering still a requested feature?
If so I'll rebase the PR, else we can close it.

jtpereyda:
@SR4ven I don't see any reason to drop it, do you? My bad for never reviewing.

SR4ven:
No problem, I'll get it up to date when I find some free time.

jtpereyda:
@SR4ven What do you think of making Request inherit from Node (instead of Fuzzable inheriting from Node)? That gives us a multiple inheritance situation. But, it seems that Requests are the only Fuzzables for which the Node behavior matters. I tried it out in code and it seems to function OK.

SR4ven:
Good point, that makes sense. I'll change it.

boofuzz/fuzzable.py
```diff
@@ -1,12 +1,12 @@
+from future.builtins import object
 from future.moves import itertools
 
-from .mutation import Mutation
+from boofuzz.mutation import Mutation
```
jtpereyda: Can we keep this as `from .mutation import Mutation` for consistency? (I'm not sure if there's any semantic difference)

jtpereyda (approved):
LGTM

SR4ven:
Thanks for reviewing @jtpereyda!

PR #409, opened 2020-03-31
Properly close sockets on PED-RPC Server

SR4ven:
After getting some random failures for the monitor tests and looking for the cause of them I noticed that the PED-RPC Server doesn't close any sockets when being stopped. The tests then fail because of [Errno 48] Address already in use, as PED-RPC Servers are frequently being set up and torn down.

I made an attempt in fixing this with a stop function that closes all sockets. In fact it even required a shutdown for the sockets as close wasn't fast enough. The stop function also halts the `server_forever` function.
It's not possible to start a PED-RPC Server again once it's been stopped as the socket open logic is included in the constructor. I don't think this is a use case anyways and simply creating a new object works fine.

I changed a lot of the monitor test code, would be nice if you could check if it still tests what was intended @mistressofjellyfish.

The test for Python 2.7 on Windows will most likely fail and right now I have no idea why. I'll take a look into that later, but if anyone can provide a hint on why it fails, it might save me some time.


mistressofjellyfish:
Looks good to me. All the tests really do is test if a server can connect and the option persistence logic. At least the latter is a direct consequence of creating a new connection for every command.

*sigh*

jtpereyda:
Makes sense. Mysterious test failure though. I saw this excerpt in the logs:
```
self = <multiprocessing.forking.ForkingPickler instance at 0x0000000015ABF3C8>
obj = <built-in method recv_into of _socket.socket object at 0x0000000015ABF0B0>
name = 'recv_into', pack = <built-in function pack>

    def save_global(self, obj, name=None, pack=struct.pack):
        write = self.write
        memo = self.memo
    
        if name is None:
            name = obj.__name__
    
        module = getattr(obj, "__module__", None)
        if module is None:
            module = whichmodule(obj, name)
    
        try:
            __import__(module)
            mod = sys.modules[module]
            klass = getattr(mod, name)
        except (ImportError, KeyError, AttributeError):
            raise PicklingError(
                "Can't pickle %r: it's not found as %s.%s" %
>               (obj, module, name))
E           PicklingError: Can't pickle <built-in method recv_into of _socket.socket object at 0x0000000015ABF0B0>: it's not found as __main__.recv_into
```

And the test failure output does seem to point indirectly to `Server.serve_forever` (pedrpc.py), which does engage in some pickling. I still don't understand why this would happen though. Bizarre.

You might try walking back your changes incrementally to see which bit causes the exception to show up. An alternative is to drop the naive pickling and fix Ped-RPC. :P

SR4ven:
This error has something to do with the OS specific start method for processes.
I'm not that deeply into threading so I didn't trace the error back. However, I'm quite sure that the error is caused by making some local variables class members.

Apparently Python 2.7 isn't able to handle the pickling on Windows. This could perhaps be fixed by using the threading module instead, but that rises other issues. We wouldn't be able to terminate threads or read their exit code. At leat not without introducing other workarounds.

The main issue here seems to be that I directly access the `stop()` method of PED-RPC server from the main process. This requires the processes to share some memory and I guess that's what causes the problem.

I'll try to figure a way out to work around this. I've got some ideas in my head, but they would change the test behaviour slightly.

Some refs:
https://github.com/ralph-group/pymeasure/issues/12
https://stackoverflow.com/questions/9670926/multiprocessing-on-windows-breaks

SR4ven:
I didn't find a quick way to fix the tests on Python2 Windows so I simply disabled them. We still run the tests on Python2 Linux so this shouldn't be too much of a problem. (Honestly, I'm hoping to drop Python2 soon)

Alternatively we leave everything as it is, but to be honest I'm really surprised that we haven't had any random CI failures yet. Tests kept failing regularly on my system.

PR #412, opened 2020-04-07
s_bit_field with list of values (as a fuzz library)

trefax:
If s_bit_field is created with a list of values to be used as the fuzz library, an error in bit_field.py occurs:

```
  File "...\boofuzz\primitives\bit_field.py", line 38, in <lambda>
    return "".join(map(lambda x: str((number >> x) & 1), range(bit_width - 1, -1, -1)))
TypeError: unsupported operand type(s) for >>: 'list' and 'int'
```

Choosing a single value, e.g. the first value in the list, resolves the issue.


jtpereyda:
Thanks @trefax!

PR #413, opened 2020-04-12
Mutations iterator

jtpereyda:
Lots of updates, most notably:
1. Refactor fuzzing engine to use iterators/generators instead of a weird stateful pseudo-iterator.
2. Boofuzz now separates mutation and encoding. The mutations step generates new Mutation objects. The encoding step takes a protocol definition and a mutation and generates a message or messages. This distinction already existed in the implementation of some fuzz primitives, but it is now formalized. The distinction will enable multiple mutations in the future, and can even enable coverage-guided mutation stacking.

I haven't fixed unit tests yet, but I wanted to get some initial feedback before doing that footwork.

Rough summary of changes:
1. All fuzz elements now inherit from Fuzzable (the new IFuzzable). Its most important methods are `mutations()` and `encode()`. For an example of how simple this can make fuzzers, see static.py:
    ```python
    class Static(Fuzzable):
        pass
    ```
    See string.py for a less trivial example.
2. Boofuzz uses FuzzableWrapper to attach metadata like name and context for each element. Fuzzable itself is nearly stateless.
3. Callbacks now have a TestCaseContext object to which one can save data to be used later in the test case, for example...
4. `TestCaseSessionReference` can be passed as a default value in a protocol definition. The name it references must have been saved by the time that message in the protocol is reached.
5. `FuzzableBlock` is a subclass of Fuzzable and superclass for Block types.
6. Added Aligned, a FuzzableBlock that aligns its contents to a certain number of bytes.
7. `Mutation`, `MutationContext`, and `TestCaseContext` support the new system.
8. Various web interface fixes.
9. A few `Session` simplifications (more deletions than insertsion! :D).

Some future improvements made easier:
1. FuzzableWrapper is a starting point for implementing odd-bit-length types. Since it wraps most calls to mutations, it can handle some special bits-bytes type yielded by a special primitive.
2. New Fuzzable types are simpler. 
3. This is a framework for moving away from the static-functions-for-protocol-definition thing.
4. Coverage feedback now possible.

Some things that are better but not ideal:
1. The Fuzzable-FuzzableWrapper thing is a little weird. It could be collapsed into one class, but then it becomes more cluttered for people making subclasses.
2. Request is still a very special case, but that's partly OK. It has the job of binding the stateless world of Fuzzables with the stateful world of Session.


boofuzz/__init__.py
```diff
@@ -643,8 +658,9 @@ def s_mirror(primitive_name, name=None):
     if primitive_name not in blocks.CURRENT.names:
         raise exception.SullyRuntimeError("CAN NOT ADD A MIRROR FOR A NON-EXIST PRIMITIVE CURRENTLY")
 
-    mirror = primitives.Mirror(primitive_name, blocks.CURRENT, name)
+    mirror = primitives.Mirror(primitive_name, blocks.CURRENT)
```
cq674350529, line 661: any special for s_mirror?

boofuzz/__init__.py
```diff
@@ -643,8 +658,9 @@ def s_mirror(primitive_name, name=None):
...
     blocks.CURRENT.push(mirror)
+    blocks.CURRENT.push(FuzzableWrapper(fuzz_object=mirror, name=name, fuzzable=True, default_value=None))
```
cq674350529, line 663: Any special for `s_mirror`? There is an extra statement `blocks.CURRENT.push(mirror)`.

jtpereyda: Good catch, `blocks.CURRENT.push(mirror)` should go.

boofuzz/fuzzable.py
```diff
@@ -0,0 +1,135 @@
...
+        Values are either plain values or callable functions that take a "default value" and mutate it. Functions are
+        used when the default or "normal" value influences the fuzzed value. Functions are used because the "normal"
+        value is sometimes dynamic and not known at the time of generation.
+
+        Each mutation should be a pre-rendered value. That is, it must be suitable to pass to encode().
+
+        Default: Empty iterator.
+        """
+        return
+        yield
```
cq674350529, line 63: redudant statement `return`? Or a magic trick?

jtpereyda: Empty generator trick. See https://stackoverflow.com/questions/13243766/python-empty-generator-function

Now that I read that answer again, I think `return iter(())` is more intuitive.

boofuzz/blocks/repeat.py
```diff
@@ -91,57 +84,47 @@ def fuzzable(self):
...
 
         @rtype:  int
         @return: Number of mutated forms this primitive can take.
+        :param default_value:
         """
 
         return len(self._fuzz_library)
 
+    def encode(self, value, mutation_context):
+        return value * child_data
```
cq674350529, line 112: Good job for the mutation refactor. Currently I don't know much about the minds behind it. But it should be cool.

Just according to the code, the `child_data` is missing, which is placed in another function. Something maybe like this:

```
def encode(self, value, mutation_context):
    child_data = self._get_child_data(mutation_context=mutation_context)
    return value * child
```

Also, the 1st arg of `_get_child_data()` in this file should be mutation_context instead of mutation.

jtpereyda: Yep, that's about what it should be. Now that I look at it, Repeat has a few changes that should be made. Repeat wasn't used in my ad hoc testing, so it probably is non-functional right now. I'll take this as a reminder to revisit the whole class.

boofuzz/fuzzable.py
cq674350529: Wow, thanks, got a new trick :)

boofuzz/blocks/repeat.py
jtpereyda: E.g. `Repeat.render_mutated()` can be deleted. `Repeat.render()` will probably get folded into `Repeat._get_child_data()`.

boofuzz/primitives/string.py
```diff
@@ -258,61 +248,32 @@ def add_long_strings(self, sequence):
...
-    def num_mutations(self):
+    def num_mutations(self, default_value):
         """
         Calculate and return the total number of mutations for this individual primitive.
 
         @rtype:  int
         @return: Number of mutated forms this primitive can take
+        :param default_value:
         """
         return len(self._fuzz_library) + len(self.this_library)
```
cq674350529, line 279: As to the `default_value` parameter, it's not used yet. According to the comments in `Fuzzable` class, it's used if exhausting the mutations() is not possible. So maybe check if it's not `None`, then just return it as the number of mutations.

boofuzz/blocks/repeat.py
cq674350529: Yeah, the `Repeat.render_mutated()` calls `self.encode()` to return a encoded value, which is what `self.encode()` does.

boofuzz/primitives/string.py
jtpereyda: The default_value parameter isn't used in this case because it doesn't impact the number of mutations for a String. Although it could in the future.

See bytes.py for an example of num_mutations that uses the default value.

But here in string.py, it needs to be here to match the method signature provided by Fuzzable.

cq674350529: Got it, thanks. I regarded the `default_value` as the default number of mutations, actually it's the default value of the `fuzz_object`. Since this attribute is seperated from fuzz_object itself, there is no explict way to get it from the `fuzz_object`.

boofuzz/__init__.py
```diff
@@ -561,7 +576,7 @@ def s_group(name, values, default_value=None):
     """
 
     group = primitives.Group(name, values, default_value)
-    blocks.CURRENT.push(group)
+    blocks.CURRENT.push(FuzzableWrapper(fuzz_object=group, name=name, fuzzable=False, default_value=default_value))
```
cq674350529, line 579: The `default_value` passed in `Group()` can be removed too, since the `Group._value` and `Group._original_value` are not used anymore.

boofuzz/primitives/group.py
```diff
@@ -32,41 +33,20 @@ def __init__(self, name, values, default_value=None):
         for val in self.values:
             assert isinstance(val, (six.binary_type, six.string_types)), "Value list may only contain string/byte types"
 
-    @property
-    def name(self):
-        return self._name
+    def encode(self, value, mutation_context):
+        return self._render(value)
```
cq674350529, line 37: `self._render()` seems not be implementated. I did't find it in `Group` or its parent.

jtpereyda: Group is another one I didn't functional test; I'll come back to it when I update the unit tests.

boofuzz/__init__.py
```diff
@@ -365,14 +365,14 @@ def s_block_end(name=None):
...
-        ipv4_src_block_name=None,
-        ipv4_dst_block_name=None,
+    block_name,
+    algorithm="crc32",
+    length=0,
+    endian=LITTLE_ENDIAN,
+    fuzzable=True,
+    name=None,
+    ipv4_src_block_name=None,
+    ipv4_dst_block_name=None,
```
jtpereyda: @SR4ven This is one place where the PyCharm "Reformat File" behavior differs from Stickler-CI's. Reformat File makes all these lines double-indented. Sticker-CI prefers a single indentation. Doesn't really matter, but makes it easy to end up with a lot of stickler-ci changes.

SR4ven: Usually I only use PyCharm to optimize imports and then `black .` to reformat.
Running `tox` will also fail at the lint check if black finds something to reformat.

I think it's possible to add a PyCharm project config file to git which might allow us to unify differences between PyCharm and black. I'm not sure how many differences there are and if it's worth the time.

The specific issue you mentioned can be fixed at Settings > Editor > Code Style > Python > Continuation Indent. Another option would be to ignore the PyCharm reformatting and use the [black integration](https://black.readthedocs.io/en/stable/editor_integration.html) instead.

jtpereyda: Thanks! I changed that PyCharm setting for now.

boofuzz/primitives/bytes.py
```diff
@@ -1,3 +1,9 @@
+import functools
+import itertools
+import operator
+
+from funcy import compose
```
SR4ven, line 5: Package not listed in setup.py

boofuzz/primitives/bytes.py
```diff
@@ -91,136 +97,119 @@ class Bytes(BasePrimitive):
...
-                # stage 1b: replace with this_library items
-                alreadyDone = len(self._fuzz_library)
-                self._value = self.this_library[self._mutant_index - alreadyDone]
-            elif self._mutant_index < len(self._fuzz_library) + len(self.this_library) + len(self._magic_debug_values):
-                # stage 1c: replace with _magic_debug_value items
-                alreadyDone = len(self._fuzz_library) + len(self.this_library)
-                self._value = self._magic_debug_values[self._mutant_index - alreadyDone]
+    def _adjust_mutation_for_size(self, fuzz_value):
+        if self.size is not None:
+            if len(fuzz_value) > self.size:
```
SR4ven, line 134: Getting the following error
```python
  File "/home/dev/Documents/boofuzz_requests/dev.py", line 111, in <module>
    fuzz()
  File "/home/dev/Documents/boofuzz_requests/dev.py", line 107, in fuzz
    sess.fuzz()
  File "/home/dev/PycharmProjects/boofuzz/boofuzz/sessions.py", line 692, in fuzz
    self._main_fuzz_loop((m for path in self._iterate_protocol() for m in self._iterate_single_node(path)))
  File "/home/dev/PycharmProjects/boofuzz/boofuzz/sessions.py", line 786, in _main_fuzz_loop
    for mutation_context in fuzz_case_iterator:
  File "/home/dev/PycharmProjects/boofuzz/boofuzz/sessions.py", line 692, in <genexpr>
    self._main_fuzz_loop((m for path in self._iterate_protocol() for m in self._iterate_single_node(path)))
  File "/home/dev/PycharmProjects/boofuzz/boofuzz/sessions.py", line 1367, in _iterate_single_node
    for mutation in self.fuzz_node.mutations():
  File "/home/dev/PycharmProjects/boofuzz/boofuzz/fuzzable_block.py", line 21, in mutations
    for mutation in item.mutations():
  File "/home/dev/PycharmProjects/boofuzz/boofuzz/fuzzable_wrapper.py", line 114, in mutations
    for value in itertools.chain(self._fuzz_object.mutations(), self._fuzz_values):
  File "/home/dev/PycharmProjects/boofuzz/boofuzz/fuzzable_block.py", line 21, in mutations
    for mutation in item.mutations():
  File "/home/dev/PycharmProjects/boofuzz/boofuzz/fuzzable_wrapper.py", line 114, in mutations
    for value in itertools.chain(self._fuzz_object.mutations(), self._fuzz_values):
  File "/home/dev/PycharmProjects/boofuzz/boofuzz/primitives/bytes.py", line 130, in mutations
    yield self._adjust_mutation_for_size(fuzz_value=fuzz_value)
  File "/home/dev/PycharmProjects/boofuzz/boofuzz/primitives/bytes.py", line 134, in _adjust_mutation_for_size
    if len(fuzz_value) > self.size:
TypeError: object of type 'functools.partial' has no len()
```

jtpereyda: Thanks; I'll start here with unit tests to chase out this error. I did a functional test but probably missed something.

For some discussion, the Bytes primitive is maybe the only one that returns a function from mutations() sometimes. It's the only one with code that modifies the default value. And with the addition of dynamic default values (taking a value from earlier in the protocol and putting it into a field), the default isn't know at the time mutations() is called. Therefore this design decision to have mutations() return a callable. It's powerful, but makes for a confusing primitive implementation.

SR4ven: It's a little confusing indeed, so thanks for clearing it up!

SR4ven: I'm curious, does this PR fix #400?

jtpereyda: Actually maybe now that I look at it. That might come up while writing unit tests. However, the "default value" needs to remain constant. So if your "default value" is dynamic and the default data can have different lengths, the ability to calculate the numbers goes out the window.

jtpereyda: @SR4ven I have an idea to simplify this, hopefully it will get simpler in a bit.

jtpereyda: @SR4ven The updated version is pushed. Description of changes here.

As you've gathered, `mutations()` can yield a callable that mutates the "normal" value, which may be dynamic.

The previous approach was doing some weird stuff with closures inside the mutations() generator:
```python
        for fuzz_bytes in self._fuzz_strings_1byte:
            i = 0
            keep_going = True
            while keep_going:

                def f(value):
                    nonlocal keep_going
                    if i < len(value):
                        return value[:i] + fuzz_bytes + value[i + 1 :]
                    else:
                        keep_going = False

                yield f
                i += 1
```
The idea was that we wanted to mutate the "value", which can change between protocol runs. BUT, the number of mutations depends on the length of the value. So I did this weird thing with the closure `f` and `nonlocal keepgoing` (nonlocal means the closure can write to the value and the parent function will see it. This still left me with two problems: 1) At the end of the function, we didn't have a value to return. 2) The usage gets weird, as the behavior of the generator depends on the function it yields being called before the next generator iteration.

The compromise I settled on was to have mutations() accept a "default value" (this is how I had it in one of the previous iterators before the PR). So now there's a distinction between "default value" and "normal value". The "default value" must be provided in the protocol definition -- this is either the value as normally provided, or the "default_value" of a `TestCaseSessionReference` object. The "normal value" is usually the same as the default value, but if `TestCaseSessionReference` is used, the normal value is the value referenced by `TestCaseSessionReference`.

Now the `Bytes.mutations()` function uses the default value to control its iteration loops, but it still yields a function that mutates the "normal value".

jtpereyda:
Discussion/thought: I'm thinking of renaming Fuzzable to Mutator, and FuzzableWrapper to Fuzzable. The former is only meant for generating mutations and encoding. The latter is needed for actual fuzzing and manages a lot of data that boofuzz needs.

boofuzz/primitives/bytes.py
jtpereyda: Just pushed finished `unit_tests/tests_bytes.py` which confirms that num_mutations is working for Bytes.

PR #416, opened 2020-04-13
add procmon capture_output option

jtpereyda:
This option enables capturing of stdout and stderr from the target process.

Implemented only in Unix procmon atm.


jtpereyda:
Thanks @SR4ven !

jtpereyda:
doh, forgot to squash. Will remember next time.

PR #419, opened 2020-04-16
Prevent duplicate CI builds

SR4ven:
Disable test workflow when pushing to non master branch. Tests are still performed when creating a PR.
This should help to reduce the amount of duplicate builds on the same commit.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,8 @@ name: Test
 
 on:
   push:
+    branches:
+      - master
   pull_request:
   schedule:
     - cron:  '0 0 * * 0' # @weekly


jtpereyda:
What's the motivation for this change?

The main trade off I can think of is that one won't see progressive builds as they are pushing commits to a branch. I guess that's only if you're developing on a branch rather than a fork though.

SR4ven:
The motivation was that on PRs from branches not coming from forks all tests show up twice.
I just don't see a reason or benefit of running tests on non master on every push. If we want to see the test results, opening a PR does the job. Or a draft PR if it's not ready.
However, if you'd like to keep this as it is, I'm fine with it.

Originally I looked into this as I noticed that the coverage status is missing on the commits recently. I doubt that this is related, but I still can't figure out why. The [coverage app](https://github.com/marketplace/codecov) is activated right? On most commits codecov doesn't recognise that the CI passed and holds back the status message. Probably just an issue at their side if the app is activated 🤷‍♂️. Anyways, that's nothing important and not worth the time fixing I guess.
Coverage results still show [here](https://codecov.io/gh/jtpereyda/boofuzz).

jtpereyda:
Thanks @SR4ven ! Codecov wasn't showing up in the integrations in the repo, but I installed it and it is showing up now. Hope that helps.

SR4ven:
The merge commit contains a codecov status again, so it all seems to be fixed.
Thanks for checking the integration @jtpereyda!

Issue #303, opened 2019-07-11
Stop Rendering All Mutations Upfront (Memory Errors when using python3)


driechersRC:
When running with python3 I seem to run out of memory easily. I have over 1 Gig of free memory and I get this error when I try to send a message just over 2Mb long. Is there anything that can be done to reduce the footprint in the python3 compatibility code? I'm sure it is not the most efficient in python2 either but I only run into the crash with python3.

```
  File "../boofuzz/boofuzz/sessions.py", line 590, in fuzz
    self._main_fuzz_loop(self._iterate_protocol())
  File "../boofuzz/boofuzz/sessions.py", line 695, in _main_fuzz_loop
    self._fuzz_current_case(*fuzz_args)
  File "../boofuzz/boofuzz/sessions.py", line 1424, in _fuzz_current_case
    self.transmit_fuzz(target, self.fuzz_node, path[-1], callback_data=callback_data)
  File "../boofuzz/boofuzz/sessions.py", line 1103, in transmit_fuzz
    data = node.render()
  File "../boofuzz/boofuzz/blocks/request.py", line 154, in render
    self._rendered += item.render()
  File "../boofuzz/boofuzz/blocks/block.py", line 231, in render
    self._rendered += item.render()
  File "../boofuzz/boofuzz/primitives/base_primitive.py", line 72, in render
    value = self._render(self._value)
  File "../boofuzz/boofuzz/primitives/string.py", line 322, in _render
    value = helpers.str_to_bytes(value)
  File "../boofuzz/boofuzz/helpers.py", line 466, in str_to_bytes
    result = six.binary_type().join(temp)
MemoryError
```

SR4ven:
See #259

Are you using subprocesses (popen or similar)? Because the problem with that is, that it forks the main process and only reduces the memory usage afterwards.
So boofuzz with a callback containing any kind of subprocess call will need twice as much memory as without.

I guess memory optimisations would be a great benefit for boofuzz as it could then be used on weaker hardware too.
After taking a brief look at the data generation code, I think it's safe to say that there is quite some potential to save memory. All test data is generated during initialisation and not freed until the script exits.
As the data generation takes close to no time, it should be possible to generate the data for each test case individually on the go and free it right afterwards. This would reduce the memory footprint to a minimum.

jtpereyda:
[...]

PR #422, opened 2020-04-18
Mutations iterator v2

jtpereyda:
Attempt 2.

Before merging: Address all comments from #413


boofuzz/sessions.py
```diff
@@ -1629,27 +1631,25 @@ def _sleep(self, seconds):
...
+        Returns:
 
-        :return: None
         """
-        self.total_mutant_index = 0
-        if self.fuzz_node:
-            self.fuzz_node.reset()
+        message_path = self._message_path_to_str(mutation.message_path)
+        primitive_path = next(iter(mutation.mutations))
+        return "{0}:{1}:{2}".format(message_path, primitive_path, self.mutant_index)
```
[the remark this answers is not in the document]
jtpereyda: Good point. The last message in the path is always the one fuzzed, so as it stands the information is redundant.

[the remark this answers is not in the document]
jtpereyda: The fix will be in Request.py, where names are currently created and parsed.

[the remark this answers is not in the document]
jtpereyda: I tried to fix this, but realized that the additional bit of context is used by `Fuzzable.render()` to determine whether a given test case applies to the current message or not. By removing the Request name, mutations can apply to the wrong message if the name of the primitive(s) is the same.

That said the display still looks a bit funny; maybe there's an easier to way to improve it. Could be left for future improvement.

boofuzz/primitives/string.py
```diff
@@ -26,293 +177,100 @@ def __init__(self, value, size=-1, padding=b"\x00", encoding="ascii", fuzzable=T
...
+            # Number of null bytes to insert (random)
+            for i in range(random.randint(1, 10)):
+                # Location of random byte
+                loc = random.randint(1, len(s))
+                s = s[:loc] + "\x00" + s[loc:]
+            self._fuzz_library.append(s)
 
-            # increment the mutation count.
-            self._mutant_index += 1
+        # Remove any fuzz items greater than self.max_len
```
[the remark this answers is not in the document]
jtpereyda: Will address while writing String unit tests.

boofuzz/sessions.py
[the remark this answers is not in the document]
SR4ven: If the extra information has a purpose, it's fine by me.
We can fix this later.

SR4ven:
From #434 this PR seems to fix #303 and fix #259

erik-smit:
Anything I can do to help with this? 

I tried going through the failing unit_tests, but I don't understand the codebase enough to understand what's supposed to happen.

Like in the test_skip_element. 

```
>       req.mutate()
E       AttributeError: 'Request' object has no attribute 'mutate'```

In test_blocks.py I see mutate() rewritten to.

mutations_generator = req1.get_mutations()
next(mutations_generator)

https://github.com/jtpereyda/boofuzz/pull/422/files#diff-1d77dcdc26ec1e48bde81e9c32a9e0a6L174-R183

Is this the way to fix the test? Should I go through the rest of the tests like this?
Should I put my questions like this in this pull request, should I add as comments in review, or should I just be patient and let you guys do your thing?

**Comment by SR4ven** [comment]
Your help is much appreciated @erik-smit!
Feel free to open PRs to the `mutations-iterator` branch, so someone can review them.

About the tests themselves, I think you're on the right way. Check the diffs of the already rewritten tests (like you already did) and try to apply these changes to the ones currently broken.
I remember that @jtpereyda told me that the original_value tests might not be applicable anymore. Is that still the case? And if so, should we replace or discard them?

**Comment by jtpereyda** [comment]
@erik-smit I've been doing basically that, stepping through failing cases and trying to chase them down. Regarding that specific item, there is no longer a  `mutate()` method. Instead, I've been using `get_mutations()` in tests. See one of the most recent commits where I mad similar changes: https://github.com/jtpereyda/boofuzz/pull/422/commits/afe064d00fbe97c1c26a9a12150a18392caa5de7

Specifically in `test_blocks.py`, here's an example of the newer approach:
```
-        self.assertTrue(s_mutate())
-        self.assertEqual(s_render().find(b"TWO"), -1)
-        self.assertTrue(s_mutate())
-        self.assertEqual(s_render().find(b"ONE"), -1)
-        self.assertFalse(s_mutate())
+        mutations = list(blocks.CURRENT.get_mutations())
+        rendered = blocks.CURRENT.render(MutationContext(mutation=mutations[0]))
+        assert b"ONE" in rendered
+        assert b"TWO" not in rendered
+        rendered = blocks.CURRENT.render(MutationContext(mutation=mutations[1]))
+        assert b"ONE" not in rendered
+        assert b"TWO" in rendered
```

Since `get_mutations()` is a generator, I convert it to a list to make it easy to grab individual mutations. Then you use the render method, but first wrap the mutation in a MutationContext. (As an aside, the MutationContext wrapper shows up a lot; when it's all said and done we might end up finding a way to drop it)

**Comment by jtpereyda** [comment]
@SR4ven I was partially mistaken when I said that. I just checked and it still exists in Fuzzable (fuzzable.py). In the past it may have returned the rendered value, but now it strictly returns the default _value_. The distinction is hopefully clearer now in the code between a node's value and and a node's rendered value. The value is the the result of mutation, whereas a rendered value could have some transformation applied to it. For example, the String primitve's values are strings, but the rendered values are bytes.

So all that to say original_value tests can be just modified to work. However, since original_value is implemented in Fuzzable.py (the super class of all fuzz blocks/primitives), those tests should be less interesting.

**Comment by erik-smit** [comment]
I have a few questions:

1. *.feature tests if callling `reset()` after mutations still returns the original value. Is this still relevant if we're now working with iterators?

2. primitives/mirror.py overrides `original_value()` https://github.com/jtpereyda/boofuzz/blob/mutations-iterator/boofuzz/primitives/mirror.py#L51-L52
 but `get_mutations()` calls original_value without mutation_context: https://github.com/jtpereyda/boofuzz/blob/mutations-iterator/boofuzz/fuzzable.py#L127

 How is this supposed to work?

3. test_blocks.py tests `skip_element()`. I think this function used to stop mutating on the current element, and continues on the next? Do we still have or need this?

**Comment by SR4ven** [comment]
1. Those are mostly the said original value tests. As @jtpereyda they are still relevant though less interesting.
2. I'll take a look into that.
3. Blocks no longer have `skip_element`, we handle skipping test cases directly in the session now. It would be nice to have a test for the new functionality but we can remove the old test as it doesn't apply anymore.
4. Good point, our old test behaviour is not applicable anymore. If I remember correctly `value` is bytes type and is then rendered into the desired format.
To convert the types we'd have to call `render()` again making the test redundant.
Any idea how to fix this?

**Comment by erik-smit** [comment]
> 4. Good point, our old test behaviour is not applicable anymore. If I remember correctly `value` is bytes type and is then rendered into the desired format.
>    To convert the types we'd have to call `render()` again making the test redundant.
>    Any idea how to fix this?

1. Make the test convert original_value to_bytes(1)
```
+++ b/unit_tests/test_bit_field_original_value.py
@@ -40,7 +40,7 @@ def mutate_thrice(context):

 @when("Calling original_value")
 def call_original_value(context):
-    context.result = context.uut.original_value()
+    context.result = context.uut.original_value().to_bytes(1, 'little')
```

2. Make Fuzzable return bytes like before this PR.
Either during setting of self._default_value in `__init__()` or at returning in `original_value()`.
Not sure how to do this properly, because to_bytes() wants a specific size. I saw helpers.str_to_bytes() but not sure...

**Comment by SR4ven** [comment]
The idea behind this is that you can have whatever value type for your primitive internally and turn it into bytes with `render()`.
Any other method to turn the primitives values into bytes might not return the intended results. That's why I find it hard to test this behaviour.

We could though run some hard coded tests like create a string primitive with value `Test`, render it and check if equals `b'Test'` if that's work testing.

**Comment by jtpereyda** [comment]
@SR4ven looks like we're good to merge, right? Stickler calls out some overly complex functions, but those are legacy carry over.

File: boofuzz/blocks/block.py [hunk 3]
```diff
@@ -38,254 +46,141 @@ def __init__(
...
+    #
+    #     # we are done mutating this block.
+    #     if not mutated:
+    #         self._fuzz_complete = True
+    #
+    #         # if we had a dependency, make sure we restore the original value.
+    #         if self.dep:
+    #             self.request.names[self.dep].reset()
+    #
+    #     return mutated
```
SR4ven, line 142: Is this todo still relevant?
Can we remove the orphaned code?

jtpereyda: Good catch -- for full backwards compatibility, we need to port this to enable the behavior where a block will do all its mutations for each group value. Interestingly, this is a special case of mutating multiple items at one time.

To do the port, we'd add a `def mutations(self, default_value):` to override the FuzzableBlock mutations code.

jtpereyda:
> Are all comments from #413 addressed?
Yes, it seems so.

> Merge #454 and #455 first or prioritize this one?
Yes let's merge those in.

> Are all changes properly documented? (I'm quite sure that there are some argument docs that are out of date, but we can fix that later)
There's little to no backwards compatibility breaking, but a bunch of new stuff could/should be documented.

>A changelog entry seems to be missing
Roger.

> Are all primitives functional? (BER lego is known to be broken)
Yes, at least as far as a manual review and unit tests can get us. But our UTs are limited, and only some have been functionally tested. I'd say there's a decent chance of some corner case bugs going forward, but I'd say that's acceptable risk.

So on the pre-merge to do list:

- [ ] Merge 455
- [ ] Add CHANGELOG entry
- [ ] Fix block.py Group strategy

And documentation I think can happen after the merge:

- [ ] New absolute and relative name references
- [ ] New protocol definition syntax
- [ ] New protocol definition syntax tricks -- e.g. TLV function
- [ ] Examples of defining your own primitive
- [ ] New procmon approach (from a previous update actually)

jtpereyda:
@SR4ven need we still rename TestCaseContext and TestCaseSessionReference?

SR4ven:
If it's not too much work, yes that would be nice.

It would spare us the exclude entry in the pytest config section.

boofuzz/blocks/block.py
erik-smit: Can you give an example of how this would be used?

SR4ven: I think we're talking about this
https://github.com/jtpereyda/boofuzz/blob/898a1d2a1927ef22128f93de5d33bcbc39075b47/request_definitions/http.py#L15-L16

Have the block iterate over all group values while fuzzing the contents of the block.

SR4ven: Currently, this branch will always use `GET` and and iterate over all child primitives, while it should repeat fuzzing with `HEAD`, ...

SR4ven: I'm trying to reimplement this feature but I can't find a way to mutate over multiple item (the group and item stack) at the same time.

Any ideas?

SR4ven: @jtpereyda do you have a hint on how to implement the group mutations feature?
Apart from the name resolving discussion this is the only thing holding us back from merging.

jtpereyda: @SR4ven I didn't realize how difficult this bit would be! The old implementation manually mucks with the Group primitive's values. The new implementation doesn't quite work that way. This is actually an instance where we want to mutate two primitives at one time. Great time to try it out!

In Fuzzable.get_mutations, see this excerpt:
```
            for value in itertools.chain(self.mutations(self.original_value()), self._fuzz_values):
                if self._halt_mutations:
                    self._halt_mutations = False
                    return
                if isinstance(value, Mutation):
                    yield value
                else:
                    yield Mutation(mutations={self.qualified_name: value})
```
Especially the last 4 lines --  the mutations method can either render a value (a bytes object), or a Mutation object. The Mutation.muations field is a map from qualified names to mutated values. get_mutations will always normalize the mutation into a `Mutation` object.

So, you can in Block create a Mutation object, add the Group mutation (`"qualified_group_name": b"the_value"`), then manually add the individual mutation based on what you iterated from the item stack.

Pseudocode:
```
for value in theGroup.mutations():
    m = Mutation(mutations={self.request.resolve_name(self.context_path, self.group): value})  # maybe use self.group.context_path?
    for item in self.stack:
        self.request.mutant = item  # not sure about this -- the "mutant" field starts to become meaningless and may need to be reworked or removed eventually
        for mutation in item.get_mutations():
            # merge mutation.mutations into m.mutations
            yield mutation
```

The new design should in theory support and render these two mutations properly, but there may be bugs to work out.

SR4ven:
```python
def mutations(self, default_value):
    for value in self.request.resolve_name(self.context_path, self.group).mutations(default_value):
        m = Mutation(mutations={
            self.request.resolve_name(self.context_path, self.group): value})  # maybe use self.group.context_path?
        for item in self.stack:
            self.request.mutant = item  # not sure about this -- the "mutant" field starts to become meaningless and may need to be reworked or removed eventually
            for mutation in item.get_mutations():
                # merge mutation.mutations into m.mutations
                yield m
```
I tried this code in `Block` with the following request but it always yields `GETa`.
```python
    s_initialize("HTTP VERBS")
    with s_block("body", group="verbs"):
        s_group("verbs", values=["GET", "HEAD"])
        s_static("a")
```
`value` changes correctly, so maybe there's something wrong with the Mutation implementation?

Also, this example yields 4 test cases, though there are only 2 unique ones.

SR4ven: Could use your help here @jtpereyda. Do you have time to look into it?

jtpereyda: Should be able to this week or next. Just had some priorities shift around enough to get back on this. I've been bummed to have left the half-done PR here for so long.

SR4ven: I've been playing around with the groups and it seems to work fine with the examples.

However, the HTTP VERBS request example from above now yields `HEADa` twice now. At least it's not 4 test cases anymore but something still seems to go wrong.
Is that maybe because `GETa` is considered as the "good" case which is skipped but num_mutations is 2 so it yields the same values twice because the mutations are exhausted? In that case this request definition should only yield 1 test case right?

boofuzz/blocks/request.py
```diff
@@ -3,9 +3,16 @@
...
 from .. import exception
+from ..exception import BoofuzzNameResolutionError
 from ..fuzzable import Fuzzable
 from ..fuzzable_block import FuzzableBlock
 
+cannot_resolve_name = "Failed to resolve block name '{0}' in context '{1}'"
+name_not_found = "Cannot find block with name '{0}'"
+too_many_matches = (
+    "Unable to resolve block name '{0}'. Use an absolute or relative name instead." " Too many potential matches: {1}"
+)
```
SR4ven, line 14: We already have some error message constants in `constants.py`. Maybe keep all error messages in one place?
But then again, we'll most likely never use those constants in any other place.
It's up to you.

boofuzz/blocks/block.py
jtpereyda: Most likely caused by the group being inside the block itself. Since the rule is to mutate the block once for each mutation of the relevant group... but in this case that group is in the block... I can't quite wrap my head around what it _should_ do. I think if you take the verbs field outside of the block and put something else in, the behavior will make more sense.

I think we can leave it for now. If there are any ideas on the "right" answer when the group is a part of the block, feel free to record them here for future reference.

SR4ven (approved):
Great work! Especially with the detailed unit tests, they're definitely come in handy in the future.

I'll push one more commit to stay compliant with the latest black code style update.
After that we're good to go from my side.
Once you approve @jtpereyda I'll push the release before merging.

jtpereyda:
Ignoring complexity warnings on existing functions, and errors in the largely outdated legos files.

Thanks for the review @SR4ven and contributions @erik-smit .

CHANGELOG.rst
```diff
@@ -1,6 +1,41 @@
...
+- Name references for `Checksum`, `Size`, etc. now resolve absolute and relative names.
+- Block and primitive names no longer need to be globally unique within a message, they only need to be locally unique within a block.
+- Independent mutation and encoding steps -- will enable multiple mutations and code coverage feedback.
+- Callbacks now have `ProtocolSession` object to which one can save data to be used later in the protocol definition.
+- Added `ProtocolSessionReference`, with which can use data from a past reply in a request definition.
+
+Fixes
+^^^^^
+- Various web interface fixes.
+
```
SR4ven, line 38: Looks like some duplicate lines sneaked in here

jtpereyda: oops

PR #425, opened 2020-04-23
Added Logos

### LINKED ISSUE #407: boofuzz Logo
Hey,

as promised in #375 I have a sketch for a boofuzz logo. At the moment It's nothing more than a quick sketch which I'll refine later when I have time. Consider all art I post CC0. Obviously, I have vectors (SVG/EPS) for each sketch as well, but they probably need some cleanup before I can post them. If you have more Ideas or if you picture something, feel free to comment.

boo-logo
boo-logo-color

Font currently used is Source Sans Pro, although that may change. I tried to include more of boos halloween costume, but unfortunately that would mess up the proportions of the logo too much, especially if it should be recognizable even when displayed small.

Personally, I like the color one better; but ultimately, that's up to you @jtpereyda.

mistressofjellyfish:
Because Artwork is not code, it's not covered by the GPL.
I therefore release all artwork under the terms of the
[CC0 license](https://creativecommons.org/publicdomain/zero/1.0/deed.en).

While attribution is not required nor sought, in case anyone
*wants* to credit, please use my real name (obtainable from
the AUTHOR of this commit) instead of any nicknamse I might
use - thank you.

As discussed in #407, I've used the purple version. I rendered a light and dark variant. The favicon is available as 16x16, 32x32, 64x64 and 96x96 in both ico and PNG format. All variants have vectors provided in EPS and SVG format. Fonts inside SVG vectors are rendered to forms to work in situations where the Source Sans Pro is not available, EPS should have the font embedded.

@jtpereyda and @SR4ven: sorry that it took me so long, I had a bit of a situation going on - nothing health related fortnuately ;) It should all be sorted now.

Fixes #407


mistressofjellyfish:
> Is artwork the right place to put these files? Maybe merge it with _static?
I could do that later in another PR, shouldn't prevent us from merging this one.

Well, _static is for images in use - first- or third-party. I created artwork specifically to have a place where the "source files" are stored. For usage in the docs, README or whatever I'd copy or link them to _static - think of the files in the artwork folder as a source of truth; everything else may be scaled, filtered, or optimized for the target media (jpeg for example could be way more size-efficient than png, even with very little actual visible compression). For example, I would never include the big > 1000px wide PNG logo in the docs, either render it smaller or use SVG, if it scales well.

SR4ven:
Sounds reasonable.
We could also move both dirs into something like `/res` to keep the root dir clean, but I'm fine with your approach.

jtpereyda:
LGTM as is. Next step will be to get it into the docs?

SR4ven:
> 
> 
> LGTM as is. Next step will be to get it into the docs?

Yes, I should be able to find some time to set that up until Sunday.

jtpereyda:
btw getting svg->png into the build process would minimize the compiled objects in the repo; not necessary to merge

SR4ven:
> 
> 
> btw getting svg->png into the build process would minimize the compiled objects in the repo; not necessary to merge

Good point @jtpereyda, the way we have requirements, docs and CI set up, this should be doable.
The easiest way would be a python package that handles the conversion. No idea if there is one.

PR #428, opened 2020-04-27
Use new logo in docs and webinterface

SR4ven:
Places where the new logo is now used:
- [README](https://github.com/jtpereyda/boofuzz/blob/logo/README.rst)
- [Docs](https://boofuzz.readthedocs.io/en/logo/) sidebar
- Docs favicon as ICO and SVG
- Webinterface as ICO and SVG

Might have not been worth the long time I needed to implement, but the favicons are now advertised as SVG and ICO to the browser. Benefit here is, that the browser gets to choose whatever is supported.

I'm not 100% happy with the implementation of the logo in the README, as for some reason, all headings are one level lower than they were before. This apparently happens when a figure is placed before the first heading, but I don't know how to fix this.

I tried no to copy and images, but refer to the ones in /artwork. However, in /boofuzz/web it was necessary, to make sure that the wheel release contains all data.


SR4ven:
@jtpereyda are you ok with the new README too? As far as I know the only way to fix the smaller headlines is to place the logo below the first heading, which again doesn't look nice either.
I mean the way it currently is, is acceptable I guess so if you're ok with it we can merge.

jtpereyda:
@SR4ven What do you mean about smaller headlines?

SR4ven:
You can see it when comparing current master and logo branch README side by side. Not too much of a difference but it still bugs me a little

jtpereyda:
Ah, I see that now. 🤷  Works for me as is.

PR #431, opened 2020-04-28
Minor console GUI optimizations

SR4ven:
Some minor optimizations for the console GUI I had laying around for a while.
Most noticeably is that the terminal won't break when quitting with CTRL+C.


jtpereyda:
Cool! We can go ahead and merge. I tried it out and the curses UI still runs, though it doesn't fix a crash problem I have. I'll report separately.

PR #434, opened 2020-05-11
Memory optimization fix for s_string

julianrother:
The `String` class has the class attribute `_fuzz_library` to share a huge part of the library between instances and save a lot of memory as is mentioned in a comment:

> store fuzz_library as a class variable to avoid copying the ~70MB structure across each instantiated primitive.

Currently this does not work, because `_fuzz_library` is not updated as a class attribute (shared between all instances) but as an instance attribute (not shared). Therefore each use of `s_string` initialises it's own `_fuzz_library` and allocates additional ~70MB of memory.

To demonstrate this:
```
# test_str.py
import sys, os, psutil
from boofuzz import *

before = psutil.Process(os.getpid()).memory_info().rss
s_initialize(name="test")
for i in range(int(sys.argv[1])):
  s_string('')
after = psutil.Process(os.getpid()).memory_info().rss
print('%.2fMB'%((after-before)/1024**2))
```

With current master:
```
$ python3 test_str.py 1
63.14MB
$ python3 ./test_str.py 10
631.70MB
```
After this fix:
```
$ python3 test_str.py 1
63.14MB
$ python3 ./test_str.py 10
63.14MB
$ python3 ./test_str.py 1000
63.65MB
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -35,6 +35,7 @@ Fixes
 - Removed deprecated logger `FuzzLoggerFile`.
 - Fixed network monitor compatibility with Python 3.
 - Minor console GUI optimizations.
+- Fixed memory optimization for `s_string`
 
 v0.1.6
 ------
--- a/boofuzz/primitives/string.py
+++ b/boofuzz/primitives/string.py
@@ -279,7 +279,7 @@ class String(BasePrimitive):
                 return False
 
             # update the current value from the fuzz library.
-            self._value = (self._fuzz_library + self.this_library)[self._mutant_index]
+            self._value = ((self._fuzz_library or String._fuzz_library) + self.this_library)[self._mutant_index]
 
             # increment the mutation count.
             self._mutant_index += 1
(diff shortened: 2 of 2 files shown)


jtpereyda:
Thanks for the PR @julianrother ! One issue with this I believe is that, since the array is modified, this will cause some mixups when you have multiple String types in use in your protocol definition.

Those memory benchmarks are extremely helpful. That suggests String is the major contributor to memory bloat in boofuzz right now.

There is a similar fix active in string.py in #422 -- unfortunately I haven't merged yet as I need to update the unit tests to pass. The PR is a major update that happens to fix this same issue (or at least it comes close). I'll consider this my reminder to prioritize those unit test updates and get the fix merged.

If you happen to have time to try out your memory benchmark on that branch (or after it's merged), please do share the results!

julianrother:
Thank you for the feedback! I've updated the branch so that the mixup issue should be resolved now (mainly because I need to backport this anyway).

You're right that #422 also fixes this problem:
```
$ python ./test_str.py 1
0.00MB
$ python ./test_str.py 1000
2.95MB
```

PR #435, opened 2020-05-12
Fix behaviour of skip_element on blocks

julianrother:
Currently if `Request.skip_element` is called while the current mutant is a primitive **inside a block**, the whole block is skipped instead of only the primitive. This is noticable when `crash_threshold_element` is exceeded on such a primitive. In this case `total_mutant_index` is advanced as if only the primitive is skipped while actually more cases are skipped, so this behaviour is inconsistent and messes up test case numbering. See also the test case I've added for what I think would be the correct behaviour.

My change basically copies the code related to `skip_element` and `_element_mutant_index` tracking from the `Request` to the `Block` class (cf. 458ed968) and recurses through all blocks until a primitive is reached.

A much simpler approach to `skip_element` in general would be to simply mutate the current mutant until it is exhausted:
```
class Request(IFuzzable):
    # ...
    def skip_element(self):
        while self.mutant.mutate():
            pass
```
This way the whole code for tracking `_element_mutant_index` could be removed with an IMHO insignificant performance impact. If you prefer this approach, I would update the PR accordingly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -35,6 +35,7 @@ Fixes
 - Removed deprecated logger `FuzzLoggerFile`.
 - Fixed network monitor compatibility with Python 3.
 - Minor console GUI optimizations.
+- Fixed crash_threshold_element handling if blocks are used
 
 v0.1.6
 ------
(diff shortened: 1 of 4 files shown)


SR4ven:
Good catch @julianrother! Thanks for the fix.

The only problem I see is that we somehow have to merge this into #422 at some point.
@jtpereyda which PR do you want to merge first?

jtpereyda:
Thanks for the PR @julianrother !

I'm not sure exactly how this will interact with #422 (skipping has changed there), but since this PR is working and ready we can merge it for now. I'll take responsibility for merging into #422, and the additional unit test will be a convenient reminder for me.

PR #437, opened 2020-05-24
Import crc32c only if needed

### LINKED ISSUE #436: Dependency crc32c requires compiler on Windows
Installing boofuzz on Windows Python 3.8 failed for me because Microsoft Visual C++ Build Tools are required. This is caused by the dependency `crc32c`, which doesn't provide a wheel for Python 3.8 on Windows amd64 yet.

Maybe we should drop this dependency in favour of a simpler installation process?
I see three possibilities here:
- Completely remove crc32c form boofuzz and refer to the callback function which users can use to implement any checksum algorithm. crc32c was originally added in #353 because this feature was not properly documented, but it is now. https://boofuzz.readthedocs.io/en/latest/user/static-protocol-definition.html#boofuzz.s_checksum
- Remove the dependency, keep the implementation. This should work as we already have an include guard for crc32c. If the package is missing, the feature simply won't be available.
- Leave everything as it is and wait/hope for crc32c to release new wheels.

I don't know how many users are utilising the crc32c feature, so any input is welcome.

SR4ven:
As discussed in #436, `crc32c` is removed from the requirements. Fixes #436 

Implementation stays in place but will fail with import error if `crc32c` algorithm is specified but package is not present.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -36,6 +36,7 @@ Fixes
 - Fixed network monitor compatibility with Python 3.
 - Minor console GUI optimizations.
 - Fixed crash_threshold_element handling if blocks are used
+- `crc32c` is no longer a required package. Install manually if needed.
 
 v0.1.6
 ------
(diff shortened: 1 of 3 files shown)


jtpereyda:
Thank you @SR4ven

PR #442, opened 2020-06-21
Fix test_target for inaccurate sleep times

SR4ven:
The weekly test failed on Windows as the actual sleep time was shorter than expected one for the target unit tests.
See https://github.com/jtpereyda/boofuzz/runs/792499474?check_suite_focus=true#step:7:57

Apparently the sleep function isn't as accurate on Windows as it is on Linux, at least for Python 2. Rounding the time to 2 decimal places should fix this.


jtpereyda:
Thanks @SR4ven !

PR #446, opened 2020-07-10
Fix execution of procmon stop_commands

### LINKED ISSUE #445: Process Monitor Stop Commands Not Working
Hello,

I discovered this issue when the process_monitor_unix.py failed to restart a program that had just crashed. I got the following error:

 File "/home/debian/fuzzENV/lib/python2.7/site-packages/boofuzz/utils/process_monitor_pedrpc_server.py", line 210, in _stop_target
    os.system(command)
TypeError: system() argument 1 must be string, not list

The process_monitor_unix.py script failed to successfully stop my program because it didn't understand my stop command. When I wrote my fuzzing script, I set the start and stop commands with:
`
procmon.set_options(start_commands=["make start"], stop_commands=["TERMINATE_PID"])
`
According to function `_stop_target()` in [process_monitor_pedrpc_server.py](https://github.com/jtpereyda/boofuzz/blob/master/boofuzz/utils/process_monitor_pedrpc_server.py), the command "TERMINATE_PID" should tell the target process to stop automatically.

However, it seems that the script ignored the "TERMINATE_PID" and skipped straight to `os.system(command)`.

After some further digging, I found that `set_stop_commands()` uses `_split_command_if_str()` to save my stop commands as a list of strings, formatted in a way that is compatible with `subprocess.Popen()`. Which means that using `os.system()` to call my command is incorrect, because `os.system()` expects a string, not a list of strings, resulting in the error above.

I made the following changes to `_stop_target()` in my local directories.

1. Changed `if command == "TERMINATE_PID"` to `if command == ["TERMINATE_PID"]' because stop commands are stored as lists of strings
2. Changed `os.system(command)` to `subprocess.Popen(command)`
3. Added `import subprocess` to the top of process_monitor_pedrpc_server.py

This seemed to fix my problems, and the "TERMINATE_PID" command works as expected. I also noticed that the start commands are called via `subprocess.Popen()` too, which is why the process manager was able to start my program in the first place.

**TL;DR: The stop commands are formatted to be compatible with `subprocess.Popen()`, but the `_stop_target()` function incorrectly treats them as regular strings**

Thanks for reading, and keep up the good work!

SR4ven:
Fixes #445 

Would be nice if you could verify that this works @tjzhang1.

> Changed if command == "TERMINATE_PID" to `if command == ["TERMINATE_PID"]' because stop commands are stored as lists of strings

This didn't work for me. `command` is not a list, just `stop_commands` is, so I left this unchanged.
Please correct me if I've missed something here.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Features
 Fixes
 ^^^^^
 - Fixed UDPSocketConnection data truncation when sending more data than the socket supports.
+- Fixed execution of procmon stop_commands.
 
 v0.2.0
 ------
--- a/boofuzz/monitors/pedrpc.py
+++ b/boofuzz/monitors/pedrpc.py
@@ -256,7 +256,7 @@ class Server(object):
             try:
                 self.__client_sock.shutdown(socket.SHUT_RDWR)
             except socket.error as e:
-                if e.errno == errno.ENOTCONN:
+                if e.errno in [errno.ENOTCONN, errno.EBADF]:
                     pass
                 else:
                     raise
--- a/boofuzz/utils/process_monitor_pedrpc_server.py
+++ b/boofuzz/utils/process_monitor_pedrpc_server.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 
 import os
 import shlex
+import subprocess
 import time
 
 from builtins import str
@@ -177,8 +178,6 @@ class ProcessMonitorPedrpcServer(pedrpc.Server):
         Kill the current debugger thread and stop the target process by issuing the commands in self.stop_commands.
         """
         self.log("Stopping target...")
-        # give the debugger thread a chance to exit.
-        time.sleep(1)
 
         if self._target_is_running():
             self._stop_target()
@@ -201,13 +200,17 @@ class ProcessMonitorPedrpcServer(pedrpc.Server):
         time.sleep(1)
         if len(self.stop_commands) < 1:
             self.debugger_thread.stop_target()
+            while self.debugger_thread.isAlive():
+                time.sleep(0.1)
         else:
             for command in self.stop_commands:
                 if command == "TERMINATE_PID":
                     self.debugger_thread.stop_target()
+                    while self.debugger_thread.isAlive():
+                        time.sleep(0.1)
                 else:
                     self.log("Executing stop command: '{0}'".format(command), 2)
-                    os.system(command)
+                    subprocess.Popen(command)
 
     def _target_is_running(self):
         return self.debugger_thread is not None and self.debugger_thread.isAlive()
--- a/boofuzz/utils/process_monitor_pedrpc_server.py
+++ b/boofuzz/utils/process_monitor_pedrpc_server.py
@@ -204,7 +204,7 @@ class ProcessMonitorPedrpcServer(pedrpc.Server):
                 time.sleep(0.1)
         else:
             for command in self.stop_commands:
-                if command == "TERMINATE_PID":
+                if command == ["TERMINATE_PID"] or command == "TERMINATE_PID":
                     self.debugger_thread.stop_target()
                     while self.debugger_thread.isAlive():
                         time.sleep(0.1)


tjzhang1:
Interesting. I tested my code using the comparison `if command == "TERMINATE_PID"` like you mentioned above, but that results in an error for an invalid system command. I was expecting it to stop the debugging thread instead of skipping to `subprocess.Popen(command)`. I copied the error below. I also added the line `print(command, "is of type:", type(command))`  to the for loop so I can see what type `command` is (output corresponds to the third line below). 

[03:07.59] Restarting target...
[03:07.59] Stopping target...
**['TERMINATE_PID'] is of type: <type 'list'>**
  File "process_monitor_unix.py", line 110, in <module>
    go()
  File "/home/debian/fuzzENV/lib/python2.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/debian/fuzzENV/lib/python2.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/debian/fuzzENV/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/debian/fuzzENV/lib/python2.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "process_monitor_unix.py", line 105, in go
    coredump_dir=coredump_dir,
  File "process_monitor_unix.py", line 57, in serve_procmon
    servlet.serve_forever()
  File "/home/debian/fuzzENV/lib/python2.7/site-packages/boofuzz/monitors/pedrpc.py", line 348, in serve_forever
    ret = method(*args, **kwargs)
  File "/home/debian/fuzzENV/lib/python2.7/site-packages/boofuzz/utils/process_monitor_pedrpc_server.py", line 226, in restart_target
    self.stop_target()
  File "/home/debian/fuzzENV/lib/python2.7/site-packages/boofuzz/utils/process_monitor_pedrpc_server.py", line 186, in stop_target
    self._stop_target()
  File "/home/debian/fuzzENV/lib/python2.7/site-packages/boofuzz/utils/process_monitor_pedrpc_server.py", line 214, in _stop_target
    subprocess.Popen(command)
  File "/usr/lib/python2.7/subprocess.py", line 394, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1047, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory

It appears that for my system, `command` is a list containing a string and `stop_commands` is a list of these lists. I believe this is intended though because `subprocess.Popen()` only accepts lists for its argument.

SR4ven:
@tjzhang1 which version of boofuzz are you using?
The difference we have is that I used Python 3, while you are using Python 2.

When I tried Python 2, I ran into some completely different error which I described in #447. The changes leading to this incompatibility were introduced a couple of months ago. As you don't seem to get this error, I guess you're not on v0.2.0 or higher.

So now we've got two things to fix...

tjzhang1:
@SR4ven 

I switched to Python 3.7 to check, but I get the same behavior with `command` being a list and not a string. Like before, the code skips the comparison `if command == "TERMINATE_PID"` and jumps to `subprocess.Popen()`. Error is below:

```
[02:57.52] Restarting target...
[02:57.52] Stopping target...
Stop command of type: <class 'list'>
  File "process_monitor_unix.py", line 110, in <module>
    go()
  File "/home/debian/fuzzENV3/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/debian/fuzzENV3/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/debian/fuzzENV3/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/debian/fuzzENV3/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "process_monitor_unix.py", line 105, in go
    coredump_dir=coredump_dir,
  File "process_monitor_unix.py", line 57, in serve_procmon
    servlet.serve_forever()
  File "/home/debian/fuzzENV3/lib/python3.7/site-packages/boofuzz/monitors/pedrpc.py", line 348, in serve_forever
    ret = method(*args, **kwargs)
  File "/home/debian/fuzzENV3/lib/python3.7/site-packages/boofuzz/utils/process_monitor_pedrpc_server.py", line 226, in restart_target
    self.stop_target()
  File "/home/debian/fuzzENV3/lib/python3.7/site-packages/boofuzz/utils/process_monitor_pedrpc_server.py", line 186, in stop_target
    self._stop_target()
  File "/home/debian/fuzzENV3/lib/python3.7/site-packages/boofuzz/utils/process_monitor_pedrpc_server.py", line 214, in _stop_target
    subprocess.Popen(command)
  File "/usr/lib/python3.7/subprocess.py", line 775, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.7/subprocess.py", line 1522, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'TERMINATE_PID': 'TERMINATE_PID'
```

Also, while I was using Python 2, I was using boofuzz v0.2.0, and I was getting the thread exception error as you describe in [#447](https://github.com/jtpereyda/boofuzz/issues/447). It didn't seem to have a significant effect on my fuzzing script though so I didn't include it in my original post.

SR4ven:
Seems like we set the stop command differently. Could you share your script please?

This is the code I tested with.

```python
start_cmd = ["python3", "./simple_http.py"]
procmon = ProcessMonitor("127.0.0.1", 26002)
procmon.set_options(start_commands=[start_cmd], stop_commands=["TERMINATE_PID"])

sess = Session(
    target=Target(connection=TCPSocketConnection("127.0.0.1", 8000),
                  monitors=[procmon]),
    ...
```

tjzhang1:
This is mine:

```
from boofuzz import *

procmon = ProcessMonitor("192.168.1.133", 26002)
procmon.set_options(
        start_commands=["make start"], 
        stop_commands=["echo stopping", "TERMINATE_PID"],
        proc_name="server.sh"
        )


def main():
    session = Session(
        target=Target(
            connection=TCPSocketConnection("192.168.1.133", 4000),
            monitors=[procmon]
        )
    )
    ...
```

I just have an extra stop command `"echo stopping"` compared to your code.

SR4ven:
I was able to reproduce your error now.

The recent commit should fix it. Would be nice if you could verify that @tjzhang1.

tjzhang1:
@SR4ven 
Works as expected! Thanks for the help!

SR4ven:
Thanks to you too @tjzhang1!

PR #450, opened 2020-07-20
Fix TCP and SSL server connections

### LINKED ISSUE #449: Fuzzing a client aborts with "Error!!!! Unexpected exception!"
I'm trying to fuzz a client by having boofuzz act as a server. For starters, I'm trying to run the [fuzz_ssl_client.py example](https://github.com/jtpereyda/boofuzz/blob/872d33be42e97b9e37bbe47a086aa585d8c50218/examples/fuzz_ssl_client.py).

Running the example opens the target connection and waits for a client to connect. However, connecting a client (for example `openssl s_client -connect 127.0.0.1`) crashes the server with
```
Error!!!! Unexpected exception! Traceback (most recent call last):
  [...]
  File [...]/boofuzz/connections/tcp_socket_connection.py", line 71, in open
OSError: [Errno 106] Transport endpoint is already connected
```

How can I make the server handle multiple connection attempts in order to continue fuzzing? I'm using boofuzz version 0.2.0 from pip.

SR4ven:
Fixes #449 

Turns out fixing this wasn't too much work after all. The main problem was the SSL socket wrapping which wasn't in the right place.

It kinda bothers me that we have SSL related code in `TCPSocketConnection` now but I couldn't think of any other way. A callback might be an option if you prefer that over the current solution @jtpereyda.

Would be nice if you verify that this fixes your problem @hrantzsch

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,7 @@ Fixes
 ^^^^^
 - Fixed UDPSocketConnection data truncation when sending more data than the socket supports.
 - Fixed execution of procmon stop_commands.
+- Fixed TCP and SSL server connections.
 
 v0.2.0
 ------
--- a/boofuzz/connections/ssl_socket_connection.py
+++ b/boofuzz/connections/ssl_socket_connection.py
@@ -37,8 +37,6 @@ class SSLSocketConnection(tcp_socket_connection.TCPSocketConnection):
             raise ValueError("SSL/TLS requires either sslcontext or server_hostname to be set.")
 
     def open(self):
-        super(SSLSocketConnection, self).open()
-
         # If boofuzz is the SSL client and user did not give us a SSLContext,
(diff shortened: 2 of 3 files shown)


jtpereyda:
@SR4ven one idea besides using callbacks:

Create one or more private methods to do some of the work of `TcpSocketConneciton.open()`. Maybe `_tcp_open_socket()` (just the open line) and `_tcp_connect_server_or_client()`. Then in the SSL `open()` method, you call these two private methods and throw the SSL lines in between.

hrantzsch:
The fix works for me, thanks a lot!

SR4ven:
Sounds good @jtpereyda, I applied the changes.

Some unit tests for server sockets would have helped to prevent this problem in the first place. When I find the time I'll see if I can come up with some.

Also Stickler complains about not being able to clone the repo. Maybe some temporary outage on their end.

Issue #370, opened 2020-01-16
Drop python 2 support


mistressofjellyfish:
So, the Python 2 End of Life is now finally around the corner and happens on April 2020. [A number of famous projects](https://python3statement.org/) already stopped supporting py2 in new releases; and even though boofuzz' support for Python 3 is relatively new, I feel we should discuss about dropping python 2 support with some future version.

Dropping python 2 would enable us to use new Python 3 exclusive features such as Typehints, keyword-only-arguments (looking at you, Target and Session), Pathlib, and others.

Now, I know that this is a somewhat big deal so it would probably make sense to make a road map and decide which version would be the last to support Python 2.

mistressofjellyfish:
Also, at least one dependency (Tornado) already dropped python 2 support.

jtpereyda:
Advantages of moving to Py 3:
1. Use all the helpful language features from Py 3.

Disadvantages:
1. Limit compatibility on some systems (since boofuzz is a research tool rather than an on-target exploit tool, I don't think this matters much TBH).
2. Unable to use more obscure dependencies.

Any others?

If that's it and we could get rid of disadvantage 2, I'd say let's go for it. I suspect these two more-obscure Windows dependencies haven't been tested on Python 3:
1. [pydbg](https://github.com/Fitblip/pydbg)
2. [pydasm](https://github.com/jtpereyda/libdasm/tree/master/pydasm)

Has anybody tried these with Python 3? Or know of any replacements?

mistressofjellyfish:
Couldn't pydasm be replaced by [capstone](https://github.com/aquynh/capstone)? Additional benefits include support for x86_64 and a boatload of other architectures like ARM.

I don't know of a replacement for pydbg, that is something that someone shoud look into.

jtpereyda:
Now that I look, boofuzz only uses pydbg directly, and pydbg uses pydasm.

I've received in-person feedback about boofuzz not supporting 64-bit applications on Windows due to pydbg's status, so switching to something modern would be a big plus. I'm sure there is something newer out there, but I don't know what.

Back to upgrading: I believe pydb is only used by the procmon app, so I guess we could isolate Py2 compatibility there. That leads to a bit of a messy situation though.

keisentraut:
[...]

PR #453, opened 2020-07-30
Mutations iterator v2 help

erik-smit:
- updated the unit_tests to iterator v2 style.
 - removed the mutation/render variants.
 - fixed pytest throwing many `boofuzz\test_case_context.py:5: PytestCollectionWarning: cannot collect test class 'TestCase` warnings.
 - add s_checksum test to show broken s_checksum
 - fix broken s_checksum()
 - make size.py `__init__` `super()` call compatible with py2.7
 - fix broken s_mirror()
 - xfailed known test_target failure on windows

Todo:

- `s_checksum()` and `s_repeat()` use different syntax to resolve `block_name`.  
`s_repeat("1_static_byte")` vs. `s_checksum("simple_boofuzz_scenario.1_static_byte")`.


unit_tests/test_bit_field_original_value.py
```diff
@@ -7,40 +7,40 @@
...
-    context.uut.mutate()
+    next(context.uut.get_mutations())
+    next(context.uut.get_mutations())
+    next(context.uut.get_mutations())
 
 
 @when("Calling original_value")
 def call_original_value(context):
-    context.result = context.uut.original_value
+    context.result = context.uut.original_value().to_bytes(1, "little")
```
SR4ven, line 43: Doesn't this do the same as `render()` or `encode()`?
Seems related to https://github.com/jtpereyda/boofuzz/pull/422#issuecomment-666663935. This doesn't feel like a very robust rest.
It might work in this particular case but maybe we should drop the original_value tests completely after all @jtpereyda?
As I mentioned in the comment above, I don't think these tests are very meaningful. All we could do is test a few selected cases where we expect a given outcome.

unit_tests/bit_field_original_value.feature
```diff
@@ -5,24 +5,6 @@ Feature: BitField.original_value
         When Calling original_value
         Then Result equals .render()
 
-    Scenario: Same as initial render after mutation
```
SR4ven, line 8: Why remove the scenarios but not the implementation? And why remove them at all? Are these tests irrelevant now?
Is it because we no longer have `reset()`? If so there are still some tests using reset.

unit_tests/test_string_original_value.py
```diff
@@ -30,7 +30,7 @@ def mutate_thrice(context):
 
 @when("Calling original_value")
 def call_original_value(context):
-    context.result = context.uut.original_value
+    context.result = helpers.str_to_bytes(context.uut.original_value())
```
SR4ven, line 33: Same as before, this is what `encode()`. Doesn't that make this test redundant?

unit_tests/test_bit_field_original_value.py
erik-smit: Testing basic functionality to see if it's doing what you expect sounds valuable to me.
Maybe it's what `render()` or `encode()` does, but the test will let us know if it's accidently broken.

erik-smit:
Revamping the tests sounds sensible. We've already got many different test-styles.

Like, 
1. test_helpers_ip_str_to_bytes.py uses pytest_bdd with `@scenario`, `@pytest.mark.parametrize`, etc.
2. test_block_original_value.py uses pytest_bdd but with `scenarios("block_original_value.feature")`.
3. test_blocks.py uses pytest and unittest. `if __name__ == "__main__": TestBlocks.debugTestCase()`.
4. test_fuzz_logger.py uses `unittest.main()`.
5. test_helpers_udp_checksum.py uses `pytest.main()`.

Also, having `test_block_original_value.py` dependent on `block_original_value.feature` was confusing to me.
I'd say `test_block_original_value.py` should depend on `test_block_original_value.feature`. 
Having them next to each other in a alphabetically sorted listing would be more clear.

Is there a strong reason why all tests are different? Wouldn't it be better to make them all similar?
I wouldn't mind writing the code for this if we can make a decision what structure we'd want.

unit_tests/bit_field_original_value.feature
erik-smit: It's because we no longer have reset. And yeah the implementation should be removed also.

unit_tests/test_string_original_value.py
erik-smit: Same as before. Testing basic functionality is useful.

unit_tests/test_bit_field_original_value.py
SR4ven: Alright I'm convinced.
It does work fine for this and the string primitive. Definitely better than nothing.

SR4ven:
Agreed, the tests are kinda messy at the moment.

I like the bdd tests because they are easy so easy to read.
If there are good reasons against bdd tests, I'd say go for pytest as it supports all features from unittest AFAIK and it's the module we execute the tests with. On top it's integrated in PyCharm kinda nicely.

Renaming the feature files seams reasonable too.

I doubt that there's any good reason for the current mess, most likely those tests were implemented by different people. Unifying all tests would be a great benefit, I'm definitely up for that.
However, #422 is already huge, so if possible, can we work on that in a different PR? Maybe just hotfix the current ones and rewrite them later?

erik-smit:
> I doubt that there's any good reason for the current mess, most likely those tests were implemented by different people. Unifying all tests would be a great benefit, I'm definitely up for that.
> However, #422 is already huge, so if possible, can we work on that in a different PR? Maybe just hotfix the current ones and rewrite them later?

Sounds like a plan.

Then I've got the following broken tests left:
```
FAILED unit_tests/test_primitives.py::TestPrimitives::test_s_mirror - TypeError: original_value() missing 1 required positional argument: 'mutation_context'
FAILED unit_tests/test_block_original_value.py::test_same_as_initial_render - AssertionError: assert None == b'\x01'
FAILED unit_tests/test_checksum_original_value.py::test_same_as_initial_render - KeyError: 'unit-test-block'
FAILED unit_tests/test_request_original_value.py::test_same_as_initial_render_1_block - AssertionError: assert None == b'\x00'
FAILED unit_tests/test_request_original_value.py::test_same_as_initial_render_many_blocks - AssertionError: assert None == b'\x01The perfection of art is to conceal art.'
FAILED unit_tests/test_size_original_value.py::test_same_as_initial_render - AssertionError: assert None == b'\x02\x00\x00\x00'
```

The "None ==" failures are from tests with children. The `original_value()` only returns the value of the parent, not the children.
If we're gonna drop the `original_value() == render()` test anyhow, fine to rewrite them now to just test for a specific result?

The s_mirror we discussed in #422. Any idea?

The KeyError 'unit-test-block' I don't understand yet. self.requests._names only contains the children, not the parent block.

erik-smit:
There's now one erroring test left: `FAILED unit_tests/test_primitives.py::TestPrimitives::test_s_mirror - TypeError: original_value() missing 1 required positional argument: 'mutation_context'`

`get_mutations()` does: `for value in itertools.chain(self.mutations(self.original_value()), self._fuzz_values):`.

In fuzzable.py: `def original_value(self, test_case_context=None):`
In mirror.py: `def original_value(self, mutation_context):`

Not sure how this is supposed to work. @jtpereyda?

erik-smit:
Testing this branch against my boofuzz ipmi fuzzer showed that `s_checksum()` was broken, but the current test didn't cover it.

Before we merge mutations-iterator we should probably add at least minimal tests for all `s_*()`?

unit_tests/test_bit_field_original_value.py
```diff
@@ -7,40 +7,22 @@
...
-    context.uut.mutate()
-    context.uut.mutate()
-    context.uut.mutate()
+    context.uut = BitField(default_value=value, width=32, output_format="ascii", name="4_bytes")
 
 
 @when("Calling original_value")
 def call_original_value(context):
-    context.result = context.uut.original_value
+    context.result = context.uut.original_value().to_bytes(1, "little")
```
SR4ven, line 25: `to_bytes()` is not available in Python 2, maybe we can use `struckt.pack()`?

.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
```
SR4ven, line 19: We cannot drop Python 2 support that easily, see #370.

Making the tests Python 2 compatible should be possible from what I've seen. It's only the syntax for `super()` calls and the `to_bytes` conversion.

erik-smit: For `super()` I copied the syntax from blocks.py which didn't break on py2.7 and now it works.
I don't fully understand the `super()` problem, but it's gone now.

erik-smit:
test_counter_repeater fails with this on their windowspy27:
```
        self.assertEqual(repeater.count, connection.count)
>       self.assertGreaterEqual(connection.second - connection.first, self.SLEEP_TIME)
E       AssertionError: 0.009999990463256836 not greater than or equal to 0.01
```

A rerun doesn't fix it, but it's  it seems recurring but only on their environment. 
pytest with python2.7 on my windows pc succeeds just fine.
I wonder if it's a "windows doesn't sleep long enough when sleep_time is too small" or if it's a "python floating point decimals is a bit fiddly" problem.

SR4ven:
This failure is unrelated to your changes and can be ignored.
It was addressed in #442
Once this PR is merged, we can update #422 so it's current again.

erik-smit:
Cool. Any idea what to do next?

erik-smit:
I fixed `s_mirror()`. I have to say I don't really understand the test_case vs. mutation_context stuff, but I just looked at how it's used in the other functions and took their style and now the test doesn't break anymore.

tox.ini
```diff
@@ -3,6 +3,9 @@ minversion=2.0
 skip_missing_interpreters = True
 envlist = py{27,35,36,37,38}-{Linux,Windows,macOS}, docs, lint
 
+[pytest]
+python_classes=NoThanks
```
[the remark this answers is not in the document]
jtpereyda: @SR4ven Rename is fine if it's still needed. All I can think of atm is `FuzzCaseSessionReference`.

unit_tests/test_bit_field_original_value.py
jtpereyda: There's also the `str_to_bytes` function from helpers.py?

jtpereyda:
Nicely done! @SR4ven the PR says there is still "1 change requested" from you. Is that still applicable? (I can't tell exactly which change it is)

SR4ven:
@jtpereyda the changes I requested are done I think.
I just reviewed briefly but I guess thats ok for now as it doesn't go straight into master.

PR #455, opened 2020-08-05
checksum: normalize block_name resolving

erik-smit:
`s_checksum` syntax had changed to use a full path name for `block_name`, in contrast to `s_size`, `s_repeat`, etc.
changed to their behaviour

PR #457, opened 2020-08-06
Minor mutations-iterator fixes

SR4ven:
While looking at the changes, I came across some inconsistencies.

The tests will fail because there seems to be a problem with absolute name resolving that I described in https://github.com/jtpereyda/boofuzz/pull/455#issuecomment-670221341.

### DIFF
--- a/boofuzz/blocks/checksum.py
+++ b/boofuzz/blocks/checksum.py
@@ -190,11 +190,3 @@ class Checksum(primitives.BasePrimitive):
 
     def __len__(self):
         return self._length
-
-    def __bool__(self):
-        """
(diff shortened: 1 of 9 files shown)


SR4ven:
Waiting for #458 to be merged

jtpereyda:
Pending discussion in #458.

SR4ven:
A part of this PR switches to the new name resolving so we consistently use it everywhere.
Some tests ware failing before #458 was merged, that's why I left the message above.

However, right now we should be independent of #458. We can merge this one and if we change the name resolving, we fix the tests at the same time.

jtpereyda (approved):

PR #464, opened 2020-08-17
Added a simple TFTP fuzzer 

Fans0n-Fan:
Added a simple TFTP fuzzer to demonstrate boofuzz usage.


jtpereyda:
Thanks for the example @Fans0n-Fan ! Thanks for the merge @SR4ven .

There is or should be a list in the docs we can add this to...

PR #465, opened 2020-08-21
Exclude examples and request_definitions from pytype checks

SR4ven:
Running pytype on example, request_definition and utils files won't work because they aren't part of the boofuzz package, which is why the name resolving fails. This can be observed in #464 

We can simply exclude these directories.


jtpereyda:
Thanks @SR4ven !

PR #467, opened 2020-09-08
Update __init__.py

Zach116:
Fixed s_update. Was missing a "_" for the "value" attribute.


jtpereyda:
Thanks @Zach116 !

This may be fixed by an upcoming PR, but we can merge in the fix for now.

PR #472, opened 2020-10-20
minor fixes: no default text logger; silence target output by default

jtpereyda:


boofuzz/sessions.py
```diff
@@ -470,8 +470,6 @@ def __init__(
         if self.console_gui and os.name != "nt":
             fuzz_loggers.append(fuzz_logger_curses.FuzzLoggerCurses(web_port=self.web_port))
             self._keep_web_open = False
-        if len(fuzz_loggers) == 0:
-            fuzz_loggers = [fuzz_logger_text.FuzzLoggerText()]
```
SR4ven, line 474: I like the idea of being able to run boofuzz quietly, but this would also be a breaking change I guess.
Imagine people updating and suddenly their scripts don't output anything anymore, wondering why.

We already have too many arguments in session but wouldn't it be better to have a "quiet" argument instead?

Or a verbosity level maybe? So that one can choose to see only errors and the most basic info.
I was working on a logger using https://github.com/willmcgugan/rich a while ago but couldn't get it working properly. I'll take a look at it again when I find the time.

jtpereyda: That is a good point.

jtpereyda:
This was implemented in a later PR; passing an empty list is now a way of explicitly saying not to create a logger.

Issue #468, opened 2020-09-30
Issue using network monitor and probably callback monitor


Eki474:
### Summary

I tried to setup a network monitor to capture packets for each test case as described in the documentation. When setting up a network monitor, the first RPC command sent is not implemented in NetworkMonitor.

### Steps to reproduce

I used LightFTP server implementation to test but I believe any server would do.

* Activate python virtual environment
* Start FTP server
* Start network monitor server (network_monitor_server_example.py down below)
* Start fuzzing example (network_monitor_example.py down below)

### Expected results

* Fuzzing script connects to network monitor server
* Network monitor starts capturing packets for each test case

### Actual results

* Fuzzing script connects to network monitor server
* Fuzzing script sends `set_crash_filename` function call
* This function call is not implemented, the server crashes

### Findings

From what I understand every monitor implementation shall implements BaseMonitor to be a monitor. This interface does not force support for `set_crash_filename` function call and this function is only implemented in ProcessMonitor.

I found out that this option is systematically called by the session on every monitor regardless of their type ([here in code](https://github.com/jtpereyda/boofuzz/blob/master/boofuzz/sessions.py#L528)).

Is this function call necessary for all monitor types and shall be implemented in each one of them ? Or shall this function only be called on ProcessMonitor ?

I just started working on boofuzz, so maybe I misunderstood some behaviours in those findings. Thanks in advance for the help.

Regards,

### Network monitor server logs

```
ERR> USAGE: network_monitor.py
[-d|--device DEVICE #]    device to sniff on (see list below)
[-f|--filter PCAP FILTER] BPF filter string
[-P|--log_path PATH]      log directory to store pcaps to
[-l|--log_level LEVEL]    log level: default 1, increase for more verbosity
[--port PORT]             TCP port to bind this agent to
Network Device List:
[0] eth0    192.168.222.165
[1] any
[2] lo      127.0.0.1
[3] bond0
[4] dummy0
[5] sit0
[6] nflog
[7] nfqueue
[03:31.19] Network Monitor PED-RPC server initialized:
[03:31.19]       device:    None
[03:31.19]       filter:
[03:31.19]       log path:  ./
[03:31.19]       log_level: 1
[03:31.19] Awaiting requests...
PED-RPC> remote method "set_crash_filename" of <__main__.NetworkMonitorPedrpcServer object at 0x7fa6650c3940> cannot be found
Exception in thread Thread-1:
Traceback (most recent call last):
[...]

PR #474, opened 2020-11-12
Fix data_bytes calculation (wait for PcapThread)

Eki474:
Problem:
- `NetworkMonitorPedrpcServer` does not wait for `PcapThread` packet handling to end.
- `data_bytes` is incorrect because post_send is triggered before packet handling ends.
- Sometimes `data_bytes` could even be 0, resulting in boofuzz detecting a crash (see `sessions.py` line 937).

https://github.com/jtpereyda/boofuzz/blob/0c03ee04817fae24b0f42f4dedcb09bff00def96/boofuzz/sessions.py#L936-L937

Fix:
- `NetworkMonitorPedrpcServer` now waits for `PcapThread` packet handling to end.
- `data_bytes` is now correct

Note: Also added Linux support

See #468 for context

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Features
 - Protocol definition: Protocols can now be defined with an object oriented rather than static approach.
 - Independent mutation and encoding steps: Will enable multiple mutations and code coverage feedback.
 - Procmon: Additional debug steps. Partial backwards compatibility for old interface.
+- Network monitor: improved network interface discovery (Linux support)
 
 
 Fixes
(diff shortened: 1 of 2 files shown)


Eki474:
Hi,

I think the test crashed on its own and I can not restart it.

Regards,

SR4ven:
I retriggered the tests, it seems like one of them got stuck.

SR4ven (approved):
Looks good to me.

@mistressofjellyfish do you still want to take a look at it?

SR4ven:
One more thing, please add two lines about the Linux support and the fix to CHANGELOG.rst

mistressofjellyfish:
LGTM!

Eki474:
I updated the changelog file.

SR4ven:
Thank you @Eki474 for the fix and @mistressofjellyfish for the review!

jtpereyda:
Thanks @Eki474 @mistressofjellyfish @SR4ven !

PR #479, opened 2020-12-10
Add UnixSocketConnection

schaeferpp:
I needed the ability to fuzz unix sockets which is why I created this patch. Perhaps someone else needs this as well.


jtpereyda:
Thanks @schaeferpp ! I'm sure this one will see some use out there.

The docs build failed so I restarted it; hopefully just a blip (I noticed an HTTP 522 error so it may have been a transient server error).

jtpereyda:
Looking good. @schaeferpp would you mind adding a line to the CHANGELOG for this?

schaeferpp:
Here you go!

PR #483, opened 2021-01-06
Local procmon

jtpereyda:
Local process monitor so as to run targets without a separate procmon command.

The implementation suggests some refactors:

1. Use local procmon as the main procmon implementation. Rewrite the ProcessMonitorPedrpcServer to just have-a ProcessMonitorLocal.
2. Clarify the differences between the server and client side process monitors. I think the method args can be identical -- right now they're mostly identical with subtle differences.
3. Have the local procmon, the client procmon, and the server procmon all inherit from the procmon base class.

However it might be best to merge as is for the purposes of moving forward.

Also slipped in a change for loggers: If you pass no loggers, the Session still creates a default for you. If you pass an empty list however, you're saying you want no loggers, and the console output will be quiet.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,7 @@ Fixes
 - Fewer duplicates from `Group` primitives.
 - Network monitor: fixed data_bytes calculation and PcapThread synchronization
 - Fixed a crash when using the network monitor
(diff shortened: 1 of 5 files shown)


SR4ven (approved):
Nice work, I like this!

Maybe open an issue for the suggested refactors.

PR #485, opened 2021-01-15
Protocol Session Metadata

jtpereyda:
Added references to the "current" and "previous" message in message-specific (rather, transition-specific) callbacks. The interface is somewhat experimental still. See https://github.com/jtpereyda/boofuzz-ftp/pull/3 for usage example.


SR4ven (approved):
Looks good, surely a nice addition!

PR #486, opened 2021-01-23
Update block and primitive documentation

SR4ven:
I noticed that parts of the new protocol definition documentation is incomplete, so I completely reworked that part.
The static protocol definition documentation remains mostly the same, to prevent breaking changes.

There's one thing I'm uncertain about: Should each primitives docstring and signature contain the `name` and `fuzzable` parameter?
Technically this is not needed because we pass kwargs down to `Fuzzable`, however I think it would make it more clear for the user which parameter a primitive/block can take.

So on top of this change I'd suggest adding the `fuzzable` parameter after `name`, just like we have it on most static definitions.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Features
 - Network monitor: improved network interface discovery (Linux support)
 - Add support for fuzzing Unix sockets with the `UnixSocketConnection` class.
 - Add metadata to ProtocolSession to support callbacks -- `current_message`, `previous_message`.
+- All primitive arguments are now optional keyword arguments.
 
 Fixes
 ^^^^^
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -310,22 +310,16 @@ def s_block(name, group=None, encoder=None, dep=None, dep_value=None, dep_values
     return ScopedBlock(block)
 
 
-def s_aligned(name, modulus, pattern=b"\x00"):
-    """
-    Open a new block under the current request. The returned instance supports the "with" interface so it will
-    be automatically closed for you::
+def s_aligned(modulus, pattern=b"\x00", name=None):
+    """FuzzableBlock that aligns its contents to a certain number of bytes
 
-        with s_block("header"):
-            s_static("\\x00\\x01")
-            if s_block_start("body"):
-                ...
-
-    :type  name:        str
-    :param name:        Name of block being opened
     :type  modulus:     int
     :param modulus:     Pad length of child content to this many bytes
     :type  pattern:     bytes
     :param pattern:     Pad using these byte(s)
+    :type  name:        str, optional
+    :param name:        Name, for referencing later. Names should always be provided, but if not, a default name will
+                        be given, defaults to None
     """
 
     class ScopedAligned(object):
@@ -364,7 +358,7 @@ def s_block_start(name, *args, **kwargs):
     :note Prefer using s_block to this function directly
     :see s_block
     """
-    block = Block(name, *args, **kwargs)
+    block = Block(name=name, *args, **kwargs)
     blocks.CURRENT.push(block)
 
     return block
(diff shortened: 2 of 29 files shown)


SR4ven:
The failing tests made me aware of one more issue.
When I made the `name` parameter optional for primitives, I moved them to the end of the parameter list.
In the examples and unit tests, positional arguments were used for `name` and `default_value`. This doesn't work as it did before now.
The `default_value` is still positional, so it has to come first and as `name` has to be passed with its keyword like `name=name`.
E.g. 
```python
Bytes(name=b"some-bytes", default_value=b"abc", max_len=3)
Bytes(b"abc", name=b"some-bytes", max_len=3)
```

Even though the keywords add some overhead to the protocol definition, I think it might make the examples or the code in general more readable. And for those who want to go the minimalistic way, it's still possible by defining just the `default_value`.
What do you think @jtpereyda?

jtpereyda:
@SR4ven In the old setup, default_value was usually first and positional, but then in some cases like s_block that didn't have a default value, name was the first argument, and not optional. This subtle difference made for a counterintuitive read if you were reading an existing protocol definition.

So the strategy I took with the new approach was to 1) Make name always come first and 2) Make name always be mandatory and the first positional. The idea was that this would make reading a little more consistent.

However, you bring up a good point -- it was convenient to be able to skip name in the old approach.

Now, we could make name optional for all primitives and blocks, including Block. Fuzzable already has a default name in the `name` property (Blocks are less useful without names, but they can still work). HOWEVER, if we leave required arguments like `default_value` positional, then protocol definitions with names have an inconsistent ordering.

jtpereyda:
Is there any precedent in Python libraries for mandatory keyword arguments? On second thought -- we could always assign a default to default_value, making all arguments optional -- e.g., 0 for numeric types, empty string for stringy types, etc.

That way:
1. Everything is optional for people that want to YOLO it.
2. The explicit-is-better-than-implicit crowd can always use keyword arguments.
3. We can implement a consistent ordering for those that want to utilize the brevity of positional arguments.

When writing definitions, I fit into either camp 2 or camp 3. I always name things, and generally prefer keyword arguments, though I might make an exception for names if they _always_ come first.

Of course if there are arguments that are truly mandatory, that would mess with the scheme a bit. I suppose Mirror, Length, Checksum don't make a lot of sense without a reference to a Block. Though we could always make them render as null/empty by default.

jtpereyda:
My first choice is to go for 100% optional keyword args, but tor order args with name always first. But that may take a bit of work to make referential types like Mirror and Checksum support a sort of default null value.

jtpereyda:
By the way, can we keep the Google style docstrings? https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

SR4ven:
I like the idea of all optional keywords args, let's go for that.
Name always fist seems reasonable as it is the only argument all blocks and primitives have. As the second I'd go for the default value, if it exists, or a reference block.

I'll see if there is an easy way to render blocks which require a reference empty by default.

How should we handle the existing s_... blocks and primitives? Just make all arguments optional or sort them to reflect the same signature as the underlying class? The later would be a breaking change so maybe we should not do that.

I chose the reStructuredText Docstring Format and reformatted all those docstrings by hand because [PEP 287](https://www.python.org/dev/peps/pep-0287/) recommends it.
But now that you say it I should have probably asked which format we're using. No matter which one we agree on, we should use it consistently.
There is [pyment](https://github.com/dadadel/pyment) for automatic conversion but when I tested it a while ago I had some problems (https://github.com/dadadel/pyment/issues/83)

jtpereyda:
For `s_` blocks, let's keep parameters in the same order for compatibility. Feel free to make args optional if it's easy, as that still allows calls with the old positional approach.

I had started slowly moving to Google style docstrings, but a tool to do it all at once would be perfect. I should have thought to look for something like that. The Google style is a lot less cluttered, especially when there's quite a few arguments.

SR4ven:
Should we add the `fuzzable` arg to the docstring and the signature? It's not required as it can be passed via kwargs, but it might be more obvious to the user that all primitives have this argument if it appears in the docstring and signature.

About the docstrings, the Google style is indeed easier to read. But then again, I have to say that I hardly look at the docstrings themself. I either see them already formatted in the IDE or in our docs. Also, Sphinx doesn't seem to be able to parse Google style docstrings out of the box and requires a preprocessor extension.

The only tool for automated style conversion seems to be pyment which hasn't been updated in a while. I've tested the conversion from reST to Google style which works somewhat ok. However the type informations gets lost and it doesn't seem to support "Yields".
I guess I'll leave the format as it is now and we try to convert all at once at another time.

jtpereyda:
@SR4ven sure, adding fuzzable sounds helpful.

SR4ven:
Alright, are we good to merge then?

jtpereyda:
@SR4ven Yes, let's just add a CHANGELOG entry to the effect of "all primitive arguments are now optional keyword arguments" and merge it. Thanks for the documentation work here!

SR4ven:
Good point, almost forgot about the changelog.

PR #488, opened 2021-02-04
Fix max_size of string primitive

### LINKED ISSUE #487: max_len invalid in s_string() ？
Test script:
```py
from boofuzz import *

def main():
    session = Session(
        target=Target(
            connection=SocketConnection("127.0.0.1", 1234, proto='tcp')
        ),
    )

    s_initialize(name="test")

    s_string("abcd", max_len=100)
    session.connect(s_get("test"))

    session.fuzz()


if __name__ == "__main__":
    main()
```

length of generated data should within 100, but i got 500000:
```
[2021-01-31 20:41:56,971] Test Case: 1292: test:test.String1:1292
[2021-01-31 20:41:56,971]     Info: Type: String. Default value: b''. Case 1292 of 1482 overall.
[2021-01-31 20:41:56,971]     Info: Opening target connection (192.168.1.5:1234)...
[2021-01-31 20:41:56,971]     Info: Connection opened.
[2021-01-31 20:41:56,971]   Test Step: Monitor CallbackMonitor#139659426646224[pre=[],post=[],restart=[],post_start_target=[]].pre_send()
[2021-01-31 20:41:56,971]   Test Step: Fuzzing Node 'test'
[2021-01-31 20:41:57,064]     Info: Sending 500000 bytes...
```

pkts in wireshark:
截屏2021-01-31 下午8 49 03

Is this a bug?

SR4ven:
@kuraraneko can you verify that this fixes #487?

### DIFF
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -685,7 +685,7 @@ def s_mirror(primitive_name, name=None):
     blocks.CURRENT.push(Mirror(name, primitive_name, blocks.CURRENT, fuzzable=True))
 
 
-def s_string(value, size=-1, padding=b"\x00", encoding="ascii", fuzzable=True, max_len=0, name=None):
+def s_string(value, size=-1, padding=b"\x00", encoding="ascii", fuzzable=True, max_len=-1, name=None):
     """
     Push a string onto the current block stack.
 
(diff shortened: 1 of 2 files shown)


SR4ven:
Currently we only yield strings which are within the defined max_len.
In the bytes primitive we yield all values and truncate them to the max_len.

Should we unify this behaviour @jtpereyda?

kuraraneko:
The generated data of s_string() seem to be normal, I think the bug has been fixed.

jtpereyda:
@SR4ven Truncate would be nice. If it's too tricky though we can fall back on skipping long strings. Naive truncation would result in some duplicates, so that's at least one tricky detail.

jtpereyda:
Code looks good to merge for now. Feel free to implement truncation now/later/never at your preference. :)

Thanks @kuraraneko for the report and @SR4ven for the fix!

SR4ven:
Sounds good, I'll merge now and implement the truncation later.
Still need to figure out a way to prevent duplicates

PR #491, opened 2021-02-05
[WIP] WebUI: vuejs

mistressofjellyfish:
So, here we go.

Current state: I've done a mockup that doesn't talk to the API, but is somewhat functioning (that is: if it's a button, it probably does something).

Web:
Bildschirmfoto 2021-02-06 um 00 13 09

Mobile:
2021-02-06 00 12 40 localhost 2b30e53c7a61

Feel free to look at the code. I placed the whole boofuzz-web stuff in a new folder. The idea is that the files at the current location will be replaced when something like `npm run dist` is executed, however that isn't implemented for now.


mistressofjellyfish:
I don't really know what's up with the tests, but I haven't fixed them yet. As you can see, I re-arranged some UI components, however I kept the overall layout. Test case log is missing ATM.

jtpereyda:
Nice! I like cleaner but still mildy hacker-esque look. The Dashboard/Testcases navigation makes sense.

PR #494, opened 2021-02-19
Fix Python 3.9 compability

youduda:
This replaces threading.Thread.isAlive with is_alive (available since Python 2.6) in the process monitor rpc calls because isAlive has been dropped in Python 3.9.


SR4ven:
Thank you for the PR @youduda!

Please also add a line about this to the CHANGELOG.rst so we can merge it.

PR #501, opened 2021-04-01
Custom Checksum example

ptdropper:
Custom checksum function using CRC16.
Example usage in the examples folder.
Clarified Checksum() class documentation.


boofuzz/blocks/checksum.py
```diff
@@ -28,6 +28,9 @@ class Checksum(primitives.BasePrimitive):
     the algorithm parameter.
 
     The length field is only necessary for custom algorithms.
+    When using your own provided custom checksum function return is the calculated crc of the data.
+
+    Function signature:  <function_name>(data_bytes, starting_value in hex. Returns a number.
```
SR4ven, line 33: A custom checksum function will only be called with one argument, the data in bytes type. The function then needs to return the checksum also as a bytes type, not a number/integer.
Also there is a parentheses missing.

examples/autoprog.py
```diff
@@ -0,0 +1,196 @@
...
+        name="CRC16_outer", block_name="autoprogB", request=reqW, algorithm=getCrc16Ccitt, length=16, fuzzable=False
+    )
+
+    block.push(dl_line_1)
+    reqW.push(crcValue)
+    block.push(dl_line_9)
+    block.push(dl_line_11)
+    reqW.push(crcValue_outer)
+    block.push(dl_line_12)
+    reqW.pop()
```
SR4ven, line 190: Why use push/pop when we could define the protocol from top to bottom with children like it's done in the http_simple example?
https://github.com/jtpereyda/boofuzz/blob/71f69bf829f98114100d3c7815caa5e827980d1a/examples/http_simple.py#L22-L30

That would make it a bit easier to read I think.

examples/autoprog.py
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env python3
+# Designed for use with boofuzz v0.2.0
```
SR4ven, line 2: v0.3.0

examples/crc16ccitt.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
+# Designed for use with boofuzz v0.2.0
```
SR4ven, line 2: v0.3.0

examples/autoprog.py
```diff
@@ -0,0 +1,196 @@
...
+    dl_line_9 = String(name="crc_end", default_value="</Checksum>", fuzzable=False)
+    dl_line_11 = String(name="outer_crc", default_value="<ChecksumTotal>", fuzzable=False)
+    # insert outer crc here
+    dl_line_12 = String(name="outer_crcend", default_value="</ChecksumTotal>", fuzzable=False)
+
+    reqW = Request("autoprog")
+    block = Block(name="autoprogB", request=reqW)
+    reqW.push(block)
+    crcValue = Checksum(
+        name="firstCRC16", block_name="autoprogB", request=reqW, algorithm=getCrc16Ccitt, length=16, fuzzable=False
```
SR4ven, line 178: The `length` argument expects the number of bytes not bits if I remember correctly. We should perhaps clarify this in the documentation.

So it should be 2 instead of 16 here I guess.

SR4ven (changes requested):
Thank you for the PR @ptdropper!
Sorry for the delayed review.

We've had a few questions about custom checksums in the past, so thanks for adding an example.
I've added some comments below.

boofuzz/blocks/checksum.py
```diff
@@ -38,11 +41,11 @@ class Checksum(primitives.BasePrimitive):
...
     :type  request: boofuzz.Request, optional
-    :param request: Request this block belongs to.
-    :type  algorithm: str, function, optional
-    :param algorithm: Checksum algorithm to use. (crc32, crc32c, adler32, md5, sha1, ipv4, udp)
-        Pass a function to use a custom algorithm. This function has to take and return byte-type data,
-        defaults to crc32
+    :param request: Request this block belongs to
+    :type  algorithm: str, function def name, optional
+    :param algorithm: Checksum algorithm to use from this list, default is crc32 (crc32, crc32c, adler32, md5, sha1,
+        ipv4, udp). See above for custom checksum function.
```
jtpereyda, line 47: How about:
> ... custom checksum function example.

examples/autoprog.py
jtpereyda: I like that approach better as well. @ptdropper this change can be optional for this PR.

jtpereyda (changes requested):
@ptdropper Thanks for the PR! If you can make the changes requested here then we can get this merged in.

jtpereyda:
@SR4ven thanks for the review!

examples/autoprog.py
ptdropper: This is the style I have used in my example and in my working code. I found the other style difficult to use correctly. My example works so I am keeping the style.

examples/autoprog.py
ptdropper: Understood and updated.

examples/crc16ccitt.py
ptdropper: Understood and updated.

examples/autoprog.py
ptdropper: This length field confused me. I changed it to the suggested value of 2. My code does not use it.

boofuzz/blocks/checksum.py
ptdropper: Understood and updated.

ptdropper:
CHANGELOG.rst failed the docs build due to missing white space.

SR4ven (changes requested):
Thanks for applying the changes @ptdropper!
Two more things

SR4ven (approved):

SR4ven:
Thanks again for the fixes.

The failing test are unrelated and will be fixed in #506.
Good to go from my side.

boofuzz/blocks/checksum.py
```diff
@@ -28,6 +28,9 @@ class Checksum(primitives.BasePrimitive):
     the algorithm parameter.
 
     The length field is only necessary for custom algorithms.
+    When using your own provided custom checksum function return is the calculated crc of the data.
```
jtpereyda, line 31: Suggested rewording:
```
When using your own provided custom checksum function, the return value should be the calculated checksum of the data.
```

jtpereyda (changes requested):
@ptdropper One more rewording suggestion. Looks like we're good to go after that.

boofuzz/blocks/checksum.py
ptdropper: Updated the wording and simplified it slightly to fit the line length limit of 120 characters.

jtpereyda (approved):

jtpereyda:
Merging in. Thanks again @ptdropper !

jtpereyda:
@ptdropper thanks! Nothing comes to mind right now. If you encounter potential room for improvement, feel free to open an issue for discussion or even a PR.

PR #505, opened 2021-04-10
Correctly set l2_dst arg type to bytes

SR4ven:
Passing the l2_dest e.g. MAC address as bytes seems to be more suitable here and prevents issues with encoding.

The default value is already a bytes type, this PR just adapts the documentation.

Please correct me if it should be a string typed value.


jtpereyda:
Good fix, thank you. I'm guessing these test failures are transient since this a is a docs-only change.

jtpereyda:
Seems to be an issue on the `choco install vcpython27` step...

SR4ven:
Seems like Microsoft removed the installer msi. The maintainers want to remove the package https://community.chocolatey.org/packages/vcpython27/9.0.0.30729#discussion.
We'll have to stop testing Python 2.7 on Windows I guess. The whole Python 2 support is going to be more and more difficult it seems.
But we should continue this in #370

PR #506, opened 2021-04-26
Remove Python 2 support

SR4ven:
Hotfix for failing Python 2.7 CI tests according to #370.

I simply replaced the pydbg parts in process_monitor.py with the debugger_thread_simple which we also use on Linux.
No testing done so far.

Will the network_monitor.py still work with Python 3? I have never used it.

While I was at it, I enabled testing for Python 3.9.


jtpereyda:
I think we can patch it, but I have other changes to the same file pending in another branch. I'll try and ready that branch to merge first.

Without pydbg, we'll lose stack traces when Windows processes crash. Oh well.

SR4ven:
Sounds good. Shall we merge this one first so the CI tests work again?

jtpereyda:
Let's wait for now and see if we can't bundle everything up, release, then do a separate release to drop Python 2 and pydbg support.

jtpereyda:
Well, I gave up on Python 2.7 compatibility in that other PR, so we can go ahead and merge this one for now. I'll make a separate PR for my Windows fixes to the simple debugger.

PR #507, opened 2021-04-29
Make impacket dependency optional

SR4ven:
The only place where impacket is needed is in the network_monitor. In the same file we also import pcapy, which is not listed as a dependency while impacket is.

To reduce the number of dependencies and so the number of possible problems and conflicts (#249), I suggest to remove it as a dependency. Whoever wants to use the network monitor will have to install impacket manually.

Another argument for removing the dependency is that the network monitor is more of an optional tool that comes with boofuzz rather than a part of the boofuzz packet itself.

I'm picking this up because I just had a case where people, let's say, were suspicious about the presence of impacket on their systems.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -178,10 +178,10 @@ and has been replaced with boofuzz’s logging mechanisms.
 However, some people still prefer the PCAP approach.
 
 .. note::
-    The network monitor requires Pcapy, which will not be automatically installed with boofuzz. You can manually
-    install it with ``pip install pcapy``.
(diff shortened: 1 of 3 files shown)


ptdropper (approved):

jtpereyda:
Makes sense. It would make sense to add impacket to the network monitor install instructions then, right? https://github.com/jtpereyda/boofuzz/blob/master/INSTALL.rst

jtpereyda:
Merging in. Thanks again @ptdropper !

SR4ven:
Good point, will add

jtpereyda:
LGTM. Thanks @SR4ven !

PR #508, opened 2021-04-30
New CLI arguments and experimental code coverage

jtpereyda:
Lots of changes bundled together. This was mostly for the experimental code coverage, but I made lots of fixes along the way that are also bundled.

1. New CLI arguments:
    1. --stdout: hide/capture/mirror stdout and stderr
    1. --qemu, --qemu-path: experimental code coverage feedback
    1. --web-port: web GUI port
    1. --restart-interval: restart every n test cases
    1. --target-start-wait: adjustable process startup wait time (default 0). Also removed duplicate wait times sprinkled throughout the code.
1. Experimental code coverage using afl-qemu-trace
    1. Includes some experimental web UI stats
1. TCP connections:
    1. Safer shutdown approach (shutdown, recv, close) -- should improve stability 
1. edge.py: Human-readable id instead of integer id.
1. Session:
    1. Add register_post_start_target_callback(): callback after a target is started or restarted.
    1. Code coverage support
    1. stop target even when exceptions are raised
    1. keep web interface open even when exceptions are raised


jtpereyda (commented):
While working with boofuzz, I realized that the past design choice to return `b""` on timeout, combined with the fact that the underlying socket API has `recv` return 0 to indicated a closed connection, leads to an ambiguity: return value of `b""` can mean a closed connection, or a timeout.

My solution was to throw an exception to indicate the socket has shut down, which is more intuitive to me personally. However, this actually flips the semantics of `recv`, which throws socket.timeout on timeout and returns `b""` on socket shutdown. Perhaps it would be wiser to follow this approach to map better to peoples' existing knowledge.

Either way, we have some choice to make:
1. Keep these changes (throw exception on socket shutdown), which removes the ambiguity but can break existing fuzzers that depend on socket shutdowns yielding an empty bytes array.
2. Flip the approach, that is, allow socket.timeout to be bubbled up as an exception. This removes the ambiguity, but breaks existing fuzzers that depend on on timeouts yielding an empty bytes array.
3. Keep the existing behavior, and add a new way to detect whether the socket was shutdown or whether a timeout happened. This is pretty lame but backwards compatible.

boofuzz/connections/tcp_socket_connection.py
```diff
@@ -84,18 +98,25 @@ def _connect_socket(self):
...
+            BoofuzzTargetConnectionShutdown: Target shutdown connection (e.g. socket recv returns 0 bytes)
+            BoofuzzTargetConnectionAborted: ECONNABORTED
+            BoofuzzTargetConnectionReset: ECONNRESET, ENETRESET, ETIMEDOUT
         """
         data = b""
 
         try:
             data = self._sock.recv(max_bytes)
+            if len(data) == 0:
+                raise exception.BoofuzzTargetConnectionShutdown()
```
jtpereyda, line 119: 1. New CLI arguments:
    1. --stdout: hide/capture/mirror stdout and stderr
    1. --qemu, --qemu-path: experimental code coverage feedback
    1. --web-port: web GUI port
    1. --restart-interval: restart every n test cases
    1. --target-start-wait: adjustable process startup wait time (default 0). Also removed duplicate wait times sprinkled throughout the code.
1. Experimental code coverage using afl-qemu-trace
    1. Includes some experimental web UI stats
1. TCP connections:
    1. Safer shutdown approach (shutdown, recv, close) -- should improve stability 
1. edge.py: Human-readable id instead of integer id.
1. Session:
    1. Add register_post_start_target_callback(): callback after a target is started or restarted.
    1. Code coverage support
    1. stop target even when exceptions are raised
    1. keep web interface open even when exceptions are raised

jtpereyda: On the other hand, I don't know how many scripts out there would actually hit this compatibility problem.

jtpereyda: Perhaps one tool to compare with is pwntools, which also has a recv method for which timeout returns an empty bytes, and a closed connection yields an exception: https://docs.pwntools.com/en/stable/tubes.html#pwnlib.tubes.tube.tube.recv

SR4ven: Hmm difficult choice.
I don't like the existing behaviour because we can't differentiate a close from a timeout. So I'd say either stick to the default socket/posix behaviour or raise an exception for everything.

The default posix way would be choice 2 if I understand correctly. Raise socket.timeout and return b"" on close.

However, as we are on application level I could also imagine raising an exception for both cases. Not sure what others think about this but it might be the most user friendly. You simply tell boofuzz to receive after a test case and if anything goes wrong (close/timeout/abort/reset) you get an exception.
On the other hand, I know some of my targets close the connection if the received data was invalid. In that case the target didn't crash but boofuzz would raise an exception. That might be a bit inconvenient.

I also just found the session option `check_data_received_each_request`. It seems this option is the reason for both close and timeout returning b"". If we switch to exceptions, we'd have to catch them here and depending on the setup maybe re-raise?

Right now I favour choice 2 as anyone with basic knowledge about sockets will understand what's going on. Also it looks like that approach could easily be adapted to the current `check_data_received_each_request` behaviour.

BTW how can we get EWOULDBLOCK if we use a blocking socket and catch socket.timeout before?
https://github.com/jtpereyda/boofuzz/blob/68ad063c72e251a2c840d47c74cf457bc748e6a8/boofuzz/connections/tcp_socket_connection.py#L131

And why do we interpret ETIMEDOUT as BoofuzzTargetConnectionReset?
https://github.com/jtpereyda/boofuzz/blob/68ad063c72e251a2c840d47c74cf457bc748e6a8/boofuzz/connections/tcp_socket_connection.py#L129

SR4ven (commented):
Nice big change! Looks good so far but I haven't done any testing yet. Do you have an example use case or some documentation on how to use the code coverage feature?

boofuzz/connections/tcp_socket_connection.py
jtpereyda: > I also just found the session option check_data_received_each_request. It seems this option is the reason for both close and timeout returning b"". If we switch to exceptions, we'd have to catch them here and depending on the setup maybe re-raise?

Yes, that Session option is a layer on top of the socket behavior. The only place in my scripts this would matter is in callbacks, where I sometimes have code set up to receive the next (typically valid) message. Whatever choice we make here, we can modify Session to still act the same way with `check_data_received_each_request`.

boofuzz/connections/tcp_socket_connection.py
```diff
@@ -21,18 +21,32 @@ class TCPSocketConnection(base_socket_connection.BaseSocketConnection):
...
         self._serverSock = None
+        self.graceful_shutdown = graceful_shutdown
 
     def close(self):
+        if self.graceful_shutdown:
+            try:
+                self._sock.shutdown(socket.SHUT_RDWR)
+                while len(self._sock.recv(1024)) > 0:
+                    pass
+            except ConnectionError:
```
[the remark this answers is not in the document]
jtpereyda: ConnectionError is a built-in exception, but only for Python 3! haha

boofuzz/connections/tcp_socket_connection.py
jtpereyda: EWOULDBLOCK is equivalent to EAGAIN, which can be raised when there is no data available to read, which is why it's treated similarly to a timeout. https://stackoverflow.com/questions/4058368/what-does-eagain-mean

SR4ven: >Whatever choice we make here, we can modify Session to still act the same way with check_data_received_each_request.

Sounds good.

True about EWOULDBLOCK, but it only applies to non-blocking sockets. We use blocking ones and catch socket.timeout so to my understanding EWOULDBLOCK or EAGAIN should never come up.

boofuzz/connections/tcp_socket_connection.py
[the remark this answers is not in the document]
SR4ven: Ok maybe we have to tell flake8 to only look at Python 3 then

boofuzz/connections/tcp_socket_connection.py
jtpereyda: ETIMEDOUT is obscure but I'm pretty sure I actually encountered it. Here's [what one site says](https://www.davepacheco.net/blog/2016/tcp-puzzlers/#:~:text=ETIMEDOUT%20is%20a%20socket%20error,some%20packet%20for%20too%20long.):

> ETIMEDOUT is a socket error you can see from read(), write(), and other operations that indicates that some timeout associated with the connection has elapsed. The cases I’ve seen most are when the remote side did not acknowledge some packet for too long. These are usually either data packets, a FIN packet, or a KeepAlive probe.

Re EWOULDBLOCK

> True about EWOULDBLOCK, but it only applies to non-blocking sockets. We use blocking ones and catch socket.timeout so to my understanding EWOULDBLOCK or EAGAIN should never come up.

I think that's correct. I'm not sure exactly why I put this in, but if we ever had an option to switch to nonblocking this would catch it. Maybe I just found them and identified them as being in the same error category.

jtpereyda: Actually aren't they nonblocking sockets since we set timeouts?

SR4ven: Ok good to know. So socket.timeout is not equal to ETIMEDOUT.

Alright, if we use non-blocking we could reuse it. Consequently, EAGAIN should be caught too then. Or just remove both and add them if we actually use non-blocking sockets.

SR4ven: >Actually aren't they nonblocking sockets since we set timeouts?

We'd have to set the socket into non-blocking mode with [socket.setblocking(False)](https://docs.python.org/3/library/socket.html#socket.socket.setblocking)

According to [this](https://docs.python.org/3/library/socket.html#notes-on-socket-timeouts), timeout mode is a 3rd option besides blocking and non-blocking.

jtpereyda: Trying to get all these options clear in my head:

```
 ----------+--------------+-----------------------+--------------+------------+------------
|          | OS socket    | OS socket nonblocking | boo previous | PR initial | PR propsed |
+----------+--------------+-----------------------+--------------+------------+------------+
| timeout  | wait forever | exception             | return b""   | return b"" | exception  |
| shutdown | return b""   | return b""            | return b""   | exception  | return b"" |
 ----------+--------------+-----------------------+--------------+------------+------------
```

Part of me wants to yield an exception in both cases, but I'm also leaning toward matching OS socket behavior.

SR4ven: The table seems to be correct.

Agreed, I feel exactly the same way. I think if you have worked with sockets before, the OS socket behaviour is more intuitive.
It's the other way around if you haven't I guess.

boofuzz/blocks/request.py
```diff
@@ -25,6 +25,7 @@ def __init__(self, name=None, children=None):
         FuzzableBlock.__init__(self, name=name, request=self)
         Node.__init__(self)
         self.label = name  # node label for graph rendering.
+        self._name = name
```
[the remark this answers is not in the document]
jtpereyda: Yes, seems that way.

jtpereyda:
> Nice big change! Looks good so far but I haven't done any testing yet. Do you have an example use case or some documentation on how to use the code coverage feature?

Here's an example using the CLI:
```bash
python /home/jpereyda/code/boofuzz-ftp/ftp.py fuzz --web-port 26000 --target 127.0.0.1:2200 --target-cmd '/home/jpereyda/code/LightFTP/Source/Release/fftp /home/jpereyda/code/LightFTP/Source/Release/fftp.conf' --stdout hide --restart-interval 32000 --target-start-wait 3.5 --qemu ftp --username ubuntu --password ubuntu
```

This is with the boofuzz-ftp cli-main branch: https://github.com/jtpereyda/boofuzz-ftp/tree/cli-main and target LightFTP. The coverage approach should work on Linux against most binaries, anything that can run in Qemu.

The relevant command line args are the `--qemu` switch, and the `--target-cmd` which specifies the target binary.

boofuzz/blocks/request.py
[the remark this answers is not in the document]
jtpereyda: It actually was necessary, but only because I set `id = name` as an alias in Request. I did that to support Nodes now having a string id. But the problem was that `Node.__init__(self)` would set id to None. Which, given the alias, would also set name/_name to None again.

Just pushed: I reversed my two changes in request.py, and now pass the name into the Node initializer: `Node.__init__(self, node_id=name)`

jtpereyda:
TODO:
1. Failing Windows tests

jtpereyda:
@SR4ven Good to merge this one? The remaining test failures should resolve with your PR.

I gave up on the whole 2.7 compatibility thing. :/

boofuzz/connections/tcp_socket_connection.py
SR4ven: Which timeout/shutdown behaviour are we going to use now? The one initially proposed in this PR or the OS like?

SR4ven:
So far so good. I left one more question about which socket behaviour we’re going to use.

I didn’t find time to run the example code yet, but I’ll try to do that today. Shouldn’t stop you from merging.

SR4ven:
There are still some stickler warnings. Some seem to be false positives.

boofuzz/connections/tcp_socket_connection.py
jtpereyda: @SR4ven The OS-like behavior. Just realized I didn't add the timeout exception though...

SR4ven: Alright. We don't need BoofuzzTargetConnectionShutdown anymore do we.

Also, do we need to adapt other connection classes to the new behaviour? UDP maybe?
We could do that in another PR if needed.

PR #509, opened 2021-05-04
Update mdns example

peberlein:
Please make sure I've correctly updated the example - it seems to work, and allowed me to fuzz a mdns server. Tested with python 3.8.6.


jtpereyda:
Thanks @peberlein ! Changes make sense.

Issue #289, opened 2019-06-07
Support for float


sajith74:
Add support for s_float.

lucca-ruhland:
Hi, while using boofuzz I have run into the same problem and created an extension for float and integer values.
You can check it out here: https://github.com/lucca-ruhland/boofuzz-numbers.
Since I think hardly anybody will find my repository, would it be possible to contribute my code to boofuzz?
If any changes needs to be done, so that the new primitives (s_float and s_int) are more in line with this project, i would gladly rework my code.

SR4ven:
Sure, every contribution is welcome! Feel free to open a pull request and we will review the code.

PR #511, opened 2021-05-17
s_float / float support

lucca-ruhland:
Hi, as already mentioned in https://github.com/jtpereyda/boofuzz/issues/289,  I've created a support for floats.
This pull request contains:
- [x]  Float primitive (_boofuzz/primitives/float.py_)
- [x]  s_float function to push Float primitives to current block
- [x] example on how to use s_float
- [x] Doc strings for Float and s_float


SR4ven (changes requested):
Thanks for the contribution @lucca-ruhland!

Looks good on first sight, I'll do some testing tomorrow (hopefully).

I've left one comment below and please also add a line about this to the CHANGELOG.rst

It would also be nice if we could generate IEEE 754 floats. As in generate random integers and convert them into IEEE floats formatted as bytes. Would be helpful in some cases for low level protocols I guess. But that's just a thought, nothing required for this PR.

boofuzz/primitives/float.py
```diff
@@ -0,0 +1,69 @@
...
+
+class Float(Fuzzable):
+    """Primitive that generates random float values within a specific range and with a fixed format.
+
+    :type name: str, optional
+    :param name: Name, for referencing lates. Names should always be provided, but if not a default name
+        defaults to None
+    :type default_value: str
+    :param default_value: Value used when the element is not being fuzzed - should typically represent a valid value.
+    :type size: int, optional
```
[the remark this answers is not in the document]
lucca-ruhland: Sorry, I must have forgotten to change this line.
I'll clean up this section and merge with the current master.
I think, you're right about the IEEE 754 floats, I will add an option to generate them.

SR4ven:
Seem like Python 3.5 has some problems with the syntax. Maybe the type hints or the trailing comma?

@jtpereyda if that prevents us from using current syntax we might want to consider dropping support for Python 3.5. It's out of support since 2nd of October 2020.

lucca-ruhland:
> Seem like Python 3.5 has some problems with the syntax. Maybe the type hints or the trailing comma?
> 
> @jtpereyda if that prevents us from using current syntax we might want to consider dropping support for Python 3.5. It's out of support since 2nd of October 2020.
This weekend I will likely update the pull requests (fixing the discussed issues / add IEEE float option).
If you want I can investigate, why Python 3.5 is failing.

SR4ven:
Sounds good. Would be nice to know what the problem with Python 3.5 is, but don't put too much effort into it.

lucca-ruhland:
> Sounds good. Would be nice to know what the problem with Python 3.5 is, but don't put too much effort into it.

I have found the error.
When formatting the floats, I used a string with "f" prefix, for example f"My string {data}".
This is not supported by python 3.5.
I will fix this later.

lucca-ruhland:
@SR4ven Everything should work not, put Stickler CI still fails. It shows _Failed to run pytype linter. It timed out during execution_.
I'm not sure how to solve this, maybe you could help me with this?

SR4ven:
@lucca-ruhland about Stickler, don't take that failure too serious. Might be an unrelated failure in CI or a problem with the tools. I'll try to run the tests locally and review the code tomorrow.

About the f-strings and type hints, I'm for dropping support for Python 3.5 in favour of those new features. [This survey](https://www.jetbrains.com/lp/python-developers-survey-2020/) states that only about 3% still use Python 3.5 so that shouldn't be a problem. What do you think @jtpereyda?

boofuzz/primitives/float.py
```diff
@@ -0,0 +1,87 @@
...
+
+        return value.encode()
+
+    @staticmethod
+    def __float_to_integer(value):
+        return sum(b << 8 * i for i, b in enumerate(struct.pack("f", value)))
+
+    @staticmethod
+    def __integer_to_binary(value, bit_length=32):
+        return bin(value).replace("0b", "").rjust(bit_length, "0")
```
SR4ven, line 84: Right now this return a string like '11000110001110111000000100000000'. Encoded in bytes that is many 0x30 and 0x31. However, I'd expect something like 0x3f800000 (32 bit big endian IEEE-754 representation of 1.0).
That could be archived by `struct.pack('>f', 1.0)`.

Maybe we should also add an `endian` argument to Float as we have it for other primitives.

Later on we could even add a `Double` primitive for 64bit IEEE doubles.

CHANGELOG.rst
```diff
@@ -18,6 +18,15 @@ Fixes
...
 
+v0.3.1
+------
+Features
+^^^^^^^^
+- Add Float primitive
+- Add s_float to push Float primitives to the current block
+- Add an example for s_float/Float usage
+- Add option to encode Floats as IEEE 754 floating point numbers
+
```
SR4ven, line 29: Please move these features to the `Upcoming` topic. We're not 100% sure what the next release number will be.

SR4ven (changes requested):
Nice work @lucca-ruhland!

Looks good, I left 2 small comments below.

CHANGELOG.rst
lucca-ruhland: Sorry, I haven't thought about that.
I have now moved the features to the upcoming topic.

boofuzz/primitives/float.py
lucca-ruhland: To clarify what the function actually does: It gives you the **bitwise** IEEE-754 presentation of a float.

Example:
```python
import struct


def __float_to_integer(value):
    return sum(b << 8 * i for i, b in enumerate(struct.pack("f", value)))


def __integer_to_binary(value, bit_length=32):
    return bin(value).replace("0b", "").rjust(bit_length, "0")


def encode(value: float):
    value = __float_to_integer(value)
    value = __integer_to_binary(value)
    return value


print(encode(1.0))
# output: 00111111100000000000000000000000
```
This should work, you can check it online, for example with: https://www.h-schmidt.net/FloatConverter/IEEE754.html.
It gives me the same result. The corresponding hex value is, like you said, 0x3f800000.

But if you think that the bitwise representation is not useful, I will change that to hexadecimal, like you already suggested.
I will definitively add an endian argument because otherwise, the correct interpretation of the value will be very platform-independent.

lucca-ruhland: I have now changed the IEEE 754 float encoding to hex, as you have requested.
The Float primitive and s_float function have the endian arguments, which accepts 'big' for big-endian and 'little' for little-endian.

lucca-ruhland: One important thing to mention is, that the encoded IEEE floats do not have the '0x' prefix. Is that a problem?
If that's so, I will change this later.

SR4ven: The example for the bitwise representation above does return the correct value, however the encoding/representation isn't the one we're looking for IMHO.
It returns a string "00111111100000000000000000000000", which after encoding to bytes returns b"00111111100000000000000000000000". But each character in this bytes object represents one byte, not one bit.
We can see that by converting bytes to hex-string:
```python
>>> b"00111111100000000000000000000000".hex()
'3030313131313131313030303030303030303030303030303030303030303030'
```
0x30 is the hex code for ASCII `0` and 0x31 is `1`.
So this bytes object isn't 32 bits long, but 32 bytes.

I haven't worked with IEEE floats in network protocols yet, but some quick research showed that Modbus TCP expects the 32 bit representation that `struct.pack('>f', 1.0)` returns.

> One important thing to mention is, that the encoded IEEE floats do not have the '0x' prefix. Is that a problem?

The 0x prefix is just a hint for hex representation, it's not part of the actual number. So no, we don't have to add that.

With the latest changes, it should now work as expected. Thanks for that.

boofuzz/primitives/float.py
```diff
@@ -0,0 +1,93 @@
...
+    def __convert_to_iee_754(self, value: float):
+        if self.endian == "big":
+            iee_value = struct.pack(">f", value)
+        elif self.endian == "little":
+            iee_value = struct.pack("<f", value)
+        else:
+            error_msg = "Invalid endian argument '%s'. Use 'big' or 'little'." % self.endian
+            raise ValueError(error_msg)
+
+        return iee_value.hex()
```
SR4ven, line 90: struct.pack("<f", value) already returns a bytes object. There is no need to convert it to a hex-string and later back to bytes by value.encode() in line 79.

In case of IEEE 754 encoding we can directly return the result of struct.pack without any further encoding.

SR4ven:
Thanks for working in those review comments @lucca-ruhland!

I've left one more comment and SticklerCI seems to have found another Python 3.5 incompatibility in a type hint. Apart from that everything looks good!

boofuzz/primitives/float.py
lucca-ruhland: Sorry, I hadn't thought much about it when I implemented the function.
Everything should be working now.

SR4ven (approved):

SR4ven:
Thanks again for your contribution @lucca-ruhland!

jtpereyda:
Very cool addition @lucca-ruhland ! Thanks @SR4ven for the review. My apologies for the lack of responses.

PR #513, opened 2021-05-18
Use a local random number generator instead of the global one

alk0ve:
String currently affects the global random module (by calling random.seed(0) ), which may affect other logic in boofuzz or scripts that use it. I modified it to use a local random.Random() instance with seed 0 - it produces the same results without affecting the random module itself.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Features
 Fixes
 ^^^^^
 - Clarified Checksum() data for custom checksum function.
+- String and RandomData primitives now use a local and independent instance of random
 
 v0.3.0
 ------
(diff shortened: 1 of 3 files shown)


SR4ven:
Good point @alk0ve! Thanks for the fix.

I think we should do the same thing for the `RandomData` primitive. It doesn't set a seed yet, but for the sake of reproducability we should set one and use a local random instance.

@alk0ve do you want to apply your fix to RandomData too or shall we fix it another time and merge this PR first?

alk0ve:
@SR4ven good idea, I'll change RandomData too, with a TODO note to add a non-zero seed in the future.

SR4ven:
I forgot to mention that we also need a line in the changelog, so I'll just add that myself.

About the todo comment, we actually want constant random numbers so we get reproducible test cases over multiple runs. So I also changed that comment to the same we have in s_string.

@alk0ve if you agree on my changes we're good to merge.

alk0ve:
Sure, looks good :)

PR #515, opened 2021-05-20
Activate dependabot for Github Actions

SR4ven:
I noticed that we can use dependabot to update our github action workflows.

That saves us the work of updating them manually.

It will open a PR weekly if there are any updates.


jtpereyda:
Cool, looks good!

jtpereyda (approved):

PR #516, opened 2021-05-20
Remove shutil_get_terminal_size dependency

SR4ven:
As we no longer support Python 2, there is no need for the shutil_get_terminal_size backport dependency.

This PR also contains a small bugfix for the TUI when aborting the session with CTRL+C.


jtpereyda (approved):
Looks good. Thanks @SR4ven !

PR #520, opened 2021-05-29
Drop Python 3.5 support

SR4ven:
In #511 we noticed that Python 3.5 is not compatible with some new features like trailing commas in argument lists or type hints.

As type hints are a great thing IMHO and less than 3% are using Python 3.5 according to this [survey](https://www.jetbrains.com/lp/python-developers-survey-2020/), I suggest to drop support for it.


jtpereyda (approved):
Sounds good!

Issue #502, opened 2021-04-04
How to specific the result directory to something other than boofuzz-results?


prRZ5F4LXZ:
```
boofuzz/constants.py
7:RESULTS_DIR = "boofuzz-results"
```

I see "boofuzz-results" is hard-code in the source code. Is there a way to specify where to save the file?

SR4ven:
Right now it seems to be fully hard coded.

I guess we could see this as a feature request? Shouldn't be too hard to implement an additional Session argument to specify the database file path.

jtpereyda:
Yep, I figure a `-o` option to specify the directory name. Right now "boofuzz-results" is a directory in which results directories are  created. I'm leaning toward following AFL's example and specifying the entire output directory names. E.g. `-o boofuzz-results/testRun1` etc.

SR4ven:
The session argument `db_filename` was implemented in #588.

A command line argument like `-o` still has to be added to boofuzz/boofuzz/cli.py.

PR #556, opened 2021-11-16
Sqldb_feature_add

Tresvian:
Adds db_filename to boofuzz.Session. This allows the sql database to be created in other paths other than /boofuzz-results/, and also lets the same database to be used continously.  However, ctrl-c and then restarting causes the first entry to duplicate a case name - this is mitigated by allowing graceful exits of boofuzz other than ctrl-c.

### DIFF
--- a/boofuzz/fuzz_logger_db.py
+++ b/boofuzz/fuzz_logger_db.py
@@ -35,16 +35,23 @@ def get_time_stamp():
 
 
 class FuzzLoggerDb(ifuzz_logger_backend.IFuzzLoggerBackend):
-    """Log fuzz data in a sqlite database file."""
+    """
+    Log fuzz data in a sqlite database file.
+    Using an existing database requires more graceful exits to prevent case number duplication.
+    """
 
     def __init__(self, db_filename, num_log_cases=0):
+        # Check if a db exists before the connect leaves an "empty" file behind.
+        db_already_exist = helpers.path_exists(db_filename)
         self._database_connection = sqlite3.connect(db_filename, check_same_thread=False)
         self._db_cursor = self._database_connection.cursor()
-        self._db_cursor.execute("""CREATE TABLE cases (name text, number integer, timestamp TEXT)""")
-        self._db_cursor.execute(
-            """CREATE TABLE steps (test_case_index integer, type text, description text, data blob,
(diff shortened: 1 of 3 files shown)


Tresvian:
Resubmitting pull request later. I need to use a corp account from what my company says to attribute work done :/

SR4ven:
Looking forward to the new PR!
It's also a feature requested in #502

Tresvian:
@SR4ven Is it okay to incorporate more changes within the same pull request? I'm changing more features to suit my company's tool better, and one of which is the the web app thread. With many sessions, the web app is not useful, but setting `Session.web_port = None` yields an exception:
```
>>> sess._index_end = 5
>>> sess.fuzz()
  File "/usr/lib/python3.8/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "/home/veerce/workbench/fuzzer/boofuzz/sessions.py", line 1272, in fuzz
    self._main_fuzz_loop(self._generate_mutations_indefinitely(max_depth=max_depth))
  File "/home/veerce/workbench/fuzzer/boofuzz/sessions.py", line 1374, in _main_fuzz_loop
    self.server_init()
  File "/home/veerce/workbench/fuzzer/boofuzz/sessions.py", line 1065, in server_init
    if not self.web_interface_thread.is_alive():
AttributeError: 'Session' object has no attribute 'web_interface_thread'
```

The context is that I'm manually managing a lot of Sessions to handle priority of protocols. Setting all Sessions to fuzz() synchronously causes them all to occupy greedy amounts of resources, but putting all requests into a single Session does not allow me to handle which protocol is more important. Also, if I wanted to more closely analyze which  protocol is causing effects, multi-threading the fuzzing would make it harder. The web app thread was just not going to be used in this case with multi-sessions.

SR4ven:
Generally changes to one component per PR are preferred. That makes it easier to review and discuss changes. However that shouldn't stop you from contributing, in case you can't split up the changes.

About disabling the web app, that works just fine for me. What boofuzz version are you using? You could try the latest master branch and see if the error persists.
I did notice that disabling the web app is not documented and will only work in `_main_fuzz_loop`. In the `_message_check` function the check `if self.web_port is not None` is missing which would result in the error you're seeing. The check should be in `server_init()` itself.

PR #558, opened 2021-11-20
Consistently disable web app and document feature

SR4ven:
It wasn't documented that it's possible to disable the web app.

Also, a check for whether or not to enable the web app was not always performed.

@Tresvian this might also help with your problem from https://github.com/jtpereyda/boofuzz/pull/556#issuecomment-973343846?


Tresvian:
I tested this on my end with multiple Sessions(web_port=None), and it works good! Thanks!

SR4ven:
Thanks for the feedback @Tresvian!

PR #560, opened 2021-11-23
Add type checking for args of Bytes primitive

### LINKED ISSUE #552: TypeError: can't concat int to bytes when mutating with layer 2 structure
I'm attempting to make a layer 2 fuzzer for a broadcast protocol, and am running into an issue with "can't concat int to bytes" on spinning up the prototype script.
```
asdf@ubuntu:~/fuzzer$ sudo python3 ./l2_prototype.py ens2
[2021-11-08 10:37:06,149]     Info: Web interface can be found at http://localhost:26000
[2021-11-08 10:37:06,150] Test Case: 1: AA:[AA.AA.DD.EE/PID:0]
[2021-11-08 10:37:06,150]     Info: Type: DWord
[2021-11-08 10:37:06,150]     Info: Opening target connection (ens2, type 0x0000)...
[2021-11-08 10:37:06,150]     Info: Connection opened.
[2021-11-08 10:37:06,150]   Test Step: Monitor CallbackMonitor#140737306343648[pre=[],post=[],restart=[],post_start_target=[]].pre_send()
[2021-11-08 10:37:06,150]   Test Step: Fuzzing Node 'AA'
[2021-11-08 10:37:06,158]     Error!!!! Unexpected exception! Traceback (most recent call last):
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/sessions.py", line 1388, in _main_fuzz_loop
                                  self._fuzz_current_case(mutation_context)
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/sessions.py", line 1754, in _fuzz_current_case
                                  self.transmit_fuzz(
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/sessions.py", line 1166, in transmit_fuzz
                                  data = self.fuzz_node.render(mutation_context)
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/blocks/request.py", line 130, in render
                                  return self.get_child_data(mutation_context=mutation_context)
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/fuzzable_block.py", line 71, in get_child_data
                                  rendered += item.render(mutation_context=mutation_context)
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/fuzzable.py", line 154, in render
                                  return self.encode(value=self.get_value(mutation_context=mutation_context), mutation_context=mutation_context)
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/blocks/block.py", line 114, in encode
                                  child_data = super(Block, self).get_child_data(mutation_context=mutation_context)
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/fuzzable_block.py", line 71, in get_child_data
                                  rendered += item.render(mutation_context=mutation_context)
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/fuzzable.py", line 154, in render
                                  return self.encode(value=self.get_value(mutation_context=mutation_context), mutation_context=mutation_context)
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/blocks/block.py", line 114, in encode
                                  child_data = super(Block, self).get_child_data(mutation_context=mutation_context)
                                File "/usr/local/lib/python3.8/dist-packages/boofuzz-0.4.0-py3.8.egg/boofuzz/fuzzable_block.py", line 71, in get_child_data
                                  rendered += item.render(mutation_context=mutation_context)
                              TypeError: can't concat int to bytes
```
Sorry I can't give the actual structure, but it's a combination of about 6 Byte, 6 DWord, 4 Block, and a single Repeat.
It's being run with this simple setup:
```
    interf=sys.argv[1]
    sess = boofuzz.Session(
        target=boofuzz.Target(
            connection=boofuzz.RawL2SocketConnection(interface=interf)
        )
    )
    req = boofuzz.Request("AA_fuzzer", children=(
        AA_struct
    ))
    sess.connect(req)
    sess.fuzz()
```

SR4ven:
It was possible to mistakenly pass a sting typed default_value or padding to the Bytes primitive resulting in an not obvious TypeError during runtime.
Now there is a type check in the constructor with an obvious message.

Fixes #552

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,10 @@ Features
 ^^^^^^^^
 - Added support for fuzzing NETCONF servers with the `NETCONFConnection` class.
 
+Fixes
+^^^^^
+- Added type checking for arguments of `Bytes` primitive to prevent incorrect use.
+
 v0.4.0
 ------
 Features
--- a/boofuzz/primitives/bytes.py
+++ b/boofuzz/primitives/bytes.py
@@ -1,5 +1,6 @@
 import functools
 import operator
+import typing
 
 from funcy import compose
 
@@ -119,11 +120,24 @@ class Bytes(Fuzzable):
         functools.partial(operator.mul, 100),
     ]
 
-    def __init__(self, name=None, default_value=b"", size=None, padding=b"\x00", max_len=None, *args, **kwargs):
-        if not isinstance(default_value, bytes):
-            raise TypeError("default_value must be bytes type")
-        if not isinstance(padding, bytes):
-            raise TypeError("padding must be bytes type")
+    def __init__(
+        self,
+        name: str = None,
+        default_value: typing.Union[bytes, memoryview] = b"",
+        size: int = None,
+        padding: typing.Union[bytes, memoryview] = b"\x00",
+        max_len: int = None,
+        *args,
+        **kwargs
+    ):
+        try:
+            memoryview(default_value)
+        except TypeError:
+            raise TypeError("default_value of Bytes must be bytes-like type")
+        try:
+            memoryview(padding)
+        except TypeError:
+            raise TypeError("padding of Bytes must be bytes-like type")
 
         super(Bytes, self).__init__(name=name, default_value=default_value, *args, **kwargs)
 
--- a/boofuzz/primitives/bytes.py
+++ b/boofuzz/primitives/bytes.py
@@ -1,6 +1,6 @@
 import functools
 import operator
-import typing
+from collections.abc import ByteString
 
 from funcy import compose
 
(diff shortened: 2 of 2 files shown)


jtpereyda:
I think a bytes-like object should do; it doesn't necessarily need to be `bytes`. `bytearray` is probably the most common other example. However, [How to test for bytes-like object?](https://stackoverflow.com/q/44874609/461834) is an unanswered question.

[This answer has a convert-to-memoryview test](https://stackoverflow.com/a/30018151/461834) that will work for at least bytes, bytearray, and memoryview; perhaps we should use this check?

SR4ven:
Good point @jtpereyda, I didn't think about bytes-like objects.

I ran some tests and the memoryview conversion seems to be the only one working for bytes, bytearray and memoryview.

According to the [docs](https://docs.python.org/3.7/library/typing.html#typing.ByteString) checking with isinstance for typing.ByteString should work too, but doesn't. It won't accept memoryview for some reason.

SR4ven:
While running some tests I noticed that we run into exceptions when using a memoryview typed default_value or padding.
This is because memoryview doesn't support `*` operand with int. But we need that for example when extending the padding to the defined size.

I've removed memoryview type from the allowed types and added a check for `collections.abc.ByteString` instead. That covers bytes and bytearray.

jtpereyda (approved):
Looks great. Thanks @SR4ven !

PR #562, opened 2021-11-25
Fix init of FuzzableBlock children in Request

### LINKED ISSUE #550: Docs: FuzzableBlock inheritance vs initialize children for Block
Here guide says that for making your own primitive you should inherit from Fuzzable or FuzzableBlock:
https://github.com/jtpereyda/boofuzz/blob/373a5cb071f68b5076b098472db31de841c87389/docs/user/protocol-definition.rst#making-your-own-blockprimitive

Howewer, Request initialize children only for a Block primitive (which is inherited from FuzzableBlock).
https://github.com/jtpereyda/boofuzz/blob/373a5cb071f68b5076b098472db31de841c87389/boofuzz/blocks/request.py#L58

So basically if you just inherit from FuzzableBlock like instruction says, you will encounter uninitialized children for your custom block and there will be an exception when you try to fuzz. To get over this exception as a workaround you can just inherit from a Block primitive.

So how to fix it:
1. Update the check to initialize children for a FuzzableBlock.
2. Update instructions with advice to inherit from a Block.
3. Don't do anything. Then some people will probably encounter this problem when they will try to make a custom block according to instructions and they will have to look into inner implementation of Request building to find a similar workaround.

SR4ven:
Currently, blocks inheriting from `FuzzableBlock` are not correctly initialized in `Request`.

As `Block` and `Aligned` both inherit from `FuzzableBlock` the isinstance check will still return true with this change.

Fixes #550
@678098 can you confirm that this works?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ Fixes
 ^^^^^
 - Fixed check for when to enable the web app.
 - Documented the possibility to disable the web app.
+- Correctly initialize all children of a request which inherit from `FuzzableBlock`.
 
 v0.4.0
 ------
--- a/boofuzz/blocks/request.py
+++ b/boofuzz/blocks/request.py
@@ -115,7 +115,7 @@ class Request(FuzzableBlock, Node):
             self.block_stack[-1].push(item)
 
         # add the opened block to the block stack.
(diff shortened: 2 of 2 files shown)


jtpereyda:
Thanks for the fix @SR4ven !

PR #588, opened 2021-12-07
Specify location of sqlite database and use the same file for logging

ruiz-battelle:
Added a kwarg in `boofuzz.Session(db_filename=None)` for allowing control of the sqlite database location, reusing the same db, and disabling the default directory boofuzz-results if a specific location is chosen. Some modifications in helper functions lowered the code footprint.l

Ungraceful exiting of boofuzz fuzzing results in the final entry being duplicated.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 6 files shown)


ruiz-battelle:
This solves #502 and is a spiritual successor to #556

SR4ven:
Thanks for your contribution @ruiz-battelle!

I merged the current master branch into this PR and added a line about the new feature to the changelog.

Merging now.

PR #594, opened 2022-01-25
Update to make boofuzz more compatible

hegusung:
I had these 3 issues when using boofuzz:
- I could not change processmonitor.py listening interface
- processmonitor.py was not compatible with Windows, multiple function used in boofuzz were not available for Windows.
- Some processes to fuzz needs a lot of time before starting up, so the hardcoded 3 seconds delay were not sufficient. A parameter "restart_wait" in Session class has been added to solve this issue

### DIFF
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -443,6 +443,7 @@ class Session(pgraph.Graph):
         restart_callbacks=None,
         restart_threshold=None,
         restart_timeout=None,
+        restart_wait=3,
         pre_send_callbacks=None,
         post_test_case_callbacks=None,
         post_start_target_callbacks=None,
@@ -480,6 +481,7 @@ class Session(pgraph.Graph):
         self.restart_sleep_time = restart_sleep_time
         self.restart_threshold = restart_threshold
         self.restart_timeout = restart_timeout
+        self.restart_wait = restart_wait
         if fuzz_loggers is None:
             fuzz_loggers = []
             if self.console_gui and os.name != "nt":
@@ -1043,8 +1045,8 @@ class Session(pgraph.Graph):
                 self._fuzz_data_logger.log_info("Restarting target process using {}".format(monitor.__class__.__name__))
                 if monitor.restart_target(target=target, fuzz_data_logger=self._fuzz_data_logger, session=self):
                     # TODO: doesn't this belong in the process monitor?
-                    self._fuzz_data_logger.log_info("Giving the process 3 seconds to settle in")
-                    time.sleep(3)
+                    self._fuzz_data_logger.log_info("Giving the process %d seconds to settle in" % self.restart_wait)
+                    time.sleep(self.restart_wait)
                     restarted = True
                     break
 
--- a/process_monitor.py
+++ b/process_monitor.py
@@ -8,28 +8,20 @@ from boofuzz.utils.debugger_thread_simple import DebuggerThreadSimple
 from boofuzz.utils.process_monitor_pedrpc_server import ProcessMonitorPedrpcServer
 
 
-def serve_procmon(port, crash_bin, proc_name, ignore_pid, log_level):
+def serve_procmon(ip, port, crash_bin, proc_name, ignore_pid, log_level):
     with ProcessMonitorPedrpcServer(
-        host="0.0.0.0",
-        port=port,
-        crash_filename=crash_bin,
-        debugger_class=DebuggerThreadSimple,
-        proc_name=proc_name,
-        pid_to_ignore=ignore_pid,
-        level=log_level,
-        coredump_dir=None,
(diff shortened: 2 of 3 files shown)


jtpereyda:
@hegusung Sounds great! I'll put this on my queue to review. Which operating systems have you tested on so far with these changes?

hegusung:
Hi, So boofuzz script was tested on both a Kali linux and a Debian. Process_monitor.py was tested on a Windows 10. 

there shouldn't be an issue with process_monitor.py on linux because the linux way is always done, il only switches to windows way when an exeption is detected.

hegusung:
I don't understand what is wrong with stickler-ci

SR4ven:
> I don't understand what is wrong with stickler-ci

Seems like it runs into a timeout. Nothing u need to worry about.
If you want to check for code style issues, you can run `tox` or `tox -e lint` (not exactly sure about the second one). Also note that it outputs all findings, not just the ones in your changes.

process_monitor.py
```diff
@@ -58,8 +50,9 @@ def serve_procmon(port, crash_bin, proc_name, ignore_pid, log_level):
 )
 @click.option("--proc-name", "--proc_name", "-p", help="process name to search for and attach to", metavar="NAME")
 @click.option("--port", "-P", help="TCP port to bind this agent to", type=int, default=DEFAULT_PROCMON_PORT)
-def go(crash_bin, ignore_pid, log_level, proc_name, port):
-    serve_procmon(port=port, crash_bin=crash_bin, proc_name=proc_name, ignore_pid=ignore_pid, log_level=log_level)
+@click.option("--ip", "-I", help="IP of the computer", type=str, default="0.0.0.0")
```
SR4ven, line 53: We should add the same option to process_monitor_unix.py for consistency.

Can we make the help string of the ip option more obvious? Maybe something like "Listening IP for incoming connections"

boofuzz/sessions.py
```diff
@@ -1043,8 +1045,8 @@ def _restart_target(self, target):
                 self._fuzz_data_logger.log_info("Restarting target process using {}".format(monitor.__class__.__name__))
                 if monitor.restart_target(target=target, fuzz_data_logger=self._fuzz_data_logger, session=self):
                     # TODO: doesn't this belong in the process monitor?
-                    self._fuzz_data_logger.log_info("Giving the process 3 seconds to settle in")
-                    time.sleep(3)
+                    self._fuzz_data_logger.log_info("Giving the process %d seconds to settle in" % self.restart_wait)
+                    time.sleep(self.restart_wait)
```
SR4ven, line 1049: I'm not 100% happy with adding yet another Session argument. How about reusing `restart_sleep_time`? It is currently used to simply sleep for the given time if no other reset handler is available.

Also, as the TODO suggests, the logic to get the target task up and running belongs in the process monitor not in Session. Nothing that has to be done here, but another argument against adding the parameter to Session.

SR4ven (changes requested):
Thanks for the PR @hegusung!

I left some comments below.

boofuzz/sessions.py
jtpereyda: I like the idea of adding the arg to the procmon best, but I'm OK with an arg in Session since it's better than nothing. I agree with @SR4ven's idea of reusing `restart_sleep_time`.

process_monitor.py
jtpereyda: Another help text suggestion: "Listen on this IP for incoming connections from boofuzz"
Let's do a 127.0.0.1 default for security.

The process_monitor_unix.py changes can be done in this PR, or separately.

jtpereyda (changes requested):
@hegusung Thanks for this work! This was a significant gap in boofuzz.

After this PR, it wouldn't be much work to integrate the two procmon scripts, since the Simple debugger will work with both Windows and Unix.

Thanks @SR4ven for the review!

hegusung:
Hi, I see that this PR is not accepted. 
Which changes should I do in order for you guys to accept it ?

boofuzz/fuzzable.py
```diff
@@ -50,6 +50,8 @@ def __init__(self, name=None, default_value=None, fuzzable=True, fuzz_values=Non
             Fuzzable.name_counter += 1
             self._name = "{0}{1}".format(type(self).__name__, Fuzzable.name_counter)
 
+        self.index_mutation = 0
```
SR4ven, line 53: What do you think about `mutation_index`?
```suggestion
        self.mutation_index = 0
```

SR4ven (changes requested):
Apologies for the delayed review.

Thanks for implementing our comments!
I've added one more tiny suggestion.

Could you add some lines about those fixes to the CHANGELOG.rst please?

PR #597, opened 2022-02-01
Remove legacy future/six code, imports and dependencies

SR4ven:
Remove unused code and dependencies for Python 2 backwards compatibility.

Code will be easier to read and possibly more performant. Plus we will have less packet dependencies.

TODO:
- [x] remove future/past
- [x] remove six (done in #603)
- [x] changelog entry

PR #603, opened 2022-02-26
Remove six library

aedrax:
I brought the six library into this repo 3 years ago and now I'm ripping it out :P viva la python3!

### DIFF
--- a/boofuzz/__init__.py
+++ b/boofuzz/__init__.py
@@ -3,7 +3,6 @@ from __future__ import absolute_import
 import functools
 import sys
 
-import six
 from past.builtins import map
 
 from . import blocks, exception, legos, primitives
@@ -591,7 +590,7 @@ def s_binary(value, name=None):
         pair = parsed[:2]
         parsed = parsed[2:]
 
-        value += six.int2byte(int(pair, 16))
+        value += int(pair, 16).to_bytes(1, "little")
 
     blocks.CURRENT.push(Static(name=name, default_value=parsed))
 
--- a/network_monitor.py
+++ b/network_monitor.py
@@ -59,7 +59,7 @@ Network Device List:
     for index, pcapy_device in enumerate(ifs):
         # if we are on windows, try and resolve the device UUID into an IP address.
         if sys.platform.startswith("win"):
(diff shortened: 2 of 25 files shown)


aedrax:
@jtpereyda I think your stickler-ci instance is dead?

@SR4ven I think this will resolve your remaining todo for #597

network_monitor.py
```diff
@@ -59,7 +59,7 @@ def create_usage(ifs):
     for index, pcapy_device in enumerate(ifs):
         # if we are on windows, try and resolve the device UUID into an IP address.
         if sys.platform.startswith("win"):
-            from six.moves import winreg  # pytype: disable=import-error
+            import winreg  # pytype: disable=import-error
```
jtpereyda, line 62: pretty sure we can remove this warning ignore comment

jtpereyda (commented):
one suggestion

network_monitor.py
SR4ven: I think we still need that ignore comment. It prevents warnings being thrown on none Windows platforms where winreg is not available.

aedrax: Right, I'm fairly certain the comment must remain

boofuzz/__init__.py
```diff
@@ -591,7 +590,7 @@ def s_binary(value, name=None):
         pair = parsed[:2]
         parsed = parsed[2:]
 
-        value += six.int2byte(int(pair, 16))
+        value += int(pair, 16).to_bytes(1, "little")
```
SR4ven, line 593: Since six defines
```python
int2byte = struct.Struct(">B").pack
```
it would be more accurate to replace it with `to_bytes(1, "big")`, wouldn't it?
https://docs.python.org/3/library/struct.html#byte-order-size-and-alignment

SR4ven: The size or length is 1 so it doesn't really matter what endianness we set. Still, it would be more accurate.

SR4ven: Alright, I agree

SR4ven (approved):
Nice work @aedrax! Thanks a lot

PR #604, opened 2022-02-26
Use file encoding write

aedrax:
Not specifying encoding when writing a file can cause UnicodeEncodeError because Python assumes the string's characters can fit in the OS's default text encoding, but that's often an invalid assumption. [Read more](https://codereview.doctor/features/python/best-practice/file-open-write-encoding)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+Fixes
+^^^^^
+- Specified encoding on file write rather than assuming default encoding
+
 v0.4.1
 ------
 Features
--- a/boofuzz/utils/debugger_thread_simple.py
+++ b/boofuzz/utils/debugger_thread_simple.py
@@ -230,7 +230,7 @@ class DebuggerThreadSimple(threading.Thread):
         if self.is_alive():
             return True
         else:
-            with open(self.process_monitor.crash_filename, "a") as rec_file:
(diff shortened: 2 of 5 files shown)


SR4ven:
Thanks for resolving the merge conflict @aedrax.

After looking at the usage of file open() in our code I think some changes might be missing.
https://github.com/jtpereyda/boofuzz/blob/0636474398c5fa43af60d3838feb8efa6e79ef40/boofuzz/connections/file_connection.py#L47
https://github.com/jtpereyda/boofuzz/blob/0636474398c5fa43af60d3838feb8efa6e79ef40/boofuzz/sessions.py#L739
https://github.com/jtpereyda/boofuzz/blob/0636474398c5fa43af60d3838feb8efa6e79ef40/boofuzz/sessions.py#L715

The `FromFile` primitive is missing an encoding constructor argument, but that can be done in another PR.
https://github.com/jtpereyda/boofuzz/blob/0636474398c5fa43af60d3838feb8efa6e79ef40/boofuzz/primitives/from_file.py#L35

aedrax:
Thanks! I believe these are actually fine because the error only exists when writing out, so the reads are okay. And the one with "wb" is also fine because it's in binary write mode. Let me know if you'd still prefer me to do it to all of them!

SR4ven:
Right, that makes sense. Thanks for the explanation.

jtpereyda:
Thank you @aedrax and @SR4ven !

PR #605, opened 2022-02-27
Remove redundant inheritance of Object class

aedrax:
Since we're now exclusively python3, a class is inherited from object by default, so explicitly inheriting from object is redundant. Removing it keeps the code simpler. [Read more](https://codereview.doctor/features/python/best-practice/avoid-inheriting-from-object)


SR4ven:
Thanks @aedrax!

Issue #607, opened 2022-03-06
Mirror primitive always returns the default value of the target primitive


hycrypt:
### Report

Hi,
it seems like the mirror primitive does not mirror the current mutation of the target primitive.
Instead the mirror primitive always returns the default value of the target primitive.
I have tried it with a clean virtual environment and copied the example from the test_primitives.py in the unit_tests folder.
When using the example with a Session object the problem occurs. 
Just executing the test_s_mirror function in test_primitives.py does not throw an error tough.

### Expected behavior

The Mirror primitive should mirror the current mutation of the target primitive.

### Actual behavior

The Mirror primitive always returns the default value of the target primitive.
In the example the closing tag is always `</x>` but actually it should match the opening tag.

![Screenshot from 2022-03-06 23-08-57](https://user-images.githubusercontent.com/79507881/156944141-93c1397f-5620-4125-abc2-828e8dc88a63.png)

### Steps to reproduce the problem

1. setup boofuzz in a clean virtual environment
2. run the provided boofuzz script

### boofuzz script

```python
from boofuzz import *

def main():
    session = Session(target=Target(connection=FileConnection("test")))

    test_group_values = [b"a", b"bb", b"ccc", b"dddd"]
    s_initialize("test_s_mirror")

    s_size("data", output_format="ascii", fuzzable=False, name="size")
    s_mirror(".size", name="size_mirror")

    with s_block("data"):
        s_static("<")
        s_group("group_start", default_value=b"x", values=test_group_values)
        s_static(">")
        s_static("hello")
        s_static("</")
        s_mirror("data.group_start", name="group_end")
        s_static(">")

    req = s_get("test_s_mirror")

    session.connect(req)
    session.fuzz()


if __name__ == "__main__":
    main()
```

### boofuzz version

0.4.1

### Python version

3.10

### Platform

Linux

cq674350529:
@hycrypt It seems that the `s_mirror()` is broken during previous refactor. https://github.com/jtpereyda/boofuzz/pull/608 should solve this.

hycrypt:
Yes, #608 fixes the problem. Thank you very much for the super fast response and fix. :)

PR #608, opened 2022-03-07
fix s_mirror primitive

cq674350529:
pass the `mutation_context` to `_render_primitive()`, try to solve https://github.com/jtpereyda/boofuzz/issues/607


jtpereyda (approved):
LGTM. Thanks for the fix @cq674350529 !

PR #614, opened 2022-03-31
Update sessions.py

siw3kosky:
- Changed plain print about completion of fuzzing session into log info message.
- Catching EOFError exception when all test cases are executed to not throw traceback.

### DIFF
--- a/boofuzz/sessions.py
+++ b/boofuzz/sessions.py
@@ -1413,10 +1413,10 @@ class Session(pgraph.Graph):
 
             if self._keep_web_open and self.web_port is not None:
                 self.end_time = time.time()
-                print(
-                    "\nFuzzing session completed. Keeping webinterface up on {}:{}".format(
+                self._fuzz_data_logger.log_info(
+                    "Fuzzing session completed. Keeping webinterface up on {}:{}".format(
                         self.web_address, self.web_port
-                    ),
+                    ) +
                     "\nPress ENTER to close webinterface",
                 )
                 input()
@@ -1431,6 +1431,9 @@ class Session(pgraph.Graph):
             raise
         except exception.BoofuzzTargetConnectionFailedError:
             # exception should have already been handled but rethrown in order to escape test run
+            pass
+        except EOFError:
+            # Fuzzing session completed
             pass
         except Exception:
             self._fuzz_data_logger.log_error("Unexpected exception! {0}".format(traceback.format_exc()))


jtpereyda:
Thanks @siw3kosky ! In what situation does the `EOFError` happen? Sometimes the test cases can end and not throw an `EOFError`, right?

siw3kosky:
Unfortunately, in my case, I always got `Fuzzing session completed....` message followed by traceback in `sessions.py` in `_main_fuzz_loop` function in `input()` call with message `EOFError: EOF when reading a line`.

SR4ven:
@siw3kosky how do you execute the boofuzz script? Is it possible that the terminal doesn't support `stdin`?
EOFError indicates that an EOF was received before any other input.
https://stackoverflow.com/questions/12547683/python-3-eof-when-reading-a-line-sublime-text-2-is-angry

Maybe we should put a try catch directly around `input()` for use cases where no `stdin` is available.

siw3kosky:
I executed it directly from the ubuntu shell and via vscode.
In both situations traceback appears at the end.
I thing that it is not possible that the terminal doesn't support `stdin`.

> Maybe we should put a try catch directly around input() for use cases where no `stdin` is available.

I think thats also a good solution, should I create new pull request? :)

jtpereyda:
@siw3kosky We should trace down that error and figure out why it's happening -- ideally we shouldn't cause the exception in the first place if we can help it. Could you share a very simple script that reproduces the problem?

siw3kosky:
Ok gentlemen, after a short investigation I know why it is happening.

> Is it possible that the terminal doesn't support `stdin`?

I run my script with multiprocessing, so it is running in the background.
When a process runs in the background, I can't send input to the process via terminal and hence `input()` can't work.

jtpereyda:
@siw3kosky That makes sense! Which python library are you using for multithreading?

In the case of multithreading, does it still make sense to leave the web server open for a time?

siw3kosky:
I am using `multiprocessing`.
I think that it still make sense to leave the web server open for a time.

jtpereyda:
@siw3kosky are you able to share an example script showing how you run the fuzzer with `multiprocessing`? This sounds like an interesting use case that would be cool to support.

If nothing else, I would wrap a try/except directly around the `input()` line. But I'm sure there's a more appropriate way to handle the use case.

SR4ven:
We could also detect if the shell is interactive or not. https://stackoverflow.com/questions/2356399/tell-if-python-is-in-interactive-mode
But I guess simply trying and possibly catching the exception would be a more pythonic way.
How would we keep the webinterface open in a multithreaded context? Sleep indefinitely?

jtpereyda:
I think it depends on the threading scheme being used. I was hoping @siw3kosky could share the approach being used.

PR #618, opened 2022-04-05
Disable pytype on stickler-ci

SR4ven:
Stickler CI is failing a lot recently because of timeouts. Pytype seems to run too long so after 5 minutes the run gets aborted.


jtpereyda:
Bummer about pytype. Thanks for tracking down and fixing this @SR4ven !

jtpereyda (approved):

PR #627, opened 2022-06-07
fix s_update primitive

cq674350529:
Upate the `s_update()`:

+  use `_default_value` instead, for there is no `_value` field anymore
+ use `qualified_name` instead of `name` field to be consistent with `blocks.CURRENT.names`

P.S. don't know why there are irrelevant commits :(


SR4ven:
Thanks for the fix @cq674350529!
Could you add a line about this to the changelog please?

PR #636, opened 2022-08-04
Fixed minimum python version in INSTALL.rst

mistressofjellyfish:
Changelog says that since v0.4.0, the minimum version is python 3.6, while INSTALL.rst (and thus the docs) still say python 3.5.

I'll assume that 3.6 is correct, this PR fixes the docs.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -4,7 +4,7 @@ Installing boofuzz
 Prerequisites
 -------------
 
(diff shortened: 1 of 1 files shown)


SR4ven:
Good point. I forgot that we have the Python version requirement in docs.
Thanks for the fix @mistressofjellyfish

PR #639, opened 2022-08-09
Proposal: Add anchors for protocol-definition.rst

SR4ven:
I thought it would be nice to have the blocks and primitives listed in the navigation tree on the left. For that I've added headings for each autofunction.

The result will look like this:
grafik

Any thoughts?


jtpereyda:
Sweet! Looks great!

PR #649, opened 2022-11-13
Fix invalid data type in socket tests

SR4ven:
This error has been around a long while.
With recent changes to pytest it now makes the tests fail.


SR4ven:
There was a second problem that was introduced by a pytest bdd update.
Apparently they changed how "Given" strings are parsed and one string was misinterpreted.

Issue #601, opened 2022-02-21
Logs to a txt or csv file missed some line for the last Test case


vincentwh2018:
### Report

Hi Team,

I have a boofuzz script as below, log 3 type files: txt, csv and db.
def main():
    fuzz_loggers = []
    txt_log = open("/srv/boofuzz/http-fuzz-results.txt", "w")
    fuzz_loggers.append(FuzzLoggerText(file_handle=txt_log))
    csv_log = open("/srv/boofuzz/http-fuzz-results.csv", "w")
    fuzz_loggers.append(FuzzLoggerCsv(file_handle=csv_log))

    session = Session(
        target=Target(connection=TCPSocketConnection("192.168.1.1", 80)),
        fuzz_loggers=fuzz_loggers,
        keep_web_open=False,
        restart_threshold=5,
        db_filename="/srv/boofuzz/http-fuzz-results.db"
    )

    define_proto_static(session=session)
    session.fuzz()

After scan finished, when I check the db file by "boo open...", I found that there are some lines missed in txt and csv files compared with the GUI.

Here is the result from GUI for the last case:
![boofuzz_gui](https://user-images.githubusercontent.com/39822591/154919899-fd7f90b1-aca6-4625-bdec-5123a3e52bb9.png)

Here is the txt log file for the last case:
![boofuzz_txt](https://user-images.githubusercontent.com/39822591/154919959-996a20d4-947d-4185-8b84-60bfc21e3213.png)

### Expected behavior

The "Check OK: No crash detected." and other lines are contained in txt&csv files for the last case.

### Actual behavior

The "Check OK: No crash detected." and other lines are missed in txt&csv files for the last case.

### Steps to reproduce the problem

1. Just run the boofuzz script as above
2. Use "boo open..." to check the result
3. Open the txt log file, missed some lines

### boofuzz script

```python
def main():
    fuzz_loggers = []
    txt_log = open("/srv/boofuzz/http-fuzz-results.txt", "w")
    fuzz_loggers.append(FuzzLoggerText(file_handle=txt_log))
    csv_log = open("/srv/boofuzz/http-fuzz-results.csv", "w")
    fuzz_loggers.append(FuzzLoggerCsv(file_handle=csv_log))

    session = Session(
        target=Target(connection=TCPSocketConnection("192.168.1.1", 80)),
        fuzz_loggers=fuzz_loggers,
        keep_web_open=False,
        restart_threshold=5,
[...]

PR #655, opened 2023-01-02
Flush log data when closing the test.

ptdropper:
https://github.com/jtpereyda/boofuzz/issues/601
Text log file and CSV log file requires a file flush so the data is consistent with the boofuzz db file. Amended file formatting using the results of the "black" program.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Fixes
 - Changed type of `default_value` from string to bytes for `FromFile`.
 - `s_update` primitive was out of date.
 - The minimum supported Python version is now 3.7.
+- Flush log data when closing the test.
 
 v0.4.1
 ------
--- a/boofuzz/fuzz_logger_csv.py
+++ b/boofuzz/fuzz_logger_csv.py
@@ -1,6 +1,6 @@
+import sys
 import csv
 import datetime
-import sys
 
 from . import helpers, ifuzz_logger_backend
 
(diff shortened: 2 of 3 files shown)


SR4ven:
Hi @ptdropper, thanks for looking into this.

I just ran some tests myself, and my guess is that the file handles were not closed correctly in #601.

With the following code I'm able to recreate the problem described:
```python
text_log = open("log.txt", "w")
csv_log = open("log.csv", "w")

sess = Session(
    target=Target(connection=TCPSocketConnection("127.0.0.1", 80)),
    fuzz_loggers=[FuzzLoggerText(text_log), FuzzLoggerCsv(csv_log)],
)

sess.connect(s_get("Request"))
sess.fuzz()

text_log.close()
csv_log.close()
```
In this example, when sess.fuzz() doesn't return cleanly (e.g. KeyboardInterrupt exception), the file handles won't be closed gracefully and buffered data is lost.

Using the following code, everything works as expected and no data is lost. (A `with` statement would be even better.)
```python
text_log = open("log.txt", "w")
csv_log = open("log.csv", "w")

sess = Session(
    target=Target(connection=TCPSocketConnection("127.0.0.1", 80)),
    fuzz_loggers=[FuzzLoggerText(text_log), FuzzLoggerCsv(csv_log)],
)

sess.connect(s_get("Request"))

try:
    sess.fuzz()
finally:
    text_log.close()
    csv_log.close()
```

Anyway, this is a general issue related to file operations. Since both logger classes take a file handle and not a file path, it is not the responsibility of the boofuzz logger to flush or close the handle. This is up to the scope in which the file was opened.

So unless I've missed something, I see no need to change the boofuzz loggers.

Issue #521, opened 2021-05-31
Improve documentation - "Boofuzz by examples"


lucca-ruhland:
Hello everyone,
 I think that the boofuzz documentation is good, but somehow not very useful for beginners.
In my opinion, the quickstart section is too short, leaving new users without any guidance on how to use this framework.
 So I would suggest adding a whole section with detailed examples, covering at least the current (examples)[https://github.com/jtpereyda/boofuzz/tree/master/examples], but maybe with a more detailed description, useful hints, best practices, etc.
 It could be called "Boofuzz by examples" or something like that.

SR4ven:
That would very helpful indeed. I remember having the same feeling when I first used boofuzz.

At the moment I don't have the time to write a detailed documentation, but any contribution is welcome.

MetinSAYGIN:
Hello eveyone,

I started to work on this issue, however if you have any new suggestion/idea let me know, the ticket is pretty old so ..

Metin

PR #660, opened 2023-02-22
Documentation improve

MetinSAYGIN:
For ticket #521

Here is a small example, i don't know if i have to go deeper on it because it's for quickstart section.
Should be worth to pull it for beginners.

Metin

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -115,6 +115,10 @@ To perform this test, you will need an HTTP server, you can use python for that.
 
     $ python3 -m http.server
     
+.. code-block:: bash
+
+    $ boo open <run-*.db>
+    
 Then run the http_simple.py script on the ip/port that the server use.
 
 
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -97,7 +97,7 @@ If you are doing crazy cool stuff, check out the :ref:`community info <community
 
 Happy fuzzing, and Godspeed!
 
-Example
+Another examples
 ==========
 Ftp Simple
 Here is an example of execution of this script : `ftp_simple <https://github.com/jtpereyda/boofuzz/blob/master/examples/ftp_simple.py>`_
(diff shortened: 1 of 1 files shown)


jtpereyda (changes requested):
@MetinSAYGIN Thanks for the PR! This is a simple and helpful improvement. I think there will be some editorial changes, but for starters, have you rendered the docs to check the format? See `tox.ini` which has the sphinx build commands. You can build locally and check out the rendered HTML.

Edit: For example, I think this bit will look funny once built:
```
To perform this test, you will need an HTTP server, you can use python for that.
python3 -m http.server
Then run the http_simple.py script on the ip/port that the server use.
```

MetinSAYGIN:
Hi @jtpereyda, it should be better now :)
Metin

jtpereyda:
Thank you for the docs contribution @MetinSAYGIN !

jtpereyda (approved):

SR4ven:
No one approved the test runs on this PR and now the docs fail to build:
https://github.com/jtpereyda/boofuzz/actions/runs/4515478396/jobs/7952811037

jtpereyda:
@SR4ven Oh, hmm, that's my bad isn't it? I saw all green checkmarks and just hit merge 😬

SR4ven:
No worries @jtpereyda, I'll fix it later.
The reason for this is that first time PR submitters have to get GitHub Actions approved manually.
SticklerCI apparently doesn't have this requirement so that's why u saw the green checkmark.

Also remember to squash and merge for clean history :D If I remember correctly, you can even set that as the default in the project settings.

jtpereyda:
@SR4ven they have to approve Actions in their own settings?

[From what I could find](https://stackoverflow.com/a/49416216/461834):

> The GitHub user interface will default the option to the last option selected by the user for the current project.

So I'll just have to be more careful next time.

SR4ven:
@jtpereyda no we have to approve test runs on PRs from users that haven't contributed to the project yet.

Yeah right, that's how the default option works.
I confused this with a general project setting in the Pull Request section. There you can disallow certain merge strategies.

PR #680, opened 2023-08-21
Parallel targets support

ttufts:
Finished implementing parallel target support, removed hardcoded target index 0 and added iterator cycling so that mutations are load leveled over all targets that are added.


SR4ven (changes requested):
Sorry for the late review and thanks for the PR @ttufts!

I just tested the change with two targets and it seems that the test are now being sent to the targets in an alternating way.
Case 1 -> Target 1
Case 2 -> Target 2
Case 3 -> Target 1
Case 4 -> Target 2

When it should be like this:
Case 1 -> Target 1
Case 1 -> Target 2
Case 2 -> Target 1
Case 2 -> Target 2

We should also add some kind of info about the current target to the logging, so that it's clear which target is currently in use.

Additionally, we should give the constructor of `Session` an option for multiple targets in form of a list.
Also, I noticed, that calling `Session.add_target()` will not execute the following code that is currently only present in the constructor.
https://github.com/jtpereyda/boofuzz/blob/12ec28c67170c62410845ef67774b39845a84a80/boofuzz/sessions.py#L578-L591
So monitors are not equally set up with both methods.

PR #687, opened 2023-09-24
Switch to poetry build system

SR4ven:
Here is an attempt to switch to Poetry and pyproject.toml to build the package.

The setup.py installation method has been deprecated for a while now and Poetry seemed versatile, well documented and easy to use.


jtpereyda:
@SR4ven One use case I can't quite figure out with Poetry is how to install and make entry points generally available. I sued to recommend running:
```
pipx install boofuzz
# or, from the boofuzz directory:
pipx install .
```

This would use editable mode, but even more helpfully, it would make the `boo` script available as standalone executable, which when run would run the `boo` script using the correct virtualenv.

Do you know how to do something similar with Poetry?

SR4ven:
After installing from source in editable mode with `poetry install`, you can use `poetry shell` to activate the virtualenv.
Then `boo` is available.

I guess that one could also activate the virtualenv bypassing poetry as it's a normal virtualenv located in `~/.cache/pypoetry/virtualenvs/`

Does that help?

jtpereyda:
@SR4ven That's what I'm doing for now, `poetry shell` -- I just miss how pipx would throw it right on your PATH and let you execute it as a regular CLI tool.

Since we're publishing to PyPI, I _think_ one could still `pipx install boofuzz` to get the latest release as a regular CLI tool. It's just harder in dev. 🤷‍♂️ It can still be done manually, but the magical thing `pipx` does is make a little python script that uses the specific virtualenv python interpreter, meaning one can run the script without starting a virtualenv first.

SR4ven:
Alright, didn't know about that `pipx` feature.
But you're right, Poetry is only for developing and publishing. If you just intend to install boofuzz for using it with regular pip or pipx, no Poetry is involved. Poetry can only install from source, something like `poetry install boofuzz` does not work.

Since we don't build and release locally, I guess that you could still use `pipx` for development. `pip install -e .[dev]` still works fine, so I don't see why it shouldn't work with `pipx`.
We could add a `pipx` install method to the docs too.

I'm using PyCharm for developing which has a Poetry integration that I find quite convenient. It automatically activates the virtualenv in the shells.

PR #706, opened 2024-02-14
[FEATURE] add prefix for app

skapin:
Hi, I patched the application to handle a PREFIX when the app is deployed on wbeserver like :
 

`myserver.com/subpath/foobuzz/`

I had to: 
- use `url_for`  everywhere 
- prefix all route
- add prefix on app start for static path

Prefix is loaded from FLASK_APP_PREFIX env. This is the quickest way. The clean way is to use blueprint and propre web server starting

### DIFF
--- a/boofuzz/web/app.py
+++ b/boofuzz/web/app.py
@@ -1,4 +1,5 @@
 import re
+import os
 
 import flask
 from flask import Flask, redirect, render_template
@@ -7,8 +8,10 @@ from .. import exception
 
 MAX_LOG_LINE_LEN = 1500
 
-app = Flask(__name__)
-app.session = None
+prefix = os.environ.get('FLASK_APP_PREFIX', '')
+
+app = Flask(__name__, static_url_path=f'{prefix}')
+app.session = None  # Initialize your session as needed
 
 
 def commify(number):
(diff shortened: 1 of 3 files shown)


skapin:
I have passed `black`, you can retrigger

jtpereyda:
Very cool @skapin ! Have you functional tested both with and without the prefix?

jtpereyda:
Docs build is failing -- not sure if that is new or existing.

skapin:
i didnt touch any doc.
why should I need ?


For tests, it would be good if someone test it too, especially since i'm not the user using boofuzz, I made the change for someone. Because of this, I tested one usecase only.

SR4ven:
Regarding the failing docs build, it's just a link-check that returns 403 Forbidden. Would ignore that for this PR and I'll look into this separately.
Maybe Stackoverflow has blocked robot requests coming from GitHub CI server or so.

jtpereyda:
@skapin Are you saying you tested it `FLASK_APP_PREFIX` but did not test it without?

jtpereyda:
Ignoring the docs issue as it's separate. Merging. Thanks @skapin @SR4ven !

PR #710, opened 2024-03-13
Support modifying the message, such as encryption.

lhpqaq:
I made modifications to session.py to support obtaining and modifying the mutated version of the message before sending it. To encrypt the fields requiring encryption in the message, you can use the following method:
```python
secret_key = None

def pre_send_callback(target, fuzz_data_logger, session, sock):
    mc = session.mutation_context

    # Method 1:
    # The mutated data packet to be sent
    original_data = session.fuzz_node.render(mc).hex()
    if secret_key:
        modified_data = original_data^secret_key
    session.modified_data = modified_data

    # Method 2:
    if list(mc.mutations.keys())[0] == "field requiring encryption"
        mc.mutations[list(mc.mutations.keys())[0]].value ^= secret_key

def post_test_case_callback(target, fuzz_data_logger, session, sock):
    if not secret_key:
        receive_data = session.last_recv.hex()
        secret_key = get_secret_key(receive_data)

session = Session(target=Target(SocketConnection(host, int(port))),receive_data_after_fuzz=True,,post_test_case_callbacks=[post_test_case_callback],pre_send_callbacks=[pre_send_callback])
```

### DIFF
--- a/boofuzz/sessions/session.py
+++ b/boofuzz/sessions/session.py
@@ -138,6 +138,8 @@ class Session(pgraph.Graph):
         target=None,
         web_address=constants.DEFAULT_WEB_UI_ADDRESS,
         db_filename=None,
+        mutation_context=None,
+        modified_data=False
     ):
         self._ignore_connection_reset = ignore_connection_reset
         self._ignore_connection_aborted = ignore_connection_aborted
@@ -858,6 +860,9 @@ class Session(pgraph.Graph):
         else:
             data = self.fuzz_node.render(mutation_context)
 
+        if self.modified_data:
+            data = self.modified_data
+            self.modified_data = False
         try:  # send
             self.targets[0].send(data)
             self.last_send = data
@@ -1081,6 +1086,8 @@ class Session(pgraph.Graph):
                 ):
                     self._fuzz_data_logger.open_test_step("restart interval of %d reached" % self.restart_interval)
                     self._restart_target(self.targets[0])
+
+                self.mutation_context = mutation_context
 
                 self._fuzz_current_case(mutation_context)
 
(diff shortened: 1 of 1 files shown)


phix33:
Curious if this approach is going to be accepted? I'd also like to encrypt some data blocks post-mutation and just prior to sending. Will probably adopt this approach in the meantime - thanks @lhpqaq

lhpqaq:
> Curious if this approach is going to be accepted? I'd also like to encrypt some data blocks post-mutation and just prior to sending. Will probably adopt this approach in the meantime - thanks @lhpqaq

My code might be a bit rudimentary, and the author may not choose to accept it.

jtpereyda:
Thanks for the PR! A few notes:

1. There is a way to do this with blocks: Create a block type that has an encode function. The encode function encodes data after mutation. For an example, see `repeat.py`.
2. That said, I'm not entirely opposed to this approach.
3. One note on the PR: It seems like it might make sense not to have modified_data in the constructor. At least for the described use case, it seems like something you always set after construction.

lhpqaq:
> Thanks for the PR! A few notes:
> 
> 
> 
> 1. There is a way to do this with blocks: Create a block type that has an encode function. The encode function encodes data after mutation. For an example, see `repeat.py`.
> 
> 2. That said, I'm not entirely opposed to this approach.
> 
> 3. One note on the PR: It seems like it might make sense not to have modified_data in the constructor. At least for the described use case, it seems like something you always set after construction.

Thanks. My understanding of boofuzz is not deep enough. I will continue to study it in the next few days.

phix33:
> Thanks for the PR! A few notes:
> 
>     1. There is a way to do this with blocks: Create a block type that has an encode function. The encode function encodes data after mutation. For an example, see `repeat.py`.

Ahh nice! Hadn't realised we could call `s_block("block2", encoder=encrypt_block2)` then `encrypt_block2(block2)` gets called with the mutated block! Much simpler, thanks!

(I _now_ see the [iso8385.py](https://github.com/jtpereyda/boofuzz/blob/master/examples/iso8385.py) example too!).

PR #712, opened 2024-04-04
Handle TLS handshake timeouts in SSLSocketConnection

constcast:
SSLSocketConnection prior to this commit would hang identifitly if a timeout during the TLS handshake occurs. This can happen if a target accepts a TCP connection but does not send any TLS handshake messages, e.g. because of a bug triggered by the fuzzing.

This commit makes boofuzz observe timeouts during the TLS handshake. TLS handshake timeouts are handled in the same way as timeouts during TCP connection setup.

### DIFF
--- a/boofuzz/connections/ssl_socket_connection.py
+++ b/boofuzz/connections/ssl_socket_connection.py
@@ -54,7 +54,21 @@ class SSLSocketConnection(tcp_socket_connection.TCPSocketConnection):
             # No SSL context set
             pass
 
-        super(SSLSocketConnection, self)._connect_socket()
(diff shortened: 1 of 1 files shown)


constcast:
In order to reproduce the problem that this pull request is trying to fix, use the following setup:

Run netcat in one shell:

     sudo nc -lnvp 443

Run example/fuzz_ssl_client.py

    python3 example/fuzz_ssl_client.py

Observation: The connection to the netcat instance is created, but netcat does not respond to the TLS handshake. The fuzzer will hang forever. After this commit, this hanging will result in a proper error.

SR4ven:
Thank you for the PR @constcast, nice catch!

I just reformatted the code according to black.

PR #760, opened 2026-01-21
Proposal for a CAN-based fuzzer example

edgecase1:
I made some experiments with a CAN-based fuzzer using Boofuzz and the automotive simulator.


jtpereyda:
Thanks @edgecase1 !

